PR #152, opened 2015-01-20
Fixed incorrect lock initializations

Finomnis:


lib/CL/devices/pthread/pthread.c
```diff
@@ -141,7 +141,7 @@ static thread_arguments* new_thread_arguments ()
 {
   thread_arguments *ta = NULL;
   POCL_LOCK (ta_pool_lock);
-  if (ta = thread_argument_pool)
+  if ((ta = thread_argument_pool))
```
pjaaskel: Why the extra parenthesis?

Finomnis: To silence a warning of you compile this code with clang. Clang thinks you meant to write ==,  and wants you to explicitly make clear that this is not the case by adding extra parenthesis.

PR #162, opened 2015-03-06
Fixed image for clCreateImage3D

matszpk:
Small fix for image type when 3D image will be created by using clCreateImage3D.


matszpk:
Thank you, pjaaskel :)

PR #166, opened 2015-04-01
Windows fixes

franz:
Most of it is cleanup, except moving POCL_DEBUG macros into pocl_debug.{h,c} and adding timestamps on windows to them.


pjaaskel:
Seems to include a lot of generic (good) cleanups also. I cannot test the CMake/Windows, but looks good otherwise.

PR #167, opened 2015-04-01
Link failures

Oblomov:
While testing some simple programs in a course I'm running I discovered an interesting issue with pocl: if a kernel called 'init' is present, the program will fail to launch due to a link error in the device code. However, the link error doesn't actually happen when building the program, but when kernels are actually invoked.

This patch series thereby tests for two issues:

1. a spec violation, by which link failures are not reported from within clBuildProgram(), even though they should (instead, they are reported further on when the kernels are actually linked at launch time);
2. the actual 'init' bug, which is caused by the kernel launcher name mangling prepending a '_' to the kernel name, causing the '_init' symbol to be double-defined;

The last patch actually fixes the second bug, by changing the mangling scheme into something more sophisticated.

Note that with this patchset the clBuildProgram test will fail due to the delayed linking behavior in current pocl.

### DIFF
--- a/lib/CL/devices/basic/basic.c
+++ b/lib/CL/devices/basic/basic.c
@@ -44,7 +44,7 @@
 #define max(a,b) (((a) > (b)) ? (a) : (b))
 
 #define COMMAND_LENGTH 2048
-#define WORKGROUP_STRING_LENGTH 128
+#define WORKGROUP_STRING_LENGTH 1024
 
 struct data {
   /* Currently loaded kernel. */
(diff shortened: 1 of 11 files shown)


pjaaskel:
Can you create a separate regression test of the failing case and mark it XFAIL? The test suite should have "expected" status at all time. Otherwise we cannot know if a test is known broken or not.

pjaaskel:
Also, the rename should be reflected in the *.kernel_obj.c generation in pocl_llvm_api.cc. This generated C file is used to link in info of the work group functions in heterogeneous device program images (read TCE). Check around line 823.

Oblomov:
New version, hopefully I addressed both points (I also spotted another couple of places where the kernel name mangling had to be corrected).

Issue #171, opened 2015-04-05
Hang upon queue.enqueueReadBuffer()


matthewharvey:
What I did:
-Ran the attached (I will attach it) code on a core i5 machine.
-The code is reasonably simple. I've tried to make it rather straightforward. There are two kernels. One must finish before the other starts. Then I read two of the buffers back to the host. I had to strip down the code a little to post it, but what is happening should be clear. I've tried running the exact attached code and can reproduce with it.

What happens:
-The code makes a call to enqueueReadBuffer, and never returns from that call. It hangs on an implicit call to clFinish based on looking at the backtrace from a gdb run.

#0  0x00007ffff6f189ca in clFinish () from /usr/lib/x86_64-linux-gnu/libpocl.so.1
#1  0x00007ffff6f13147 in clEnqueueReadBuffer () from /usr/lib/x86_64-linux-gnu/libpocl.so.1
#2  0x0000000000403e8f in cl::CommandQueue::enqueueReadBuffer (this=0x7fffffffe060, buffer=..., blocking=1, offset=0, size=20480, ptr=0xdc2080, 
    events=0x7fffffffe000, event=0x0) at /usr/local/include/CL/cl.hpp:5290
#3  0x0000000000402a75 in main (argc=1, argv=0x7fffffffe2a8) at src/test.cpp:134

What is expected:
-I'm pretty sure my code makes sense, so I'd expect it to work, but even if my code is malformed, I'm not sure hanging is the correct response.

What I've found out:
-If I remove the queue.finish() at line 126 of test.cpp, then everything works as expected. That probably has something to do with it.

I tried to get a development environment up with the latest pocl code, but I couldn't work it out. This was run on the version of pocl in debian Jessie repositories, which is labelled as 0.10-10.

matthewharvey:
I thought I'd just be able to attach it. Turns out it's a little more complex. Hopefully others can access this:
https://www.dropbox.com/s/rnz4458br91da89/pocl_bug_test_2.tar.gz?dl=0

PR #172, opened 2015-04-05
Always clone event list

Oblomov:
This fixes issue #171. I've also taken the opportunity to do some whitespace cleanup and typo comment fixes en passant.


pjaaskel:
This one is fine, and cleanup in general is great, but better not commit pure white space fixes too much as they mess up the readability of the commit history, especially 'git blame'.  Fixing around code which is also otherwise modified is fine.

Oblomov:
Good point about the cleanup, sorry. I started in the whereabouts of where I was looking or thought I should be looking, and then got carried away. I'll be more selective in the future 8-)

PR #173, opened 2015-04-05
Two more tests

Oblomov:
This patchset adds two small tests:

* the first is a simple test for reading, copying and writing buffers (nothing special, passes as-is);
* the second is a test for the “neverending loop” that was fixed by the (merged) pull requesnt #172 (I tested it without those patches and it fails as expected, after a 30s timeout, it passes with the latest master).

### DIFF
--- a/tests/runtime/test_event_cycle.c
+++ b/tests/runtime/test_event_cycle.c
@@ -1,0 +1,152 @@
+/* Test that recycling wait lists doesn't lead to lockups
+
+   Copyright (C) 2015 Giuseppe Bilotta <giuseppe.bilotta@gmail.com>
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to deal
+   in the Software without restriction, including without limitation the rights
+   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+   copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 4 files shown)


tests/runtime/test_event_cycle.c
```diff
@@ -0,0 +1,125 @@
+#include <stdio.h>
```
pjaaskel: Copyright + license text missing.

tests/runtime/test_read-copy-write-buffer.c
```diff
@@ -0,0 +1,93 @@
+#include <stdio.h>
```
pjaaskel: License + copyright missing as well.

tests/runtime/test_event_cycle.c
```diff
@@ -0,0 +1,125 @@
...
+        return EXIT_FAILURE;
+
+      cl_event buf1_event, bufcp_event, buf2_event;
+
+      /* we test if recycling the wait list leads to neverending loops */
+      cl_event wait_list[1];
+
+      /* Note that this must be CL_TRUE because to trigger the bug the next command must have a completed event
+       * in the waiting lists */
+      if (clEnqueueWriteBuffer(queue, buf1, CL_TRUE, 0, buf_size, host_buf1, 0, NULL, &buf1_event) != CL_SUCCESS)
```
pjaaskel: These lines seem they could be above 80 columns.

Oblomov:
New version with licenses and clipped lines ;-)

PR #175, opened 2015-04-08
Fix kcache

franz:
Master merged in today.

There are a few remanining issues:
1) TCE device switches remains broken
2) no pthread locks around cache functions
3) deprecation warnings in pocl_llvm_api.cc, despite defining CL_USE_DEPRECATED... 
4) tmpnam() is probably not ideal (annoying warnings)
5) no option to disable filelocks (not sure if that matters)


include/pocl_file_util.h
```diff
@@ -0,0 +1,87 @@
...
+void* acquire_lock(const char* path, int shared);
+
+int lock_is_owned(void* lock);
+
+void release_lock(void* lock);
+
+
+
+/* Remove a directory, recursively */
+int pocl_rm_rf(const char* path);
```
pjaaskel: perhaps 'remove_dir' is more descriptive to non-unix users

include/pocl_file_util.h
```diff
@@ -0,0 +1,87 @@
...
+
+/* Make a directory, including all directories along path */
+int pocl_mkdir_p(const char* path);
```
pjaaskel: Same here. 'mkpath' or similar?

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -230,50 +228,25 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
   POCL_RETURN_ERROR_COND((event_wait_list != NULL && num_events_in_wait_list == 0),
     CL_INVALID_EVENT_WAIT_LIST);
 
+  cache_lock = pocl_cache_acquire_writer_lock(kernel->program,
+                                              command_queue->device);
+  assert(cache_lock);
```
pjaaskel: Is this the situation that can happen after someone hits ctrl-c during a kernel build and runs the program again? I think a more descriptive error message (how to recover from this) should be in place.

lib/CL/pocl_cache.c
```diff
@@ -0,0 +1,470 @@
...
+// required in llvm API
+void pocl_cache_program_bc_path(char*        program_bc_path,
+                                cl_program   program,
+                                unsigned     device_i) {
+    program_device_dir(program_bc_path, program,
+                       device_i, POCL_PROGRAM_BC_FILENAME);
+}
+
+// required in llvm API
+void pocl_cache_kernel_so_path(char* kernel_so_path, cl_program program,
```
pjaaskel: It's actually "work_group_function_so_path", there can be many per kernel, one for each local size.

lib/CL/pocl_cache.c
```diff
@@ -0,0 +1,470 @@
...
+                                        size_t       local_z) {
+    assert(bc);
+
+    char kernel_parallel_path[POCL_FILENAME_LENGTH];
+    make_kernel_cachedir_path(kernel_parallel_path, program, device_i,
+                              kernel, local_x, local_y, local_z);
+
+    assert( strlen(kernel_parallel_path) <
+            (POCL_FILENAME_LENGTH - strlen(POCL_PARALLEL_BC_FILENAME)));
+    strcat(kernel_parallel_path, "/" POCL_PARALLEL_BC_FILENAME);
```
pjaaskel: "/" is already in the POCL_PARALLEL_BC_FILENAME, isn't it? Remove it from the macro?

lib/CL/pocl_cache.c
```diff
@@ -0,0 +1,470 @@
...
+#endif
+
+
+/******************************************************************************/
+
+static void pocl_cache_init_topdir() {
+
+    char *tmp_path=NULL;
+
+    tmp_path = getenv("POCL_CACHE_DIR");
```
pjaaskel: Use pocl_runtime_config.c for envs. We might add a configuration file or similar method to change these parameters later on.

lib/CL/pocl_util.c
```diff
@@ -411,103 +294,7 @@ void pocl_command_enqueue (cl_command_queue command_queue,
...
-
-  POCL_LOCK(cache_lock);
 
-  if (!pocl_get_bool_option("POCL_KERNEL_CACHE", POCL_BUILD_KERNEL_CACHE))
-    {
-      cache_dirty = 1;
-      goto bottom;
-    }
-
-  /* If program contains "#include", disable caching
```
pjaaskel: Now there's no limitation with #includes, right? You added preprocessing? Can you add a test that tests this? I.e., caching of kernels with #includes?

lib/CL/pocl_llvm_api.cc
```diff
@@ -262,22 +202,17 @@ int pocl_llvm_build_program(cl_program program,
...
+  // TODO this currently passes NULL as device tmpdir
+  cl_device_id device = program->devices[device_i];
+  if (device->ops->init_build != NULL)
     {
-      assert (device_tmpdir != NULL);
-      char *device_switches = 
-        device->ops->init_build (device->data, device_tmpdir);
-      if (device_switches != NULL) 
+      char *device_switches =
+        device->ops->init_build (device->data, NULL);
```
pjaaskel: Better remove the device_tmpdir argument from the driver API if it's not supported.

pjaaskel:
```
LLVMFileUtils.cc: In function 'int pocl_write_module(void*, const char*, int)':
LLVMFileUtils.cc:208:64: error: no matching function for call to 'llvm::raw_fd_ostream::raw_fd_ostream(const char*&, std::error_code&, llvm::sys::fs::OpenFlags)'
     raw_fd_ostream os(path, ec, sys::fs::F_RW | sys::fs::F_Excl);
                                                                ^
LLVMFileUtils.cc:208:64: note: candidates are:
In file included from /home/visit0r/local/stow/llvm-3.5/include/llvm/Support/FormattedStream.h:18:0,
                 from LLVMFileUtils.cc:25:
/home/visit0r/local/stow/llvm-3.5/include/llvm/Support/raw_ostream.h:359:3: note: llvm::raw_fd_ostream::raw_fd_ostream(int, bool, bool)
   raw_fd_ostream(int fd, bool shouldClose, bool unbuffered=false);
   ^
/home/visit0r/local/stow/llvm-3.5/include/llvm/Support/raw_ostream.h:359:3: note:   no known conversion for argument 2 from 'std::error_code' to 'bool'
/home/visit0r/local/stow/llvm-3.5/include/llvm/Support/raw_ostream.h:354:3: note: llvm::raw_fd_ostream::raw_fd_ostream(const char*, std::string&, llvm::sys::fs::OpenFlags)
   raw_fd_ostream(const char *Filename, std::string &ErrorInfo,
```

With my LLVM 3.5. Can you provide a fallback for older LLVMs?

I'll fix the TCE part in master without this merge.

include/pocl_file_util.h
franz: Good point. Technically its remove_dir_recursively, also as mkdir_p is recursive... do you mind longer function names ( i dont), or should i stick to just remove_dir ?

lib/CL/clEnqueueNDRangeKernel.c
franz: By situation you mean "cache_lock == NULL" ? No, that doesn't happen in the scenario you described. Lockfiles contain the pid of the locking process, which is checked when you try to acquire the lock. So leftover lockfiles are not a problem. The NULL might happen if some process holds the lock for very long time, like minutes (so i think it shouldn't happen much).

lib/CL/pocl_cache.c
franz: OK, will change

lib/CL/pocl_cache.c
franz: Nope, not in POCL_PARALLEL_BC_FILENAME. That define is still in pocl_cl.h because it's used in other places (like devices/common.c). But i can move it to pocl_cache and add the "/", it would be more consistent, i guess.

lib/CL/pocl_util.c
franz: Yup, removed all the old code for grepping #includes. Yeah i guess a test would be nice :)

lib/CL/pocl_llvm_api.cc
franz: Not sure what you mean here. I thought you already removed it from ops->init_build with your commit 3ed57f99b7dbb6 or not ? So if i merge that commit, it should be gone.

pjaaskel: Yes I did, after commenting here.

pjaaskel:
Older LLVMs are lower priority, so there's no need for fancy solutions for them in case of the cache locking, for example (compared to the current status, anything is better).

include/pocl_file_util.h
pjaaskel: remove_dir is fine, IMO.

PR #179, opened 2015-04-12
Subdevices

Oblomov:
A test for and a draft implementation of clCreateSubDevices.

### DIFF
--- a/lib/CL/clCreateSubDevices.c
+++ b/lib/CL/clCreateSubDevices.c
@@ -1,17 +1,18 @@
 /* OpenCL runtime library: clCreateSubDevices()
 
    Copyright (c) 2014 Lassi Koskinen
-   
+                 2015 Giuseppe Bilotta
+
    Permission is hereby granted, free of charge, to any person obtaining a copy
    of this software and associated documentation files (the "Software"), to deal
    in the Software without restriction, including without limitation the rights
    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
    copies of the Software, and to permit persons to whom the Software is
    furnished to do so, subject to the following conditions:
-   
+
    The above copyright notice and this permission notice shall be included in
    all copies or substantial portions of the Software.
-   
+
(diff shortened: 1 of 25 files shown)


lib/CL/clCreateSubDevices.c
```diff
@@ -38,37 +40,117 @@ POname(clCreateSubDevices)(cl_device_id in_device,
...
+     // CUs of in_device
+     POCL_GOTO_ERROR_COND(
+       (properties[1] == 0 || properties[1] > in_device->max_compute_units),
+       CL_INVALID_VALUE);
+     // error out if properties isn't zero-terminated
+     POCL_GOTO_ERROR_COND(properties[2] != 0, CL_INVALID_VALUE);
+
+     count_devices = in_device->max_compute_units / properties[1];
+     num_props = 3; // partition type, CUs per device, terminating 0
+   } else {
```
pjaaskel: Shouldn't this be else if ( ... Cl_DEVICE_PARTITION_BY_COUNTS)  and the actual else to return 0 due to an unsupported partitioning type?

pjaaskel: Or, actually, if all the described partitioning types are required by the standard, the 'else' should use some default one and print out a POCL_UNIMPLEMENTED warning?

pjaaskel:
To me it's logical that the subdevices (due to them being children of a device) belong to the parent's context by default.  Due to the fact that it's a _subdevice_ (by definition tightly bound to a device) and there's the specs clause that states that subdevice can be used wherever its parent can (without further actions like somehow putting it to the same context with the parent).

Oblomov:
While for it's logical, for me it's not, and neither is for others (see e.g. recent debare on `#opencl`). Pending official clarifications on the matter, I'd rather go with the simpler-to-implement strategy of subdevices as independent devices than the more sophisticated, invasive and fragile approach of letting them act as if they were their own parent devices. If the latter is then officially sanctioned as the correct behavior by Khronos, we can always extend the implementation to handle things that way.

pjaaskel:
Fair enough.

PR #185, opened 2015-04-25
Memory access violation fixes

Oblomov:
This is a second round of fixes that are determined by more-or-less static analysis (where 'static' refers to my own eyes this time). I got interested in this round of fixes when I noticed a double-free when I happened to use an invalid event in a waitlist, so I went through all the clEnqueue* that seemed to be affected by similar issues, wrote tests triggering the relevant errors (mostly segmentation faults) and fixed those that I found.

### DIFF
--- a/lib/CL/clEnqueueMapImage.c
+++ b/lib/CL/clEnqueueMapImage.c
@@ -39,14 +39,16 @@ POname(clEnqueueMapImage)(cl_command_queue   command_queue,
                           size_t *           image_slice_pitch,
                           cl_uint            num_events_in_wait_list,
                           const cl_event *   event_wait_list,
-                          cl_event *         event,
-                          cl_int *           errcode_ret ) 
+                          cl_event *         event_ret,
+                          cl_int *           errcode_ret )
 CL_API_SUFFIX__VERSION_1_0
 {
   cl_int errcode;
   size_t offset;
   void *map = NULL;
+  int using_host_buffer = 0;
   cl_device_id device;
+  cl_event event = NULL;
   _cl_command_node *cmd = NULL;
   mem_mapping_t *mapping_info = NULL;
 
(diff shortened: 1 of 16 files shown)


lib/CL/clEnqueueMapImage.c
```diff
@@ -46,7 +46,9 @@ CL_API_SUFFIX__VERSION_1_0
   cl_int errcode;
   size_t offset;
   void *map = NULL;
+  int using_host_buffer = 0;
   cl_device_id device;
+  cl_event _event = NULL;
```
pjaaskel: Why underscore?

lib/CL/clEnqueueMapImage.c
```diff
@@ -139,40 +146,34 @@ CL_API_SUFFIX__VERSION_1_0
...
     (*errcode_ret) = CL_SUCCESS;
 
   return map;
- 
- ERROR:
-  POCL_MEM_FREE(map);
+
+ERROR:
+  POCL_MEM_FREE(_event);
+  // POCL_MEM_FREE(map); // NOPE! TODO FIXME this should be a call to unmap_mem
```
pjaaskel: Why not call it then? And only C-style comments, pls.

tests/runtime/test_read-copy-write-buffer.c
```diff
@@ -51,6 +51,10 @@ main(void)
     if (err != CL_SUCCESS)
       return EXIT_FAILURE;
 
+    /* Only test the devices we actually have room for */
+    if (ndevices > MAX_DEVICES)
+	    ndevices = MAX_DEVICES;
```
pjaaskel: Extra indent.

pjaaskel:
Only nitpicks found, will merge if tests pass.

lib/CL/clEnqueueMapImage.c
Oblomov: Actually, with the reordered code map should never be != NULL here, since there is no point of failure after it has been assigned a non-NULL value. I can replace that line with an assert or just remove it.

PR #207, opened 2015-07-07
Use XDG_CACHE_HOME

### LINKED ISSUE #200: [wishlist] Should use XDG ~/.cache/pocl instead of ~/.pocl/kcache
It'd be nice if POCL adhered (at least on Linux) to the XDG spec:

http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

That makes it easy to wipe all sorts of caches at once.

larsmans:
Fixes #200. Also checked the return value from `snprintf` while I was at it.


pjaaskel:
Thanks for your contribution. What is your full name that I should add to CREDITS?

larsmans:
It's in the commit log.

inducer:
Awesome, thank you so much!

PR #208, opened 2015-07-07
Remove outdated LAUNDRY file

larsmans:
This is both a patch to remove an outdated file, and a question.

Am I right in thinking that the `find_env` function in `pocl_runtime_config.c` is a cache for `getenv`, and if so, does this speed anything up? The thing is that `getenv` is not, as the `LAUNDRY` file asserts, a system call: on all Unix systems I've seen, the environment is kept in user memory, in the global `char **environ`, and `getenv` is a regular function that searches this list. See, e.g., the [OpenBSD version](http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/lib/libc/stdlib/getenv.c?rev=1.10). It seems rather wasteful to reinvent `getenv` in this way, even if it has the benefit of only searching the variables actually looked at by pocl.


pjaaskel:
You seem to be right that getenv is not a syscall in Linux, but are we sure it's not in other platforms such as Windows or the Linux variant of Android? Even though it's now a dummy wrapper for getenv(), another purpose for pocl_runtime_config is to be a wrapper for pocl-controlling settings so we can later more easily add configuration files or some other means to change the pocl's behavior (Windows registry or some such) that this wrapper hides.

larsmans:
Looks like Windows [implements the environment in the same way](https://msdn.microsoft.com/en-us/library/aa246807%28VS.60%29.aspx) (not surprising since the environment came from DOS, which took it from Unix). Android gets [its getenv](https://android.googlesource.com/platform/bionic/+/master/libc/upstream-openbsd/lib/libc/stdlib/getenv.c) from OpenBSD :)

pjaaskel:
OK. Well, it doesn't do harm to cache the variables there, the code snippet is rather small, so I wouldn't remove it just for the sake of a cleanup.

PR #209, opened 2015-07-08
Lots of small fixes

larsmans:
See commit messages for details.


pjaaskel:
Some bots now scream:
```
pocl_util.c:164:16: error: assignment makes pointer from integer without a cast [-Werror]
pocl_util.c:165:3: error: return makes pointer from integer without a cast [-Werror]
cc1: all warnings being treated as errors
```
http://tce.cs.tut.fi:8010/builders/x86-64-precise-kcache/builds/160/steps/compile/logs/stdio

PR #214, opened 2015-07-15
Initialize sampler argument

hugwijst:
Initialize the device specific `dev_sampler_t` structure from the `cl_sampler_t` structure for the basic and pthread devices.

### DIFF
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -43,6 +43,8 @@
 #include "pocl_runtime_config.h"
 #include "pocl_llvm.h"
 
+#include "_kernel_constants.h"
+
 #define COMMAND_LENGTH 2048
 
 /**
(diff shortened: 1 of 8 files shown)


lib/CL/devices/common.h
```diff
@@ -26,7 +26,7 @@
 #define POCL_COMMON_H
 
 #include "pocl_cl.h"
-#include "dev_image.h"
+#include "_kernel_c.h"
```
pjaaskel: _kernel_c.h should be included only from the kernel builtin function definitions (lib/kernel). It doesn't seem to build with gcc at the moment due to #if __has_attribute(__always_inline__) which is recognized by Clang but not gcc 4.8. Can you revert this part in such a way that the dev_image.h has only the image struct which _kernel_c.h includes?

hugwijst: I've copied the required `#define`s to `dev_image.h`.

Is the `cl_int` type also defined in the kernel context? In that case the types in `include/pocl_types.h` could be changed to `cl_int` and the `#define`s can be moved to that file. That way `dev_images.h` can be removed.

pjaaskel: I believe cl_int is not defined in kernel context, it's simply int there, I think.

lib/CL/devices/common.c
```diff
@@ -135,6 +135,39 @@ void fill_dev_image_t (dev_image_t* di, struct pocl_argument* parg,
...
+ */
+void fill_dev_sampler_t (dev_sampler_t *ds, struct pocl_argument *parg)
+{
+  cl_sampler_t sampler = *(cl_sampler_t *)parg->value;
+
+  *ds = 0;
+  *ds |= sampler.normalized_coords == CL_TRUE ? CLK_NORMALIZED_COORDS_TRUE :
+      CLK_NORMALIZED_COORDS_FALSE;
+
+  switch (sampler.addressing_mode) {
```
pjaaskel: What is the point of this switch? Why not or straight it to *ds?

hugwijst: The `CL_ADDRESS_*` constants defined in "include/CL/cl.h" don't correspond with the `CLK_ADDRESS_*` constants as defined in "include/_kernel_c.h" and currently copied in "lib/CL/devices/dev_image.h".

hugwijst:
I've moved the constants from `include/_kernel_c.h` to `include/_kernel_constants.h`, so it can be included from `lib/CL/devices/common.c` and doesn't have to be duplicated.

PR #228, opened 2015-08-14
cl2.hpp + removing OpenGL header dependency

jrprice:
This patch adds the new `cl2.hpp` header recently made available by Khronos. The primary motivation for doing this is to remove the requirement to have OpenGL headers available when building pocl.

This is mostly just very minor changes to the tests that use `cl.hpp` to include the new header instead, and to define the macros which control the OpenCL version to target. The Makefiles have also been updated to use `-std=c++11` as required.

I've cut out the OpenGL checks from `configure.ac` and the main `CMakeLists.txt`. Not sure if the stuff that I've left in is necessary/useful.

At present, I've left the old header in, and the patches and wrapper in `fix-include`, as I wasn't sure if they were still needed for some of the examples. The new header doesn't seem to need the patch, and I don't think it needs the wrapper that suppresses the deprecation warnings, since the Makefiles all use `-Wno-deprecated-declarations` anyway.

Tested this on OS X and Linux (x86 + armhf). Might be worth testing on Android + Windows too, but I don't have systems set up that I can do this on.


pjaaskel:
Thanks. Can you add a note to CHANGES?

jrprice:
Done. Also made sure `cl2.hpp` gets installed with other headers (and included in `make dist`).

PR #234, opened 2015-08-25
[cmake] Make it possible to override CL_DISABLE_HALF.

dsandersimgtec:
I suspect the check is currently wrong. It seems to be checking whether C
allows half precision types in arguments and returns. On Mips, this is not
currently allowed. However, I suspect OpenCL allows this regardless of
target. Regardless of the truth of the matter, we should be able to forcefully
enable half precision types in pocl.

This patch removes the unconventional caching mechanism for
CL_DISABLE_HALF in favour of cmake's usual method which allows the user
to specify it with -DCL_DISABLE_HALF=ON/OFF.

### DIFF
--- a/cmake/LLVM.cmake
+++ b/cmake/LLVM.cmake
@@ -650,9 +650,7 @@ set_cache_var(CL_DISABLE_LONG "Disable cl_khr_int64 because of buggy llvm")
 
 ####################################################################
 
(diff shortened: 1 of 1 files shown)


pjaaskel:
Yes, that has been wrong as the C's half type has (at least used to have) a completely different (and more buggy) code generation path than OpenCL C's half.

PR #240, opened 2015-09-04
Substitute spir-* triples for the default target triple so that Address Spaces are lowered correctly.

dsandersimgtec:
The root of the problem is that the spir-* triple does not indicate the underlying
target architecture and the lowering of address spaces is therefore skipped
when the target is MIPS.

This fixes example1-spir32 for MIPS.

### DIFF
--- a/lib/CL/pocl_llvm_api.cc
+++ b/lib/CL/pocl_llvm_api.cc
@@ -675,6 +675,7 @@ int pocl_llvm_get_kernel_metadata(cl_program program,
       return 1;
     }
 
+  input->setTargetTriple(program->devices[device_i]->llvm_target_triplet);
 
 #ifdef DEBUG_POCL_LLVM_API        
   printf("### fetching kernel metadata for kernel %s program %p input llvm::Module %p\n",


lib/llvmopencl/TargetAddressSpaces.cc
```diff
@@ -92,6 +92,9 @@ TargetAddressSpaces::runOnModule(llvm::Module &M) {
 
   llvm::StringRef arch(M.getTargetTriple());
 
+  if (arch.startswith("spir"))
+    arch = OCL_KERNEL_TARGET;
```
pjaaskel: This works only for host devices? Remember that pocl can control platforms with multiple devices that are not the same as host. And the same SPIR can be retargeted to all devices in the platform (which could consist of, say a MIPS CPU and a HSA-supported GPU in the future).   So a better fix would be to dig out the device we are compiling to somehow.

dsandersimgtec: I don't have a suitable system to double check with but I believe this would work with foreign target devices too. However, it does exclude heterogeneous systems.

> So a better fix would be to dig out the device we are compiling to somehow.

I'm thinking that the code that decided the target device ought to be switching the spir triple for the target triple and passing that to the code generator, etc. Can you tell me where that decision is made?

pjaaskel: Yes, I meant heterogeneous platforms with one or more non-host device. 

For the codegen, check 'basic.c', scan for llvm_codegen() (it's in common.c). But the kernel compilation passes (that work on LLVM IR) are driven from pocl_llvm_api.cc. It _might_ work by just forcing the real target machine be created in GetTargetMachine() instead of the SPIR target and fetching the actual target machine (instead of the module triplet) in TargetAddressSpaces.

dsandersimgtec:
This version is an improvement but I don't think it's quite right yet. In this version we substitute the SPIR triple for the Device triple as soon as we read from the cache. As a result, it should now support heterogeneous systems.

I don't think SPIR inputs should be entering the cache with a 'target triple = "spir...' line though.

pjaaskel:
In kcache the final binary (parallel.so) is what matters the most, if the intermediate files are in SPIR it's no big deal. 

Eventually I'd like to get an early address space flattening inside pocl also to fix issues we see now with the vectorizer and Broadwell. That is, run TAS early and at the same time switch from SPIR to the target triplet in the Module to get proper target info to the LLVM IR optimizations when starting from SPIR.

I'll test your patch here.

pjaaskel:
There's one new failure with this, didn't you get it with MIPS?
```
compilation -*-
108. testsuite-runtime.at:92: testing clGetKernelArgInfo ...
./testsuite-runtime.at:95: $abs_top_builddir/tests/runtime/test_clGetKernelArgInfo
stderr:
Assertion: 
(kernel_arg.address==CL_KERNEL_ARG_ADDRESS_CONSTANT) && "arg of test_kernel is not CONSTANT addr space"
failed on test_clGetKernelArgInfo.c:80
stdout:

NON-SPIR

SPIR with metadata
opening File: /home/visit0r/src/pocl-public/tests/runtime/clGetKernelArgInfo.spir64_meta
program size: 2596
./testsuite-runtime.at:95: exit code was 1, expected 0
108. testsuite-runtime.at:92: 108. clGetKernelArgInfo (testsuite-runtime.at:92): FAILED (testsuite-runtime.at:95)
```

dsandersimgtec:
I'm still running the full set of tests. Some of them take a very long time on the first run, particularly test_convert_type_16 which takes ~1.5hrs at the moment.

I've manually run test_clGetKernelArgInfo and I get the same assertion. I'll look into it.

dsandersimgtec:
I'm fairly certain I have a better solution that patches llvm instead of pocl. It definitely fixes example1-spir32 without regressing test_clGetKernelArgInfo and I'm currently re-running the tests to see if everything else is ok.

The plan is to (like X86) reserve address spaces 1-255 for software use and treat conversions between these spaces as no-ops.

PR #260, opened 2015-10-12
[cmake] When cmake supports it, use USES_TERMINAL on the check target.

dsandersimgtec:
This should partially fix the timeouts seen on the mipsel-32r2-cmake
builder since ninja will stop buffering the ctest output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,14 @@ set(POCL_VERSION ${VERSION_STRING})
 
 ######################################################################################
 
+# Recent versions of CMake can make use of Ninja's console pool to avoid
+# buffering the output of particular commands.
+if(CMAKE_VERSION VERSION_LESS 3.2.0)
+  set(COMMAND_USES_TERMINAL)
+else()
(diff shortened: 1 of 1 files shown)


franz:
Hi Daniel,

Pekka added the 0.12 milestone - do you agree & want this in 0.12 ?

If so, please next time fork official pocl's release_0_12 branch and create your commits there, i'll pull those into both master & 0.12. Merging master into release_0_12 messes up not only history, but also versioning, so we like to avoid that.

This one PR actually can merge into both master &  release_0_12 without issues i think, so no problem now.

So, which one is it ? master only, or master + 0.12 ?

dsandersimgtec:
I don't think it's necessary for 0.12, it's really intended for the buildbot I'm bringing up. Without it, ninja will still work but will buffer test output until all tests have finished and then emit it all at once. On the buildbot this triggers the no-output timeout.

Outside the buildbot, I currently work around this using 'ninja -v check' to get the underlying ctest command, then I kill it and then run ctest directly.

franz:
OK, then i think its good to go.

Issue #181, opened 2015-04-13
Testsuite fails with pocl-0.11 and llvm-3.6


fabiand:
The testsuite fails on Fedora Rawhide (23) with llvm 3.6:

```
 20: Kernel functions abs abs_diff add_sat hadd mad_hi mad_sat mul_hi rhadd sub_sat (loopvec) ok
 21: Kernel functions abs abs_diff add_sat hadd mad_hi mad_sat mul_hi rhadd sub_sat (loops) ok
 22: Kernel functions << >> rotate                   ok
 23: Trigonometric functions                         ok
 24: Sampler address clamp                           FAILED (testsuite.at:264)
 25: Image query functions                           FAILED (testsuite.at:272)
 26: Kernel functions: shuffle charN                 ok
 27: Kernel functions: shuffle shortN                ok
```

fabiand:
The whole log is available http://fpaste.org/210445/42895330/

besser82:
Full build logs are [available on Koji](https://koji.fedoraproject.org/koji/taskinfo?taskID=9454480).

pjaaskel:
I testsuite.log or better the individual logs in testsuite.dir/{24,25} somewhere? Is the Rawhide's LLVM standard or some SVN revision?

besser82:
LLVM in Rawhide is from released 3.6.0-tarball, no scm-snapshot or something…

Here goes the individual logs:
* [test log from 24](http://paste.fedoraproject.org/210713/42900490)
* [test log from 25](http://paste.fedoraproject.org/210714/14290049)

Here goes the complete [testsuite.log](http://paste.fedoraproject.org/210717/90050441).

pjaaskel:
Have you tried the latest pocl master? I think there was some work on images recently which might have fixed this.

cjacker:
get_image_width/height... failed with llvm/clang 3.6.0/3.6.1/3.6.2/3.7.0rc1.

pjaaskel:
Not a regression (if broken also in 0.11) and hard to reproduce. Found a system where these two tests fail, but don't know why.

pjaaskel:
...and that system got fixed with a clean build too.

gerddie:
I hit the same error when compiling the version 0.11 within a virtual machine running Debian/sid, target x86_64. llvm an clang versions are 3.6.2. 

The build configuration can be obtained from 

    git://anonscm.debian.org/collab-maint/pocl.git

gerddie:
[...]

PR #271, opened 2015-10-31
Fix checks for images/samplers when parsing kernel metadata.

jrprice:
This should fix the test failures listed in #181. The issue was that image and sampler arguments were not being correctly detected (they were being mislabeled as regular pointer arguments).


pjaaskel:
Thanks James.

PR #274, opened 2015-11-05
Fix getprograminfo kernelnames

guicho271828:


guicho271828:
doesn't it have a continuous testing?

pjaaskel:
Yes we have buildbots but we don't test each API separately (we will rely on the conformance tests on that) but have larger "smoke tests" (external OpenCL apps) mainly.  For regressions we often add smaller unit tests (please do, BTW).

guicho271828:
copy from opencl-2.0 manual:

```
param_value is a pointer to memory where the appropriate result being queried is returned. If
param_value is NULL, it is ignored.
param_value_size is used to specify the size in bytes of memory pointed to by param_value.
This size must be >= size of return type as described in table 5.17.
param_value_size_ret returns the actual size in bytes of data copied to param_value. If
param_value_size_ret is NULL, it is ignored.
```

guicho271828:
BTW, this error was found during the continuous tests of a library I am writing for Common Lisp (also has CL abbreviation here :) ). It thoroughly tries to query the all possible values that can be accessed through opencl. It worked on most platforms, but not with POCL. With the private branch of mine, all the tests passed. Yey!
https://circleci.com/gh/guicho271828/eazy-opencl/39
https://github.com/guicho271828/eazy-opencl

pjaaskel:
OK, looking forward to your merge requests :)

pjaaskel:
After the style is cleaned, I can merge this in.

guicho271828:
Do you mean the entire code in the same file?
I think 9012691 had fixed the style of the portion I was concerned.

guicho271828:
From the specification, it is unclear if the resulting string should be null-terminated or not. But it seems the original author intend to append a null byte, so I'll revert the first part of the pullreq (string size was correct).

guicho271828:
When I tried c-file-style : "gnu" using emacs, I noticed that the indentation in the switch statement is incorrect. Fix to this will re-indent almost everything in this file. I'm wondering should I really do this...

guicho271828:
I reconsidered the behavior of this function, and realized that the problem was not merely about the null terminator.

Previous code stops writing remaining kernel names when the array boundary (specified by `param_value_size`) is reached: e.g., when `param_value_size` = 5 and there are 2 kernels named `foo` and `bar`, it sets `foo` to `param_value`, 3 to `param_value_size_ret` and returns `CL_SUCCESS`.

However, another reasonable behavior that can be interpreted from the spec is this: set `param_value_size_ret` to 7 (for `"foo;bar"`), and return `CL_INVALID_VALUE` because the size of `&param_value` suggested by `param_value_size` is insufficient for storing the result, according to its specification:

> CL_INVALID_VALUE if param_name is not valid, or if size in bytes specified by param_value_size is < size of return type as described in table 5.17 and param_value is not NULL.

This is a conflict & error in the specification. How should I handle this? What do you think?

Also, there is a little ambiguity in the spec. The spec says 

> param_value_size_ret returns the actual size in bytes of data *copied* to param_value.

so it sounds as if the implementation can stop copying the string at an arbitrary point where the limit is reached. However, in the specifications of other functions, it uses another definition:

> param_value_size_ret returns the actual size in bytes of data
> being *queried* by param_value.

guicho271828:
@vkorhonen

lib/CL/clGetProgramInfo.c
```diff
@@ -101,26 +101,71 @@ POname(clGetProgramInfo)(cl_program program,
...
+      size += 1;                /* a NULL */
+      
       if (param_value_size_ret)
-        *param_value_size_ret = size;      
+        *param_value_size_ret = size;
+
+      if (param_value)
+        {
+          /* only when param_value is non-NULL */
+          if ( size > param_value_size )
```
pjaaskel: Extra white spaces here. Pls check http://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting

lib/CL/clGetProgramInfo.c
```diff
@@ -101,26 +101,71 @@ POname(clGetProgramInfo)(cl_program program,
...
+            return CL_INVALID_VALUE;
+          /* should not break from the switch clause because of POCL_ABORT_UNIMPLEMENTED */
+
+          for (i = 0; i < num_kernels; ++i)
+            {
+              if (i == 0)
+                strcpy(param_value, kernel_names[i]); /* copy including NULL */
```
pjaaskel: Too few white spaces before the argument list. Should be strcpy (...

pjaaskel:
OK. If there's ambiguity in the specs then it's up to the conformance tests to complain :) Sorry to nitpick about the style, but I see it important to get the style right in new committers' first merge requests and also to try to keep the style of the code base improving, not degrading.

guicho271828:
I'm very thankful for the comments on the style, I learned new things anyway. I'd be more careful about trailing whitespaces.
Also, sending a fix to a base project like this is a rare opportunity for me. Especially C projects. I don't write C usually.

PR #276, opened 2015-11-09
SVM work

franz:
This implements

* clEnqueueSVM{Map,Unmap,Memcpy,MemFill,Free}
* clSVMAlloc , clSVMFree
* clSetKernelArgSVMPointer, clSetKernelExecInfo (the latter pretty much empty)
* clCreatePipe, clGetPipeInfo (only the API calls, pipes require much more work)
* clGetDeviceInfo - add 2.0 properties
* cl.h, cl*.h - update headers to 2.0

non-SVM related: clEnqueueFillBuffer

Issues
1. maybe we should still advertise OpenCL 1.2 despite having 2.0 headers ?
2. SVM indirect pointers work only by bitcasting to ulong (structs don't work, and uintptr_t is 32bits = broken)

Tested:
1) the usual autotools / cmake build + make check
2) intel svm examples SVMBasicCoarseGrained, SVMBasicFineGrained (issue 2)

SVM calls on CPU device are untested, FillBuffer is untested (will update the PR)


lib/CL/clEnqueueFillBuffer.c
```diff
@@ -0,0 +1,113 @@
...
+  if (errcode != CL_SUCCESS)
+    return errcode;
+
+  /* CL_INVALID_VALUE if pattern is NULL or if pattern_size is 0
+   * or if pattern_size is not one of {1, 2, 4, 8, 16, 32, 64, 128}. */
+  POCL_RETURN_ERROR_COND((pattern == NULL), CL_INVALID_VALUE);
+  POCL_RETURN_ERROR_COND((pattern_size == 0), CL_INVALID_VALUE);
+  POCL_RETURN_ERROR_COND((pattern_size > 128), CL_INVALID_VALUE);
+
+  POCL_RETURN_ERROR_ON((__builtin_popcount(pattern_size) > 1), CL_INVALID_VALUE,
```
pjaaskel: Is this builtin portable? Can you try avoiding it or write a pocl-only helper?

lib/CL/clEnqueueSVMFree.c
```diff
@@ -0,0 +1,86 @@
...
+                                                       cl_uint num_svm_pointers,
+                                                       void *svm_pointers[],
+                                                       void  *user_data),
+                  void *user_data,
+                  cl_uint num_events_in_wait_list,
+                  const cl_event *event_wait_list,
+                  cl_event *event) CL_API_SUFFIX__VERSION_2_0
+{
+#ifndef BUILD_HSA
+  POCL_MSG_PRINT_INFO("This pocl was not built with HSA\n");
```
pjaaskel: Why require HSA for SVM?

lib/CL/clEnqueueSVMMap.c
```diff
@@ -0,0 +1,91 @@
...
+                 cl_bool blocking_map,
+                 cl_map_flags map_flags,
+                 void *svm_ptr,
+                 size_t size,
+                 cl_uint num_events_in_wait_list,
+                 const cl_event *event_wait_list,
+                 cl_event *event) CL_API_SUFFIX__VERSION_2_0
+{
+#ifndef BUILD_HSA
+  POCL_MSG_PRINT_INFO("This pocl was not built with HSA\n");
```
pjaaskel: Same

lib/CL/clEnqueueSVMMemFill.c
```diff
@@ -0,0 +1,94 @@
...
+                     void *svm_ptr,
+                     const void *pattern,
+                     size_t pattern_size,
+                     size_t size,
+                     cl_uint num_events_in_wait_list,
+                     const cl_event *event_wait_list,
+                     cl_event *event) CL_API_SUFFIX__VERSION_2_0
+{
+#ifndef BUILD_HSA
+  POCL_MSG_PRINT_INFO("This pocl was not built with HSA\n");
```
pjaaskel: ditto

lib/CL/clEnqueueFillBuffer.c
franz: Ahh yes. GCC definitely has a fallback implementation, so it works with GCC on every arch, but im not sure about clang & msvc. Will look into it

franz: OK, i've checked that both clang & gcc supports it, also added an MSVC macro. Is that good enough ?

pjaaskel:
basic.c:774:7: error: ‘for’ loop initial declarations are only allowed in C99 mode for (size_t i = 0; i < size; i++)  seems we don't need the --std=c99 at the moment and it fails to build with gcc 4.8.4 default C standard (gnu90). Better just fix this instead of starting to add --std switches.

```
./../../../../lib/CL/pocl_icd.h:175:50: warning: "/*" within comment [-Wcomment] NULL, /*&POclCreateCommandQueueWithProperties, /* &clUnknown109 */      \
```
```
clGetPlatformIDs.c:33:8: warning: excess elements in struct initializer [enabled by default]
```

Probably there's now a problem when trying to build against 1.2 ICD headers / ICD loader. Better retain support for this so users can enable 1.2 support even without a 2.0 loader.

PR #279, opened 2015-11-26
Refactor clEnqueueCopyBufferRect and clEnqueueCopyImage

Oblomov:
clEnqueueCopyImage used to call clEnqueueCopyBufferRect, which couldn't
work because the latter would check that the source and destination
were actual images.

Refactor the code into a separate function, with extra parameters to
know which should be a image and which a buffer, and copy it from both
the API calls instead.


lib/CL/pocl_img_buf_cpy.c
```diff
@@ -0,0 +1,199 @@
+/* pocl_img_buf_cpy.c: common parts of image and buffer copying
```
pjaaskel: Pls check the style in this file especially. http://portablecl.org/docs/html/development.html#coding-style

lib/CL/pocl_img_buf_cpy.c
```diff
@@ -0,0 +1,199 @@
...
+      " but the given dst & src row pitch differ\n")
+
+    POCL_RETURN_ERROR_ON((check_copy_overlap(mod_src_origin, mod_dst_origin, mod_region,
+      src_row_pitch, src_slice_pitch)), CL_MEM_COPY_OVERLAP, "src and "
+      "dst are the same object, and source and destination "
+      "regions overlap");
+
+  }
+
+  POCL_CHECK_DEV_IN_CMDQ
```
pjaaskel: Could this macro made such that it allows adding a semicolon at the end of the line. I have had problems with editors getting confused and breaking auto indent after seeing this type of statements.

PR #280, opened 2015-11-27
Mark failing tests on MIPS with WILL_FAIL on CMake.

kraiskil:
Just to keep the buildbot green. 
I don't have a mips to test on, could I trouble @dsandersimgtec to test/approve this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -878,6 +878,8 @@ add_subdirectory("scripts")
 # for tests & examples
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc")
   set(POWERPC 1)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "mips")
+  set(MIPS 1)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "armv7")
   set(ARMV7 1)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "armv6")
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -57,6 +57,6 @@ add_subdirectory("scalarwave")
 add_subdirectory("trig")
 
 if(MIPS)
-  set_tests_properties("examle1-spir32"
+  set_tests_properties("example1-spir32"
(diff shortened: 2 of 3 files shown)


kraiskil:
Ok, thanks.

I leave the decision of updating the buildslave to you - whenever is convenient for you. 
The idea here is just to get the buildslave to switch to green, so developers notice easier if they cause new regressions. Marking "unexpected success"es is easy enough, even without access to a mips machine.

dsandersimgtec:
I've updated the buildslave so the majority of these should be fixed in the next build. I'll update it again once the final release is out.

This might actually make the buildbot green by itself since the remaining known failure was kernel/test_hadd_loops and buildbot says that's passing in build 43.

kraiskil:
Great.
I'd still really want the line:
```elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "mips")```
tested & into the master. Even if there are no more failures :)

dsandersimgtec:
Ok

dsandersimgtec:
Updating the compiler to 3.7.1rc2 made the buildbot green so we don't need the WILL_FAIL parts of this change any more.

Just for completeness, cmake emitted some errors for the set_properties statements. It turned out that example1-spir32 should have been spec_tests/example1_dot_product_spir32, and the kernel/test_shuffle_* tests were being marked WILL_FAIL before the relevant add_test. I just deleted these statements in my testing since I'd already seen the buildbot turn green.

> elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "mips")
>  set(MIPS 1)

This works correctly so we should go ahead this this bit.

pjaaskel:
Shall I pull this?

dsandersimgtec:
I think it needs to be revised to just the:
> elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "mips")
>   set(MIPS 1)

The parts to mark expected failures has been made unnecessary by the compiler update and will cause unexpected passes.

kraiskil:
Right - lets not merge this request, just add the above lines to master. Thanks for testing @dsandersimgtec.

pjaaskel:
Added in 2136936b6.

PR #281, opened 2015-11-30
OpenCL 2.0 (SVM) Atomics

franz:
Implements (a subset of) C11-style atomic operations for Host and HSA devices.

Had to bump the device/pocl OpenCL version to 2.0 (clang doesn't recognize the atomic types unless it's given -cl-std=CL2.0 option). While doing that, i fixed some Host-specific code which in fact shouldn't depend on Host at all (like pocl-features.h).

Also added clCreateCommandQueueWithProperties(), required for CL2.0 samples to work.

Some problems that remain unsolved:
* Host atomics: clang builtins for min/max atomics exist only for int and uint. I have added an .ll file for x86_64 that directly uses atomicrmw, but all other archs are unlucky (for now).
* HSAIL code: ignores scope argument, always uses "system" - this is a shortcoming of the atomicrmw instruction
* atomic_work_item_fence missing
* HSA atomics seems to work only with max workgroup size 64, reason unknown. HSA runtime reports max wg size 1024; clinfo with proprietary amd drivers reports 256, but 256 does not work either.
* Bumping CL version to 2.0 might have broken TCE build (did not test)

Several new AMD SDK 3.0 examples work: CalcPie, FineGrainSVMCAS (FineGrainSVM does not, or maybe it's just taking way too long)


include/_kernel.h
```diff
@@ -32,6 +32,30 @@
 #endif
 #include "_kernel_c.h"
 
+/* If the -cl-std build option is not specified, the highest OpenCL C 1.x
+ * language version supported by each device is used as the version of
+ * OpenCL C when compiling the program for each device.
```
pjaaskel: I think it's OK to have different OpenCL version devices in the same platform. The devices do not have to match the host OpenCL version.

lib/CL/devices/hsa/pocl-hsa.c
```diff
@@ -363,9 +362,9 @@ pocl_hsa_init_device_infos(struct _cl_device_id* dev)
 
   uint16_t wg_sizes[3];
   HSA_CHECK(hsa_agent_get_info(agent, HSA_AGENT_INFO_WORKGROUP_MAX_DIM, &wg_sizes));
-  dev->max_work_item_sizes[0] = wg_sizes[0];
-  dev->max_work_item_sizes[1] = wg_sizes[1];
-  dev->max_work_item_sizes[2] = wg_sizes[2];
+  dev->max_work_item_sizes[0] = 64; //wg_sizes[0];
```
pjaaskel: Why the 64 magic number? Is this intentional?

lib/CL/devices/hsa/pocl-hsa.c
```diff
@@ -384,9 +383,12 @@ pocl_hsa_init_device_infos(struct _cl_device_id* dev)
   dev->max_clock_frequency = 700;
 #endif
 
-
+/*
   HSA_CHECK(hsa_agent_get_info
     (agent, HSA_AGENT_INFO_WORKGROUP_MAX_SIZE, &dev->max_work_group_size));
+*/
+  dev->max_work_group_size = 64;
```
pjaaskel: Same here. 64 is rather little.

pjaaskel: OK. Noted your PR comment. Please add a comment to the file explaining this.

pjaaskel:
There are no new tests added in this commit even though you claim new ones are passing?

include/_kernel.h
franz: Yep, it is OK. That quote is directly from: https://www.khronos.org/registry/cl/sdk/2.0/docs/man/xhtml/preprocessorDirectives.html

Now that i think of it, i took care of this particular thing in pocl_llvm_api.cc so the ifndef is never going to happen.

franz:
The new AMD SDK passing tests are from AMD SDK 3.0, they don't exist in 2.9 so i'd have to import SDK 3.0 into pocl.

pjaaskel:
Yes.

franz:
I pushed a partial fix for TCE, but it's still not working for me (and i have no idea if it's a problem with my system, or this PR). pocl compiles but running tests fail on compilation. Can you try with TCE ?

PR #293, opened 2016-01-22
Various fixes2

franz:
Main items:
* LLVM 3.8 support
* pocl bug / llvm 3.8 crash on atomics
* new CMake tests

still TODO items: 
* someone has to fix WI-AA to actually work


franz:
Should note that most of the tests with 3.8 are failing, message is:
```
kernel: /home/unused_LLVM/LLVM_git/lib/IR/Constants.cpp:1531: static llvm::Constant* 
llvm::ConstantExpr::getCast(unsigned int, llvm::Constant*, llvm::Type*, bool): 
Assertion `CastInst::castIsValid(opc, C, Ty) && "Invalid constantexpr cast!"' failed.
```

franz:
Added CMake tests for TCE. They seem to fail in the exact same place as with autotools (fp16 tests)
Currently these tests have the tce label: tests/tce/{fp16,tcemc,ttasim} , examples/example{1,2,2a}

I've also updated tools/script/run_{hsa,tta}_tests to automatically select cmake tests if you run them from the builddir.

franz:
OK so currently these tests fail:
```
	  2 - kernel/test_as_type (OTHER_FAULT)
	 19 - kernel/test_local_struct_array (OTHER_FAULT)
	 67 - spec_tests/example1_dot_product_spir64 (OTHER_FAULT)
```
SPIR is currently broken in LLVM 3.8, 
i haven't looked into test_local_struct_array yet,
and kernel/test_as_type probably has problems with generic addrspace.

franz:
Current status:
* generic AS is converted to private
* SPIR example is fixed (but only works on x86_64 and i386)
* test_local_struct_array has been broken for a while now, not a regression
* test_as_type should work now

I'm going to test with 3.7 & 3.8 & autotools now, if they pass i think it's ready to merge.

franz:
3.7 HSA / 3.7 TCE / 3.8 tested with both CMake and autotools, two minor bugs found & fixed. I'll do some more extensive testing with the external testsuites

PR #299, opened 2016-02-11
Pocl proper binary format : poclcc

rjodin:
**POCL PROPER BINARY FORMAT**

- poclcc binary format:

[char *_pocl_id_][uint32_t _version_][uint64_t _device_id_][uint32_t _num_kernels_]
[
--[uint32_t _sizeofKernelName_][uint32_t _num_args_][uint32_t _num_locals_][uint32_t _sizeofBinary_]
--[char *_kernel_name_][struct pocl_argument *_dyn_arguments_][struct pocl_argument_info *_arg_info_]
--[unsigned char *_binary_]
]+

- binary informations:

_pocl_id_: needed to make the difference between poclcc binary and LLVM IR binary (in _clCreateProgramWithBinary_ at least)
_version_: to identify with version of poclcc has been use to generate the file
_device_id_: It is a number created by hashing the _vendor_id_, _llvm_target_triplet_ and _llvm_cpu_ of the _cl_device_id_ (with the FNV-1A algorithm). It enables pocl to know for which device a kernel has been compiled
_num_kernels_: number of kernels contained in the file for this device

- kernel informations:

_sizeofKernelName_: the size of _kernel_name_ (with does not have a '\0' at the end)
_num_args_: number of arguments of the kernel. We need it as we won't be able to find the information in the binary and _clCreateKernel_ needs it.
_num_locals_: same as _num_args_ but for local variables
_sizeofBinary_: the size of the binary (ELF file)
_kernel_name_: the name of the kernel
_dyn_arguments_: table needed by _clCreateKernel_
_arg_info_: table needed by _clCreateKernel_
_binary_: the binary for this kernel (and this device only) 

**CHANGES FOR AN OPENCL USER:**

- bin _poclcc_: It takes some arguments (./poclcc --help) to produce a file which contains everything necessary to run an OpenCL program without having to make a call to a compiler. (pocl-standalone has been removed)

**CHANGES FOR AN POCL DEVELOPER**

- add of arguments in cl_program (_pocl_binaries_, _pocl_binary_sizes_, _is_pocl_binary_).

- When running a kernel created from _clCreateProgramWithBinary_ the backend needs to set up the _local_size_ tab in the pocl_context.

- Everything that deal with the new binary format are in _lib/CL/pocl_binary.{c,h}_.

**TESTSUITE**

There is a new test in _tests/regression/_ named _test_enqueue_kernel_from_binary_.
It compares the run of the same kernel created with _clCreateProgramWithBinary_ and _clCreateProgramWithSource_.
(It also compares with a third kernel create with _clCreateProgramWithBinary_ but with different local_size to check that the loops on local_size are efficient).


include/CL/cl.h
```diff
@@ -1375,6 +1375,18 @@ clEnqueueTask(cl_command_queue  /* command_queue */,
...
+clExportPoclccBinary(cl_program /* program */,
+                     void **    /* poclcc_binary */,
+                     cl_uint *  /* poclcc_binary_size */);
+  
+extern CL_API_ENTRY cl_int CL_API_CALL
+clExtractPoclccBinary(void *            /* input_buffer */,
+                      cl_uint           /* input_buffer_size */,
+                      size_t **         /* lengths */,
+                      unsigned char *** /* binaries */);
+
```
clementleger: cl.h should not be modified as it's provided by Khronos. 
An extension file should be added for pocl in cl/cl_pocl.h

include/CL/cl.h
```diff
@@ -1375,6 +1375,18 @@ clEnqueueTask(cl_command_queue  /* command_queue */,
               const cl_event *  /* event_wait_list */,
               cl_event *        /* event */) CL_EXT_SUFFIX__VERSION_1_2_DEPRECATED;
 
+/* Poclcc OpenCL APIs */
+extern CL_API_ENTRY cl_int CL_API_CALL
+clExportPoclccBinary(cl_program /* program */,
```
pjaaskel: What is the purpose of these new API functions? First, we cannot add them to cl.h nor with direct cl-prefix (those are preserved for official OpenCL APIs), second, why don't the clGetProgramInfo() suffice?

scripts/poclcc/poclcc.c
```diff
@@ -0,0 +1,179 @@
+#include <stdio.h>
```
pjaaskel: Can you move this to bin/poclcc.c  As this should replace pocl-standalone (the last script we have), it could be removed and the scripts directory removed as well.

pjaaskel:
Thanks, I'll give it a more thorough review with better time.

include/CL/cl.h
rjodin: These functions are just helpers.

I don't know how you think the binary format will be used... I chosed to have a binary format with (possibly) several device in it. But the _clGetProgramInfo_ gives a tab of devices. So I made those functions to put devices together and to add few other info (pocl identifier, version number).

Maybe I won't be usefull to have binaries for severals devices in the binary format. In this case, those function wouldn't have any purpose. And we would just have to add the version number in the device format (as it is not present at the moment).

scripts/poclcc/poclcc.c
rjodin: ok

include/CL/cl.h
pjaaskel: OK, I see. I didn't realize it before that they are separated in the APIs. Better make the binary format such that it works directly and fully with clGetProgramInfo and clCreateProgramWithBinary, thus one binary blob per device. It might get complicated otherwise.

rjodin: Ok.
So I will remove those functions. And add what is needed in the device format (which will become the poclcc binary format).

lib/CL/pocl_cl.h
```diff
@@ -532,6 +533,10 @@ struct _cl_program {
      sequential bitcode produced from the kernel sources.*/
   size_t *binary_sizes; 
   unsigned char **binaries; 
+  /* Poclcc binary format*/
+  unsigned isPoclccBinary;
```
clementleger: This should probably be lowercased with '_'

include/pocl.h
```diff
@@ -242,6 +242,7 @@ typedef struct _cl_command_node_struct
   const cl_event *event_wait_list;
   cl_uint num_events_in_wait_list;
   cl_device_id device;
+  cl_uint isPoclccBinary;
```
clementleger: This should probabnly not be here but only in program since it's possible to get the program from the command node.

scripts/poclcc/poclcc.c
rjodin: There is no _bin_ folder at the moment, should I create one? or just let _poclcc.c_ in the script folder and clean every trace of pocl-standalone?

pjaaskel: Yes, please create a new 'bin' folder and remove the obsolete 'scripts' folder. I think we can integrate all functionality of pocl-standalone to the new poclcc binary, right?

bin/poclcc.c
```diff
@@ -0,0 +1,188 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <CL/opencl.h>
+#include <assert.h>
+#include <string.h>
+
+char *kernelSource = NULL;
```
pjaaskel: Please check variable naming convention. kernel_source etc.

bin/poclcc.c
```diff
@@ -0,0 +1,188 @@
...
+    printf("\t\tBuild the program with <options> options\n");
+    printf("\t-d <device_type>\n");
+    printf("\t\tSelect <device_type> as the device_type for clGetDeviceIDs." 
+           "Default: CL_DEVICE_TYPE_DEFAULT\n");
+    printf("\t-o <file>\n");
+    printf("\t\tWrite output to <file>\n");
+    return -1;
+}
+
+int poclcc_error(char *msg){
```
pjaaskel: Please check the C style (gnu).

bin/poclcc.c
```diff
@@ -0,0 +1,188 @@
...
+
+    return buffer;
+}
+
+int process_kernel_file(int arg, char **argv, int argc){
+    if (arg >= argc)
+        return poclcc_error("Incomplete argument for input file!\n");
+
+    char *filename = argv[arg];
+    char *ext = ".pocl";
```
pjaaskel: We don't want to force an extension to the filename.

bin/poclcc.c
```diff
@@ -0,0 +1,188 @@
...
+char *build_options = NULL;
+
+int print_help(){
+    printf("USAGE: poclcc [OPTION]... [FILE]\n");
+    printf("\n");
+    printf("OPTIONS:\n");
+    printf("\t--build_options=<options>\n");
+    printf("\t\tBuild the program with <options> options\n");
+    printf("\t-d <device_type>\n");
+    printf("\t\tSelect <device_type> as the device_type for clGetDeviceIDs."
```
pjaaskel: It might be better to add selection to the pocl device id (starting from 0) and default to the first found device. In the future there might be multiple GPUs in the same platform etc.

examples/poclcc/CMakeLists.txt
```diff
@@ -33,12 +33,11 @@
 #clean-local:
 # rm -f $(EXTRA_DIST:.cl=.bc) $(EXTRA_DIST:.cl=.h)
 
-add_test("pocl-standalone" "/bin/sh" "${CMAKE_BINARY_DIR}/scripts/pocl-standalone"
-          -h "${CMAKE_BINARY_DIR}/standalone.h"
-          -o "${CMAKE_BINARY_DIR}/standalone.bc"
-          "${CMAKE_CURRENT_SOURCE_DIR}/standalone.cl")
+add_test("poclcc" "/bin/sh" "${CMAKE_BINARY_DIR}/bin/poclcc"
+          -o "${CMAKE_BINARY_DIR}/poclcc.c.pocl"
```
pjaaskel: Should it be poclcc.cl.pocl?

lib/CL/clCreateKernel.c
```diff
@@ -45,6 +46,7 @@ POname(clCreateKernel)(cl_program program,
   int errcode;
   int error;
   unsigned device_i;
+  poclcc_global poclcc;
```
pjaaskel: What is this?

lib/CL/clCreateProgramWithBinary.c
```diff
@@ -123,14 +124,51 @@ POname(clCreateProgramWithBinary)(cl_context                     context,
   program->kernels = NULL;
   program->build_status = CL_BUILD_NONE;
 
+  /* We do no want a binary with different kind of binary in it
```
pjaaskel: Please add a section to the user manual that describes the different supported binary input formats.

lib/CL/clGetProgramInfo.c
```diff
@@ -25,6 +25,157 @@
...
+      fclose(f);
+      
+      pocl_cache_release_lock(lock);
+      
+      //ALLOC KERNEL INFOS
+      char *kernel_name;
+      struct pocl_argument *dyn_arguments;
+      struct pocl_argument_info *arg_info;
+      
+      int sizeofKernelName = strlen(kernel[i]->name);
```
pjaaskel: Check naming. Also there are some style issues here too.

lib/CL/clGetProgramInfo.c
```diff
@@ -25,6 +25,157 @@
 #include "pocl_llvm.h"
 #include "pocl_util.h"
 #include "pocl_cache.h"
+#include "poclcc_binary.h"
+
+
+cl_int compile_kernels(cl_program program, cl_device_id device, poclcc_device *devicecc)
```
pjaaskel: This function requires a comment about what does it compile and why.

lib/CL/pocl_llvm_api.cc
```diff
@@ -1238,6 +1238,7 @@ namespace pocl {
     extern size_t WGLocalSizeX;
     extern size_t WGLocalSizeY;
     extern size_t WGLocalSizeZ;
+    extern size_t isKernelWithDynamicLocalSize;
```
pjaaskel: Should be 'bool WGDynamicSize;' or similar?

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
+#include "poclcc_binary.h"
```
pjaaskel: This new file requires the license and copyright heading.

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
...
+
+void poclcc_free(poclcc_global *binary_format){
```
pjaaskel: Better name might be 'pocl_binary_free' and poclcc_global should be simply pocl_binary ?

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
...
+    if (binary_format->devices != NULL)
+      {
+        int i;
+        for (i=0; i<binary_format->num_devices; i++)
+          poclcc_device_free(&(binary_format->devices[i]));
+        POCL_MEM_FREE(binary_format->devices);
+      }
+}
+
+void poclcc_device_free(poclcc_device *device){
```
pjaaskel: Then this also should be pocl_binary_device? poclcc is the binary for building kernels to pocl_binaries (like nvcc in NVIDIA).

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
...
+int poclcc_check_device(poclcc_device *device){
+  if (device->version != POCLCC_VERSION)
+    return 0;
+  return !strncmp(device->pocl_id, POCLCC_STRING_ID, POCLCC_STRING_ID_LENGTH);
+}
+
+#define FNV_OFFSET UINT64_C(0xcbf29ce484222325)
+#define FNV_PRIME UINT64_C(0x100000001b3)
+uint64_t poclcc_get_device_id(cl_device_id device){
+  //FNV-1A with vendor_id, llvm_target_triplet and llvm_cpu
```
pjaaskel: Is there a need to hash these? What about just appending them together? It would be more readable in the contents?

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
...
+  int size = 0;
+  for (i=0; i<num_devices; i++)
+    size += binaries_sizes[i];
+
+  size += sizeof(poclcc_global) - sizeof(poclcc_device *);
+  return size;
+}
+
+
+void copyKernel2Buffer(poclcc_kernel *kernel, char **buf){
```
pjaaskel: naming

lib/CL/poclcc_binary.c
```diff
@@ -0,0 +1,443 @@
...
+  buffer += sizeofDynArgs;
+
+  memcpy(buffer, kernel->arg_info, sizeofArgInfo);
+  buffer += sizeofArgInfo;
+
+  memcpy(buffer, kernel->binary, sizeofBinary);
+  buffer += sizeofBinary;
+  *buf = buffer;
+}
+int copyBuffer2Kernel(char **buf, poclcc_kernel *kernel){
```
pjaaskel: Ditto.

lib/CL/poclcc_binary.c
pjaaskel: Should this be pocl_binary_add_kernel_data() or similar?

lib/CL/poclcc_binary.h
```diff
@@ -0,0 +1,86 @@
+#ifndef POCL_BINARY_FORMAT_H
```
pjaaskel: Needs the license+copyright heading.

bin/poclcc.c
rjodin: How will we know which id will be the device we what the binary for?
Maybe we could extract a binary per device found, and add something in the name of the binary to specify for which device the binary i?

lib/CL/clCreateKernel.c
rjodin: This is a temporary variable needed for the functions that will get the binary of the kernel to be created.
I didn't put it in the _"if"_ statement where it is used because it needs to be free is case of error. 
If you want if to be in the _"if"_ statement, I will remove the _POCL_GOTO_ERROR_COND_ and manage error (part of it) in the _"if"_ statement.

lib/CL/poclcc_binary.c
rjodin: What we dump in the binary file is the _poclcc_device_ struct.
Maybe we don't what this struct to be named _poclcc_global_ but I'm not sure that _pocl_binary_ matchs it well.
Maybe something like _poclcc_binaries_ but it could still be confusing...

lib/CL/poclcc_binary.c
rjodin: I wanted something of a fix size to be simple. But it can be the concatenation of them, so that it will be more readable in the contents yes.

lib/CL/poclcc_binary.c
rjodin: I'm not sure... This is a translation from the struct (with pointers) to a continuous buffered form of the structure (without pointers).

bin/poclcc.c
pjaaskel: Sounds good.

lib/CL/clCreateKernel.c
pjaaskel: Fine. Just a better type name is needed.

lib/CL/poclcc_binary.c
pjaaskel: Ok then poclcc_device needs a better name too.

lib/CL/poclcc_binary.c
pjaaskel: OK. Then something like pocl_binary_serialize_kernel() or similar.

lib/CL/clCreateProgramWithBinary.c
rjodin: Where is the user manual?

pjaaskel: doc/sphix/source

rjodin: Should I add something under "Design Notes", or should I create a new section?

pjaaskel: Yes, design notes looks like an appropriate location for now.

rjodin: I added a section in design notes about the binary inputs formats allowed by pocl.
See commit : d7f688652df2fa1f21ae955d69d6516f8ff4cdda

vkorhonen:
Does this still produce static local_size workgroup functions when clCreateProgramWithSource is used?

rjodin:
@vkorhonen : Yes, it doesn't change anything when you're using clCreateProgramWithSource.

pjaaskel:
@rjodin does it include the static WG ones in the binary, if they have happened to be generated previously?

rjodin:
@pjaaskel I'm not sure to understand...
If you're using clEnqueueNDRange with a program created with clCreateProgramWithSource, it uses a static WG binary.
If you're using clEnqueueNDRange with a program created with clCreateProgramWithBinary, it uses the binary you've given to it.
When you call clGetProgramInfo with _param_name=CL_PROGRAM_BINARIES_, you get a binary with a dynamic WG binary (no static WG version is include in it, even if you've called clEnqueueNDRange with a program created with clCreateProgramWithSource before).

pjaaskel:
OK. In my original plan I envisioned this could happen implicitly due to just packing up the temp dir contents, but we can implement it later. It can be useful for performance purposes to have static WG functions included also, if they have been generated for interesting local sizes.

franz:
OK so just a few things in no particular order:

1. could you please fix the warnings that appear with -Wall (your code only, not all warnings in pocl ;)
2. CMakeLists.txt is missing from bin/, here's one i quicky put together:
   ```
      add_executable(poclcc poclcc.c)

      target_link_libraries(poclcc ${OPENCL_LIBS})

      install(TARGETS "poclcc" RUNTIME
           DESTINATION "${POCL_INSTALL_PUBLIC_BINDIR}")
   ```
3. any particular reason you didn't use getopt() or some other option parsing library for poclcc ? (BTW you could have used C++ for poclcc, but i guess doesn't matter now)
4. the binary format contains garbage. The problem is you're memcpy()-ing pocl_binary and other structs directly to the output buffer AFAICT. The C compiler is free to chose any alignment it wants for the struct members (and struct itself), so
```
  uint32_t version;
  uint64_t device_id;
  uint32_t num_kernels;
```
on x86_64 results in 32bits of garbage after ```version``` and ```num_kernels``` - and the resulting binary is also machine-dependent (and incompatible with 32bit machines). If you want to memcpy structs, you must force the alignment or use packed structs.

rjodin:
@franz 
I take care of your comment in this commit : 9a94ae9c0cac79a6866877fd6bc13cb8a5822b61

About poclcc, I wasn't aware of "getopt()". Maybe I'll change poclcc, but I will not have the time for it soon.

PR #309, opened 2016-03-14
POCL without llvm

rjodin:
Works for both build configuration (autogen + configure / cmake)

Every changes are under the macro OCS_AVAILABLE (online compiler support)

The offline compiler support version of POCL works only on Linux as I'm not using LLVMFileUtils.cc.


lib/CL/clCreateKernel.c
```diff
@@ -76,8 +78,10 @@ POname(clCreateKernel)(cl_program program,
       if (program->binaries[device_i] &&
           pocl_cache_device_cachedir_exists(program, device_i))
         {
+#ifdef OCS_AVAILABLE
           pocl_llvm_get_kernel_metadata(program, kernel, device_i,
                                         kernel_name, &errcode);
+#endif
```
pjaaskel: Shouldn't it fetch the kernel metadata from the binary dump?

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -241,9 +241,13 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
   int realdev_i = pocl_cl_device_to_index(kernel->program, realdev);
   if (kernel->program->source || kernel->program->binaries[realdev_i])
     {
+#ifdef OCS_AVAILABLE
       error = pocl_llvm_generate_workgroup_function(realdev,
                                                     kernel,
                                                     local_x, local_y, local_z);
+#else
+      error = 1;
```
pjaaskel: Why it errors here?  Doesn't it fallback to the dynamic WG size kernel inside the binary?

lib/CL/devices/basic/basic.c
```diff
@@ -251,8 +251,9 @@ pocl_basic_init_device_infos(struct _cl_device_id* dev)
   dev->native_vector_width_double = POCL_DEVICES_PREFERRED_VECTOR_WIDTH_DOUBLE;
   dev->native_vector_width_half = POCL_DEVICES_PREFERRED_VECTOR_WIDTH_HALF;
   dev->max_clock_frequency = 0;
+#ifdef OCS_AVAILABLE
```
pjaaskel: Better have some kind of default here.

lib/CL/pocl_cache.c
```diff
@@ -332,6 +332,9 @@ int pocl_cache_write_kernel_parallel_bc(void*        bc,
                                         size_t       local_x,
                                         size_t       local_y,
                                         size_t       local_z) {
+#ifndef OCS_AVAILABLE
```
pjaaskel: Is this really needed? Shouldn't the caller be disabled?

pjaaskel: Shouldn't the caller be disabled instead?

lib/CL/pocl_file_util.c
```diff
@@ -0,0 +1,202 @@
...
+#include <string.h>
+#include <dirent.h>
+
+#include "pocl.h"
+#include "pocl_file_util.h"
+
+/* #define to disable locking completely */
+#undef DISABLE_LOCKMANAGER
+
+// using namespace llvm;
```
pjaaskel: remove commented out code

lib/CL/pocl_file_util.c
```diff
@@ -0,0 +1,202 @@
...
+
+/*****************************************************************************/
+
+int
+pocl_rm_rf(const char* path) {
```
pjaaskel: Check the style of a C file. Should be GNU.

lib/CL/clCreateKernel.c
rjodin: If you're creating a kernel form a program create with binary, it won't go in this section at all.
But as this function (pocl_llvm_get_kernel_metadata) doesn't exist in the offline compiler support version of POCL, I've just put macros around it to avoid modifying the code to much.

lib/CL/clEnqueueNDRangeKernel.c
rjodin: If the program has been created with a binary, it won't go in this 'if'. If it does, it means that the program has not been created with binary, and as there is no compiler support, it's an error.

lib/CL/devices/basic/basic.c
rjodin: Done

lib/CL/clCreateKernel.c
franz: @pjaaskel for fetching metadata from the binary dump there's ```pocl_binary_get_kernel_metadata(...) ```

pocl_llvm_get_kernel_metadata() is only for program.bc LLVM bitcode.

pjaaskel:
```
make
Makefile:1050: *** missing separator.  Stop.
```

This is from
```

if OCS_AVAILABLE
  if !POCL_ANDROID
    SUBDIRS += bin examples tests
  endif
endif
```

It generates to
```
...
    tags tags-am uninstall uninstall-am uninstall-icdDATA \
    uninstall-pkgconfigDATA

  
  if !POCL_ANDROID
  endif

pocl.icd:
...
```

I don't believe you can have multiple levels of ifs inside Makefile.ams. Did you test this locally?

franz:
I reviewed most code, except the autotools parts.

Here are the commits:
https://github.com/franz/pocl/tree/pocl_without_llvm

Mostly fixes are just moving ```#ifdef  OCS_AVAILABLE``` to outside of functions which we don't need anyway for OCS-less builds, and some better error handling.

Still TODO:
* autotools
* ```cl_device_id->llvm_cpu``` and triplet - I see that you changed pocl_binary to use ```device->long_name``` instead of llvm_cpu/triplet for the device_hash field. I haven't checked whether all long names (HSA, TCE) are sane so this needs some consideration; @visit0r what do you think is ```device->long_name``` suitable for device_hash ?

pjaaskel:
> at do you think is device->long_name suitable for device_hash ?

It's not enough in TCE. TCE needs to add the contents of the ADF to identify an architecture. That's what the pocl_tce_build_hash() does.

pjaaskel:
...but it seems it's currently broken in master too. It probably now breaks if one tries to load a poclbin produced for an ADF A to ADF B?

franz:
I've attempted to fix it in my branch.

For host CPU, LLVM-less builds require manually specifying one additional build-time variable. 

For TCE, device hash equals pocl_tce_build_hash() (== ADF file content hashed + extra options env var).

HSA has hardcoded device hashes for the 2 devices we recognize.

PR #346, opened 2016-05-23
Implement prefetch builtin as no-op

jrprice:
If a specific device wants to implement real software prefetch they can override this implementation.


lib/kernel/prefetch.cl
```diff
@@ -0,0 +1,56 @@
+/* OpenCL built-in library: prefetch()
+
+   Copyright (c) 2012 Pekka Jääskeläinen / Tampere University of Technology
```
pjaaskel: Please update the year and author.

jrprice:
Updated copyright message as requested.

PR #347, opened 2016-05-23
Pass offset to driver unmapping routines

jrprice:
This is required for any devices that don't have unified memory with the host CPU and therefore need to manually update device memory during an unmap operation.


pjaaskel:
Please update the other drivers also to include this parameter in their implementations.

jrprice:
As far as I can tell the other drivers just use the `basic` implementation for unmapping, is this not the case?

pjaaskel:
True. Except for TCE, which simply doesn't implement it ATM, I think.

PR #362, opened 2016-07-07
release_0_13 branch updates

rcrowder:
Hi there. We have been using PoCL (branch 0.13, LLVM 3.7) on the Raspberry Pi3 (Raspbian Jessie) to run our OpenCL kernel code. It has worked perfectly for that. Thank you.

Our kernel code mainly uses one (CL_R) and two (CL_RG) channel 2D and 3D float images, so the following changes were required to the release 0.13 code. Changes include; 

- Addition of two read_imagef and one write_imagef functions
- clCreateImage update to take into account image depth when calculating memory size
- channels_out update for two byte CL_RG, CL_RGx, and CL_RA image formats

The build instructions we used can be found in this Gist - https://gist.github.com/rcrowder/f1acf9da5f632b289dc5393f5ad578a1

`make check` passes almost all tests. The final EinsteinToolkit test is the only one that fails, due to a missing install of that toolkit.

We're not sure whether you're accepting changes to a previously released branch, or whether it's possible to have a new branch created for these 0.13 updates?


pjaaskel:
Thanks for the PR. So far we haven't maintained the release branches because we aim to keep the master stable, but I don't see why we couldn't pull fixes into it. Is there a reason you are not simply using 'master'? Due to a need to stick to LLVM 3.7?

rcrowder:
When we started a few weeks ago with PoCL on the RPi3, I did initially try from a clone of master. At the time LLVM APT repositories had turned off 3.9 installs, so required a build of LLVM+CLang 3.9 from source. I don't recall the exact problems, but it didn't work with the PoCL master at that time on the RPi3 (maybe OpenCL 2.0 atomics?). Whereas LLVM 3.7 plus PoCL 0.13 worked first time.
We currently only require OpenCL 1.2 functionality, so are not in a rush to move towards PoCL master (0.14+) and can stick with PoCL 0.13 for the moment.

pjaaskel:
Master should support LLVM 3.8. The latest trunk might not work before getting closing to a new pocl release.  BTW is the pocl use case of yours secret? Might be interesting one to write about in portablecl.org?

PR #366, opened 2016-08-16
Fixed lib/CL/pocl_timing.c

noma:
Falls back to CLOCK_MONOTONIC for Linux distributions that come without CLOCK_MONOTONIC_RAW, e.g. SLES (used in Cray's software environment).

- Including "linux/time.h" instead of "time.h" would also solve the issue, but conflicts with other "time.h" includes across the project, which would lead to a much larger patch, so keeping it simple.
- The generic fall back to CLOCK_REALTIME would also work, but CLOCK_MONOTONIC is closer to CLOCK_MONOTONIC_RAW, see http://man7.org/linux/man-pages/man2/clock_gettime.2.html
- Also changed the "#warn" to "#warning", as gcc 5.3.0 wouldn't accept "#warn" and yield an error

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -46,4 +46,5 @@ Pavan Yalamanchili <pavan@arrayfire.com>
 Romaric Jodin <rjodin@kalray.eu>
 Masataro Asai <guicho2.71828@gmail.com>
 Richard Crowder <riccro@gmail.com>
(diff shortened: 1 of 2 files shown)


pjaaskel:
Can you add your real name to the end of CREDITS if not already there? Thanks.

noma:
Done.

PR #375, opened 2016-08-29
Crosscompilation support

franz:
Note this is only for LLVM-less builds.

Toolchain.example has the required steps to use it.

### DIFF
--- a/ToolchainExample.cmake
+++ b/ToolchainExample.cmake
@@ -1,13 +1,17 @@
-# This is an example Toolchain file to cross-compile for ARM
-# or other boards from x86_64
+# This is an example Toolchain file to cross-compile for ARM/MIPS/other
+# boards from x86_64. Copy & modify
+#
 # Steps:
-# 1) Install g++ and gcc cross-compilers (apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf)
+# 1) Install g++ and gcc cross-compilers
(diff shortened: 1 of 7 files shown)


ToolchainExample.cmake
```diff
@@ -0,0 +1,31 @@
+# This is an example Toolchain file to cross-compile for ARM
+# or other boards from x86_64
+# Steps:
+# 1) Install g++ and gcc cross-compilers (apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf)
```
pjaaskel: Your display is too wide :P

cmake/FindHwloc.cmake
```diff
@@ -134,7 +134,30 @@ if(WIN32)
...
+
+  find_library(Hwloc_LIBRARIES
+    NAMES
+      hwloc
+    PATHS
+      ENV HWLOC_ROOT
+  )
+
+  if(Hwloc_INCLUDE_DIRS AND Hwloc_LIBRARIES)
+    # this is kindof arbitrary
```
pjaaskel: This might be a problem with hwloc >= 2 with a new API. Just a heads up.

franz: Yess. I think it's possible to get pkg-config to get the information from target (not host) files, but have to look into it more.

PR #404, opened 2016-12-08
Add a trivial (non functional) implementation for clUnloadPlatformCom…

tom-gall:
Add a trivial (non functional) implementation for clUnloadPlatformCompiler
which pyOpenCL wants, and is also needed for general standards
compliance.

The spec allows this quoting "or that the compiler will actually be unloaded by the implementation"


lib/CL/CMakeLists.txt
```diff
@@ -119,6 +119,7 @@ set(POCL_LIB_SOURCES  "clCreateContextFromType.c"
                    "clReleaseDevice.c"
                    "clRetainDevice.c"
                    "clCreateSubDevices.c"
+				   "clUnloadPlatformCompiler.c"
```
pjaaskel: Is there a problem with indentation here?

lib/CL/clUnloadPlatformCompiler.c
```diff
@@ -0,0 +1,32 @@
+/* OpenCL runtime library: clUnloadPlatformCompiler()
+ *
+ *    Copyright (c) 2016 Tom Gall Tampere Univ. of Tech.
```
pjaaskel: Perhaps the correct credits here would be "Tom Gall for Tampere Univ. of Tech." ?

PR #408, opened 2016-12-29
two build fixes

kraiskil:
Make pocl build on my computer again :)


pjaaskel:
There was some discussion in IRC regarding this. Was the conclusion that this PR is a valid fix?

kraiskil:
- 6c1209535 did not cause discussion that I know of
- 9d12b7404 is a work-around the case where we compile pocl with some other compiler than was used to compile LLVM (e.g. the case of binaries from LLVM.org). @franz felt this was inviting trouble down the line, I disagreed :)
- 46a71e255 is an ugly hack. We need better target detection in CMake on ARM to know which FPU is used. This is a feeble start, i.e. enable vfpu4 by default. With no FPU specified, pocl would not compile with LLVM 3.9 (cortex-a7, ubuntu14)

Issue #413, opened 2017-01-09
Pocl 0.13 (with Clang 3.8.1) unusable on Skylake


inducer:
Due to what appears to be [this LLVM bug](https://llvm.org/bugs/show_bug.cgi?id=29162), running even the simplest pocl kernels on Skylake processors results in an "Illegal Instruction" signal. The root of the issue *may* be that there will ultimatelyl be two versions of Skylake--one without AVX-512 and one with. From the bug log, it's not quite clear whether or not this is fixed in 3.9.

This [workaround](https://github.com/inducer/staged-recipes/blob/pocl/recipes/pocl/work-around-skylake-codegen.patch) suffices to avoid generating illegal instructions. (If you decide to merge it, you should also bump the cache version--if you have such a thing--so that old cached binaries don't get reused.)

cc @mattwala

vchuravy:
We also did run into this for Julia and it is now fixed upstream (see https://reviews.llvm.org/D28221)

pjaaskel:
Please add LLVM 3.8 version checks for your workaround and send a PR. I believe it's fixed in 3.9.

inducer:
Done.

PR #414, opened 2017-01-11
Work around Skylake codegen issue in LLVM 3.8 (0.13 branch)

inducer:
See #413.


pjaaskel:
If the issue is only with LLVM 3.8, please use the LLVM_3_8 macro instead, and also please add a brief comment why the workaround is there.

inducer:
Done (on both PRs).

pjaaskel:
I pushed to master, but rather not start maintaining the previous release branches. I do understand that you prefer 0.13 until the regressions affecting pytoolkit are solved, but let's hope we can tackle them soon enough.

PR #426, opened 2017-02-02
use memcpy to avoid unaligned access in pocl_binary

rjodin:
When a processor does not support unaligned access, reading a pocl binary file creates an unauthorized memory access.
"memcpy" function is normaly implemented so that it does no unaligned access if the architecture does not support it.

### DIFF
--- a/lib/CL/pocl_binary.c
+++ b/lib/CL/pocl_binary.c
@@ -128,11 +128,14 @@ typedef struct pocl_binary_s
 /***********************************************************/
 
 #define BUFFER_STORE(elem, type)                  \
-  *(type*)buffer = (type)TO_LE((type)elem);       \
-  buffer += sizeof(type)
+  do {                                            \
+    type b_s_tmp = (type)TO_LE((type)elem);       \
+    memcpy(buffer, &b_s_tmp, sizeof(type));       \
+    buffer += sizeof(type);                       \
+  } while(0)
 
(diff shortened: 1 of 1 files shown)


lib/CL/pocl_binary.c
```diff
@@ -128,11 +128,14 @@ typedef struct pocl_binary_s
 /***********************************************************/
 
 #define BUFFER_STORE(elem, type)                  \
-  *(type*)buffer = (type)TO_LE((type)elem);       \
-  buffer += sizeof(type)
+  do {                                            \
+    type b_s_tmp = (type)TO_LE((type)elem);       \
```
pjaaskel: Otherwise looks good but please mind the style (GNU C style in C files). https://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting  at least the function argument lists need additional white spaces. Like: func_call (arguments, here);  Yes, the original was also wrongly formatted, but the intention is to get every future commit to improve on this matter also.

rjodin: Should I put space between the cast type and the casted variable?

pjaaskel: Yes. https://www.gnu.org/prep/standards/html_node/Syntactic-Conventions.html#Syntactic-Conventions

rjodin: Changes done.

PR #429, opened 2017-02-06
Applying read_image optimizations to write_image as well

ardacoskunses:


ardacoskunses:
Requested changes applied and pushed.

PR #441, opened 2017-02-13
Detect presence of SPIR CC patch, make visible in version number

inducer:
Here's a snippet of CMake that makes the patch from #1 detectable both for other CMake code and for user code through the device version number.

Full disclosure: I learned what little CMake I needed for this roughly a half hour ago. If there is a better way to do this, let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,6 +319,31 @@ check_function_exists(fork HAVE_FORK)
 check_function_exists(vfork HAVE_VFORK)
 
 ######################################################################################
+# Test for presence of Clang calling convention patch from
+# https://github.com/pocl/pocl/issues/1
+
+if(CLANGXX_WORKS AND EXISTS "${CMAKE_SOURCE_DIR}/cmake/spir-cc-test-kernel.cl")
+  execute_process(
+    COMMAND
+    "${CLANGXX}" -S -x cl -emit-llvm "${CMAKE_SOURCE_DIR}/cmake/spir-cc-test-kernel.cl" -o -
+    OUTPUT_VARIABLE SPIR_PATCH_TEST_IR
+    ERROR_VARIABLE _DUMMY
+    RESULT_VARIABLE RESV1)
+  if(RESV1)
+    message(FATAL_ERROR "Clang exited with non-zero status when trying to compile calling convention test")
+  endif()
+  string(FIND "${SPIR_PATCH_TEST_IR}" "spir_kernel" RESV1)
+  if("${RESV1}" MATCHES "-1")
+    set(CLANG_IS_PATCHED_FOR_SPIR_CC 0 INTERNAL "Whether clang has the patch from https://github.com/pocl/pocl/issues/1")
+    message(STATUS "Clang is NOT patched for SPIR CC")
+  else()
+    set(CLANG_IS_PATCHED_FOR_SPIR_CC 1 INTERNAL "Whether clang has the patch from https://github.com/pocl/pocl/issues/1")
+    set(POCL_VERSION "${POCL_VERSION}-spirccpatch")
+    message(STATUS "Clang IS patched for SPIR CC")
+  endif()
+endif()
+
+######################################################################################
 
 if(NOT DEFINED DEFAULT_USE_VECMATHLIB)
   if(CLANGXX_WORKS AND EXISTS "${CMAKE_SOURCE_DIR}/lib/kernel/vecmathlib/vecmathlib.h")
--- a/cmake/spir-cc-test-kernel.cl
+++ b/cmake/spir-cc-test-kernel.cl
@@ -1,0 +1,12 @@
+#pragma OPENCL EXTENSION cl_khr_fp64 : enable
+
+struct fakecomplex {
+    double x;
+      double y;
+};
+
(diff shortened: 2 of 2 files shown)


pjaaskel:
On it. I'll fix the testsuite also.

inducer:
Thx!

inducer:
@pjaaskel Looks like you removed the version marker that allows me to tell whether the patch is applied from client code. Are you opposed to that, or was that an accident?

My point is that I want to be able to tell from PyOpenCL what type of POCL I am talking to without having to build a kernel and count the arguments...

pjaaskel:
Oh, I see. But that patch is for LLVM, it doesn't affect anything inside pocl per se, thus the version string should be patched to llvm-config --version? We have done similar in TCE. https://github.com/cpc/tce/blob/master/tce/tools/patches/llvm-4.0-custom-vector-extension.patch#L564

inducer:
But, but, it does. :) To be able to successfully pass complex numbers to kernels (which previously would explode into two arguments), PyOpenCL carefully psychonanalyzes Pocl 0.13 and the x86-64 ABI and feeds it just the arguments required to make it go, OpenCL spec be damned. (Crazy: feeding it two reals instead of one complex does seem to work.)

Now, heaven forbid, if pocl suddenly starts following the CL spec (such as with this patch), PyOpenCL will currently feed it the *wrong* data, and all sorts of entertainment will ensue.

What I'd like is for PyOpenCL to be able to tell whether it needs to feed pocl exploded-complex crazytalk, or civilized, spec-conforming complex structures. To do so, I need to be able to tell from the other side of the ICD whether pocl has this patch applied.

pjaaskel:
Yes, I understand your need, but my point holds. This patch changes Clang, not pocl. Meaning that you could dynamically link to a non-patched LLVM/Clang and it might just report it wrongly. And also you can use the same patched Clang elsewhere with the modified behavior. It just happens that the issue pocl suffers from is in Clang and this makes it go away. Thus, your check for if the workaround is needed should ask llvm-config --pyopencl-patches or similar as pocl code base and behavior (!) is exactly the same with and without this patch.

pjaaskel:
Or is there a problem asking it from the Clang pocl uses here?

pjaaskel:
Hmm. There might be a problem in case Clang is linked statically to libpocl.so and llvm-config cannot be trusted to point to the same one? I suppose you are right that pocl needs to report this. I wonder if there's a more suitable platform string though?

inducer:
Some comments:

* `llvm-config` is a non-starter, because the LLVM development bits aren't guaranteed to be around. (Debian for example packages them separately in a `-dev` package)
* I'd argue that it almost *has* to show up in the driver version (i.e. that's the right platform string to use), because what PyOpenCL does (and what I imagine quite a few other CL-using pieces of software do) is cache binaries based on that version. And I'd argue we would want the "patch there: yes/no" bit to be part of the cache key--pocl *does* build a different a substantially different binary with the patch after all.

pjaaskel:
Yes, I added it as a separate salt in pocl's kernel cache key. I'd argue the whole PyOpenCL's kernel caching is redundant AFAIU. But, sure, let's add it back to the version string until accepted to the LLVM upstream. Not a big deal to keep arguing about for days ;)

inducer:
> I'd argue the whole PyOpenCL's kernel caching is redundant AFAIU.

For POCL it is. I'll make a widget to turn it off for ICDs that are known to be caching.

> let's add it back to the version string until accepted to the LLVM

OK, how about adding it with inverse semantics (i.e. `-nospirccpatch`).

Also, you patch or me patch?

pjaaskel:
Yep, in your case you can just turn off your cache if you see 'pocl'. But of course you have a point there. 

If you want to make it happen sooner, please send a PR. I'm travelling for some time so don't know when I can focus on pocl again, but I can test and push it for sure.

pjaaskel:
> OK, how about adding it with inverse semantics (i.e. -nospirccpatch)?

The inverse won't work as there might be already pocls built against non-patched LLVMs which should use your trickery for the ABI issues. So I guess we need to have a tag there until it's upstreamed after which the LLVM version itself should imply the patch inclusion.

inducer:
OIC. I hadn't noticed the LLVM version in the platform version previously.

From PyOpenCL's perspective, I think I'll actually be able to do without this--I'll do the test kernel compilation after all. That'll give me a reasonably quick and definite answer without needing complex version heuristics.

And I'll turn off the cache for Pocl, so that concern disappears for me as well.

So what remains is exhibiting the right platform/driver version so that any other pieces of software that might be caching are OK.

inducer:
> And I'll turn off the cache for Pocl, so that concern disappears for me as well.

inducer:
> And I'll turn off the cache for Pocl, so that concern disappears for me as well.

FYI: On pocl 0.13, PyOpenCL's test suite takes about 83 s with pyopencl's binary cache and about 140 s with pocl's built-in cache. Both cases are run with a warm cache. (0.14: 140 s for both) As a result, I'm keeping my cache for now.

pjaaskel:
OK. Would be interesting to profile the binary program implementation (and the cache) at some point. Weird that it takes so long.

PR #457, opened 2017-04-02
NDRange fixes

### LINKED ISSUE #450: Missing error detection for when the NDRange-requested local size doesn't match reqd work group size attribute's
Run the following in an environment that has PyOpenCL:

```
$ git clone https://github.com/inducer/boxtree.git
$ cd boxtree
$ pip install -e .
$ cd test
$ pip install mako pytest
$ PYOPENCL_TEST=port python -m pytest --capture=no
```

After a non-deterministic amount of time, observe that the program just crashes/stops and prints the following:
```
INFO:boxtree.tree_build:tree build complete
INFO:root:building traversal build kernels
INFO:root:traversal build kernels built
INFO:boxtree.traversal:start building traversal (... or similar program output)
Dynamic WG size binary does not exist
```

This *doesn't seem to happen with:
```
$ PYOPENCL_NO_CACHE=1 PYOPENCL_TEST=port python -m pytest --capture=no
```

Oblomov:
This is a set of cleanups and fixes for `clEnqueueNDRangeKernel`. The main highlights are checks that for kernels with a required work-group size the user submitted that work-group size (which should fix #450), and a rewrite of the logic used to determine the work-group size in the NULL case, designed around the hard sizing constrains of dividing the global size, being less than the maximum allowed local size in each dimension while still trying to take the preferred work-group size multiple of the device into account.

### DIFF
--- a/lib/CL/clEnqueueNDRangeKernel.c
+++ b/lib/CL/clEnqueueNDRangeKernel.c
@@ -117,7 +117,23 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
       local_z = work_dim > 2 ? local_work_size[2] : 1;
     }
 
-  if (local_work_size == NULL ||
+  /* If the kernel has the reqd_work_group_size attribute, then the local
+   * work size _must_ be specified, and it _must_ match the attribute specification
+   */
+  if (kernel->reqd_wg_size != NULL &&
+      kernel->reqd_wg_size[0] > 0 &&
+      kernel->reqd_wg_size[1] > 0 &&
+      kernel->reqd_wg_size[2] > 0)
+    {
+      POCL_RETURN_ERROR_COND((local_work_size == NULL ||
+          local_x != kernel->reqd_wg_size[0] ||
+          local_y != kernel->reqd_wg_size[1] ||
+          local_z != kernel->reqd_wg_size[2]), CL_INVALID_WORK_GROUP_SIZE);
+    }
+  /* otherwise, if the local work size was not specified or it's bigger
+   * than the global work size, find the optimal one
+   */
+  else if (local_work_size == NULL ||
       (local_x > global_x || local_y > global_y || local_z > global_z))
     {
       /* Embarrassingly parallel kernel with a free work-group
--- a/lib/CL/clEnqueueNDRangeKernel.c
+++ b/lib/CL/clEnqueueNDRangeKernel.c
@@ -162,11 +162,14 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
           local_y != kernel->reqd_wg_size[1] ||
           local_z != kernel->reqd_wg_size[2]), CL_INVALID_WORK_GROUP_SIZE);
     }
-  /* otherwise, if the local work size was not specified or it's bigger
-   * than the global work size, find the optimal one
+  /* otherwise, if the local work size was not specified find the optimal one.
+   * Note that at some point we also checked for local > global. This doesn't
+   * make sense while we only have 1.2 support for kernel enqueue (and
+   * when only uniform group sizes are allowed), but it might turn useful
+   * when picking the hardware sub-group size in more sophisticated
+   * 2.0 support scenarios.
    */
-  else if (local_work_size == NULL ||
-      (local_x > global_x || local_y > global_y || local_z > global_z))
+  else if (local_work_size == NULL)
     {
       /* Embarrassingly parallel kernel with a free work-group
          size. Try to figure out one which utilizes all the
--- a/lib/CL/clEnqueueNDRangeKernel.c
+++ b/lib/CL/clEnqueueNDRangeKernel.c
@@ -44,6 +44,38 @@
 
 //#define DEBUG_NDRANGE
 
+/* Euclid's algorithm for the Greatest Common Divisor */
+static size_t gcd(size_t a, size_t b)
+{
+  int c;
+  while (a) {
+    c = a; a = b % a; b = c;
+  }
+  return b;
+}
+
+/* Find the largest divisor of dividend which is less than limit */
+static inline size_t upper_divisor(size_t dividend, size_t limit)
+{
+  /* The algorithm is currently not very smart, we
+   * start from limit and subtract until we find something
+   * that divides dividend. In optimal conditions this is found
+   * quickly, but it takes limit steps if dividend is prime.
+   * TODO FIXME improve algorithm
+   */
+  if (dividend < limit) return dividend; // small optimization
+  assert(limit > 0); // should never be called with limit == 0
+  while (dividend % limit != 0) --limit;
+  return limit;
+}
+
+/* Check that a divides b and b divides c */
+static inline int divide_chain(size_t a, size_t b, size_t c)
+{
+  return (b % a == 0 && c % b == 0);
+}
+
+
 CL_API_ENTRY cl_int CL_API_CALL
 POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
                        cl_kernel kernel,
@@ -58,6 +90,11 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
   size_t offset_x, offset_y, offset_z;
   size_t global_x, global_y, global_z;
   size_t local_x, local_y, local_z;
+  /* cached values for max_work_item_sizes, since we are going to access them repeatedly */
+  size_t max_local_x, max_local_y, max_local_z;
+  /* cached values for max_work_group_size, since we are going to access them repeatedly */
+  size_t max_group_size;
+
   int b_migrate_count, buffer_count;
   unsigned i;
   int error = 0;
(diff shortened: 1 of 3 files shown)


franz:
Looks OK-ish to me, but i only stared at it for a while :)

It could be beneficial to move some of the guessing logic to a device driver hook, so each device can decide itself (e.g. HSA device + kernel with image arguments might be better of with "square" local sizes), but that's low on the TODO list.

I'm going to run this through some of pocl testsuites, if nothing breaks it's an ACK from me.

jrprice:
One concern I have with this patch is that (if I'm reading it correctly) it will greedily pick the largest work-group size it can when `local_size==NULL`. For example, if I enqueue an NDRange with global size `(4096,1,1)`, this will pick a work-group size of `(4096,1,1)`. This means you only have a single work-group, which means you will not be able to take advantage of any parallelism over multiple CPU cores.

Maybe this should also take the number of compute units into account?

Oblomov:
@jrprice I agree that the greedy algorithm might not always be the optimal one, and as @franz suggested there's plenty of room even for device-specific optimizations. Even on CPU, for example, there's no guarantee that splitting a 4096 global work size across multiple compute units would be better than on a single one, depending on amount of work, nature of memory accesses, potential effects of NUMA, etc.

But that's not the aim of this patchset. Here I focused on fixing what was definitely broken in the previous logic, rather than aiming at specific improvements. I actually strove to keep as close as possible to the original logic, both outside and inside the reduction loop. Further improvements are obviously possible, but a more careful analysis of the impact of the local size on execution is needed for those.

pjaaskel:
Nice to see someone working on this! @Oblomov can you first point out what was broken in the original logic? 

You are right that small grid sizes are problematic. We'd need to check how large kernel it is in terms of LLVM IR instructions (and if it has outer loops) for example to decide the number of cores to utilize with multiple WGs, but I think _always_ resorting to a single core has brings too major regression on performance with multicores.

Another aspect is that the original code (IIRC) tries to make the preferred width to multiple dimensions and now that I think about it, there's at least this logic behind: The vectorizer might be able to vectorize across some other dimension than X when/if it can utilize loop interchange if X is not suitable (e.g. due to memory access patterns not being sequential). I haven't checked how well loop interchange works in LLVM at the moment, but I've relied on it eventually taking care this case and also "outer loop vectorization".

Oblomov:
@pjaaskel The main problem of the original logic is that it did not take into account the two hard constraints (local work size must divide global work size, and local work size in each direction must not be larger than the maximum allowed size in that direction): instead, it started from a local size as large as the global work size and then halved (or clipped) directions until the overall local size fit within the hard third constraint (being no larger than the maximum overall local size).

This meant that the logic might produce invalid local sizes, e.g. by having a local size larger than the maximum allowed in that direction or a local size that did not divide the global size. This would be caught in the subsequent checks, but guessing an invalid local size is IMO wrong in principles.

So the main difference in my logic is that local size guessing now starts from the hard constraints (divisibility, and maximum size per direction). The logic used for the reduction until we fit in the third constraint is actually _mostly_ the same.

BTW, I feel this whole explanation should go in the commit message itself. Let me do just that.

pjaaskel:
OK, I see. Still I think there should be multiple WGs generated by default. I have a feeling the performance regression resulting from your patch will be more noticeable than the original issue has been.

Oblomov:
I don't think the new logic would introduce performance regressions, since for cases where the old logic produced valid work-group sizes, the new logic would produce workgroups with the same overall size: the main difference in the reduction is that we now prefer to halve Z first, then Y, then X instead of X first, then Y, then Z.

So, for example, assuming local size limits of (4096, 4096, 4096) with total limit of 4096, a global size of (4096, 2) would produce a local size of (2048, 2) with the old logic, and local size of (4096, 1) with the new logic. In terms of CU usage, there should be no difference.

Oblomov:
I've force-pushed an update for the message of the last commit, with a more detailed explanation of the differences between the old and new logic and highlighting some of the aspects discussed in the comments (particularly regarding the aim of the patch and the possibility of performance regression).

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -44,6 +44,38 @@
...
+{
+  int c;
+  while (a) {
+    c = a; a = b % a; b = c;
+  }
+  return b;
+}
+
+/* Find the largest divisor of K which is less than L */
+static inline size_t upper_divisor(size_t K, size_t L)
```
pjaaskel: Why capital K and L?

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -110,15 +147,63 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
   if (local_work_size != NULL)
     {
       local_x = local_work_size[0];
       local_y = work_dim > 1 ? local_work_size[1] : 1;
       local_z = work_dim > 2 ? local_work_size[2] : 1;
-      if (local_x > global_x || local_y > global_y || local_z > global_z)
-        goto DETERMINE_LOCAL_SIZE;
+
+      POCL_RETURN_ERROR_ON((local_x * local_y * local_z > max_group_size),
+        CL_INVALID_WORK_GROUP_SIZE, "Local worksize dimensions exceed device's max workgroup size\n");
```
pjaaskel: Check long lines.

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -110,15 +147,63 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+       */
+      POCL_RETURN_ERROR_COND((global_x % local_x != 0), CL_INVALID_WORK_GROUP_SIZE);
+      POCL_RETURN_ERROR_COND((global_y % local_y != 0), CL_INVALID_WORK_GROUP_SIZE);
+      POCL_RETURN_ERROR_COND((global_z % local_z != 0), CL_INVALID_WORK_GROUP_SIZE);
+
     }
-  else
+
+  /* If the kernel has the reqd_work_group_size attribute, then the local
+   * work size _must_ be speficied, and it _must_ match the attribute specification
```
pjaaskel: Typo speficied

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -110,15 +147,63 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+    }
+  /* otherwise, if the local work size was not specified find the optimal one.
+   * Note that at some point we also checked for local > global. This doesn't
+   * make sense while we only have 1.2 support for kernel enqueue (and
+   * when only uniform group sizes are allowed, but it might turn useful
+   * when picking the hardware sub-group size in more sophisticated
+   * 2.0 support scenarios, so keep the conditional around, commented
+   */
+  else if (local_work_size == NULL)
+    /* || (local_x > global_x || local_y > global_y || local_z > global_z */
```
pjaaskel: Please remove commented out code.

Oblomov:
Updated. Hopefully I addressed all comments. I've also included a preliminary “try to fill all CUs” algorithm after the greedy one. (It's intentionally separate, since the logic for the two is quite distinct.)

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -129,81 +217,212 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+if (local_##c1 > 1 && local_##c1 <= local_##c2 && local_##c1 <= local_##c3 && \
+    local_##c2*local_##c3 >= min_group_size) \
+  { \
+    local_##c1 = 1; nwg_##c1 = global_##c1; \
+    continue; \
+  }
+
+          while (splits > 1)
+            {
+              TRY_SPLIT(z) else TRY_SPLIT(y) else TRY_SPLIT(z)
```
pjaaskel: Should it be TRY_SPLIT (x) in the last else?

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -129,81 +217,212 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+      size_t nwg_y = global_y/local_y;
+      size_t nwg_z = global_z/local_z;
+
+      size_t splits; /* number of splits to bring ngws to reach ncu */
+      /* Only proceed if splitting wouldn't bring us below the minimum
+       * group size */
+      while (((splits = ncus/(nwg_x*nwg_y*nwg_z)) > 1) &&
+             (local_x*local_y*local_z > splits*min_group_size))
+        {
+          /* Very simple splitting approach: find a dimension divisible
```
pjaaskel: I think this is OK, but a tad better would be to favour splits that maintain as many dimensions as possible the preferred SIMD width -- but this can be optimized forever.

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -129,81 +217,212 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+       * resources.
+       * Lacking that, as a first step we will simply try to guarantee that we
+       * have at least one work-group per CU, as long as the local work size
+       * does not drop below a given threshold.
+       */
+
+      /* Pick a minimum work-group size of 4 times the preferred work-group size
+       * multiple, under the assumption that this would be a good candidate
+       * below which a Compute Unit will not do enough work. */
+      const size_t min_group_size = 4*preferred_wg_multiple;
```
pjaaskel: Please check the GNU C style. There are some missing white spaces before and after operators and before (.

https://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting

lib/CL/clEnqueueNDRangeKernel.c
```diff
@@ -129,81 +217,212 @@ POname(clEnqueueNDRangeKernel)(cl_command_queue command_queue,
...
+
+      /* We need the number of Compute Units in the device, since we want
+       * at least that many work-groups, if possible */
+
+      cl_uint ncus; /* number of compute units */
+      prop_err = POname(clGetDeviceInfo)(command_queue->device,
+        CL_DEVICE_MAX_COMPUTE_UNITS, sizeof(cl_uint), &ncus, NULL);
```
jrprice: Is there any reason not to just access `device->max_compute_units` directly?

pjaaskel:
Oh, and please describe the update in CHANGES.

Issue #458, opened 2017-04-04
Issue with poclcc while compiling kernel with barriers


rjodin:
Hi all,

I have an issue when I run a kernel create from binary that contains barrier.
At the moment I didn't manage to get a simple code that reproduce this bug.

This issue is that the kernel does not produce the outputs expected. The program ends well, but the computation is false.

You can find here attached a tar that contains the kernel and everything to run it.
In my kernel, barriers come from calls to wait_group_events.
As the code is not that simple, it is hard to know where the compiler made an error. But from what I tried to have a simple code, it seems that I need the barrier to be in an if which is in a loop... but I'm not sure.

Does anyone have seen something similar? Any idea on where to look, what to check?
The only difference between compiling offline (with poclcc) and online (as usual) that I'm aware of is that the iterator of the 3 loops (on the 3 dimensions) are variables that are loaded dynamically during kernel execution from pocl_context. Is there anything else?

To run my code, you have to write the path to your pocl and llvm install repositories in the Makefile.
Then you have 4 different targets:
```
make run_bin
make run_bin_no_wait_event
make run_src
make run_src_no_wait_event
```

The only target that fail is "make run_bin" which runs the kernel from binary with barriers (from wait_group_events).
"make run_bin_no_wait_event" runs the kernel from binary but without barriers.
"make run_src" and "make run_src_no_wait_event" runs the kernel from source.

Thanks in advance,
[issue_poclcc_with_barriers.tar.gz](https://github.com/pocl/pocl/files/893272/issue_poclcc_with_barriers.tar.gz)

pjaaskel:
These are the trickiest barrier cases to handle statically so I'm not surprised if you encouter a problem. You can try if switching the work group method (repl vs. loops) makes it fail also online. If it does, you need to dig into the kernel compiler and figure out the failure it creates to the CFG (it that's the case). A way to minimize/expose CFG errors is to use printfs() that output work item ids. This has been used in tests under tests/workgroup/*.cl. The trickiest case is tricky_for.cl which I believe might still fail with one of the WG methods. DebugHelpers.h is useful for visualizing the CFGs before and after the CFG transformations.

rjodin:
Thanks @pjaaskel,
When creating a kernel with poclcc, you cannot choose the work group method. It has to be loops as we don't know the local sizes.
When I run the kernel from source, both work group methods work find.

I did try to puts some printf.
My code is basically (with the printf):
```
for (i) {
   compute(i);
[...]

PR #461, opened 2017-04-10
fix issue #458

rjodin:


rjodin:
I'm working on adding a regression test.

What should I do about the coding style? I try to respect the actual coding style of the file. If I understand correctly, you're saying that the actual coding style of the file is not the expected one. Should I modify my patch so that it matches the C++ style, even if it will mix coding style in this file?

pjaaskel:
Yes please modify your patch for the correct formatting. I know the existing code has style issues still, but my idea is that every commit brings the style towards the correct one, not further.

rjodin:
Changes made, tell me if you see something else.

PR #465, opened 2017-04-21
Add NVIDIA GPU support

jrprice:
This is a bit of a monolithic patch, but for the most part the changes are self-contained.

Essentially, this is a PTX/CUDA backend for pocl. We use the NVPTX backend in LLVM to generate PTX, which requires a handful of LLVM IR transformations (all in `lib/CL/devices/cuda/pocl-ptx-gen.cc`). We then use the CUDA driver API to implement the runtime functionality.

This is far from finished, but at this stage it's passing 74 tests from the internal test suite, and can run some real programs (including things that NVIDIA's own OpenCL implementation cannot, such as SPIR-based codes).

I've added a doc page describing how to build it and run it, along with the current status.

Any and all feedback/questions welcome!


lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
...
+    call->setCalledFunction(new_vprintf);
+  }
+
+  vprintf_func->eraseFromParent();
+}
+
+void pocl_cuda_link_libdevice(llvm::Module *module, const char *kernel,
+                              const char *gpu_arch) {
+  // TODO: Can we link libdevice into the kernel library at pocl build time?
+  // This would remove this runtime dependency on the CUDA toolkit.
```
pjaaskel: Makes sense to me -- if the license allows it.

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
...
+  // Run optimization passes to clean up unused functions etc
+  llvm::PassManagerBuilder Builder;
+  Builder.OptLevel = 3;
+  Builder.SizeLevel = 0;
+  Builder.populateModulePassManager(passes);
+
+  passes.run(*module);
+}
+
+void pocl_gen_local_mem_args(llvm::Module *module, const char *kernel) {
```
pjaaskel: Can you add a comment here what this function really does.

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
...
+  for (auto U = callers.begin(); U != callers.end(); U++) {
+    llvm::CallInst *call = llvm::dyn_cast<llvm::CallInst>(*U);
+    if (!call)
+      continue;
+    call->eraseFromParent();
+  }
+  func->eraseFromParent();
+}
+
+void pocl_cuda_fix_printf(llvm::Module *module) {
```
pjaaskel: This is an interesting function. Can you add a brief comment what goes on here? CUDA doesn't have varargs support ?

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
...
+
+  function->eraseFromParent();
+}
+
+void pocl_insert_ptx_intrinsics(llvm::Module *module) {
+  struct ptx_intrinsic_map_entry {
+    const char *varname;
+    const char *intrinsic;
+  };
+  struct ptx_intrinsic_map_entry intrinsic_map[] = {
```
pjaaskel: We just discussed an issue related to this with Michal. We are doing this backwards in pocl (legacy code). We should use builtins to start with which can be then lowered to whatever the target wants (by default to use the hidden context struct).

pjaaskel: Just a note, the code seems fine.

lib/CL/pocl_llvm_api.cc
```diff
@@ -1404,7 +1404,9 @@ static PassManager& kernel_compiler_passes
   }
   // Add the work group launcher functions and privatize the pseudo variable
   // (local id) accesses.
-  passes.push_back("workgroup");
+  // TODO: get the workgroup pass working with CUDA?
```
pjaaskel: I think the workgroup pass is not needed with CUDA. It would be best not to have CUDA branch here in the code, but have something similar to 'SPMD' property above, but I think the best of all would be to move the pass creation to the device completely so all targets could customize their kernel code generation as they see best. If you consider one of these two much effort at this point, I think this is fine for now.

lib/kernel/cuda/get_global_id.c
```diff
@@ -0,0 +1,51 @@
...
+extern uint _local_id_x;
+extern uint _local_id_y;
+extern uint _local_id_z;
+
+// extern uint _global_offset_x;
+// extern uint _global_offset_y;
+// extern uint _global_offset_z;
+
+size_t _CL_OVERLOADABLE
+get_global_id(unsigned int dimindx)
```
pjaaskel: Would it be possible to call the PTX builtins directly here instead of the LLVM conversion phase you do?

pjaaskel: Same comment to other ID related functions below.

lib/kernel/cuda/get_local_id.c
```diff
@@ -0,0 +1,38 @@
+/* OpenCL built-in library: get_local_id() for CUDA
```
pjaaskel: Isn't these identical to the "default implementations" in lib/kernel? In that case you can just use the generic source instead of copying it for CUDA. This has a CMake mechanism. See the other device libs.

lib/llvmopencl/TargetAddressSpaces.cc
```diff
@@ -565,6 +565,11 @@ TargetAddressSpaces::runOnModule(llvm::Module &M) {
     addrSpaceMapDown[POCL_AS_FAKE_GLOBAL] = 1;
     addrSpaceMapDown[POCL_AS_FAKE_LOCAL] = 3;
     addrSpaceMapDown[POCL_AS_FAKE_CONSTANT] = 2;
+  } else if (arch.startswith("nvptx")) {
```
pjaaskel: You might want to try if NVPTX works without the fake address spaces now. By default we don't use them anymore and it's likely that the code path will be phased out. TAS pass is specifically fragile and the main reason I wanted to get rid of the fake ids.

franz:
A little bit of CMake work required (i can help here, it's quite easy).

```
 *** INFO ***  loading libdevice from '/nvvm/libdevice/libdevice.compute_30.10.bc'
[CUDA] failed to open libdevice library file
```

```
  const char *cuda_path = pocl_get_string_option("POCL_CUDA_TOOLKIT_PATH", "");
  const char *libdevice_fmt = "%s/nvvm/libdevice/libdevice.compute_%d.10.bc";
```

Is it possible / usual to have multiple installations of CUDA on a single linux machine ? if not, then i think we can CMake-discover and compile in the cuda path, no need to snprintf() it. The "ldsm" parameter still need to be snprintf()-ed, ofc.

README.CUDA
```diff
@@ -0,0 +1,40 @@
+Support for NVIDIA GPUs via CUDA is currently experimental and many
```
jrprice: Yes, I'll just remove it. Maybe worth adding a note about `-DENABLE_CUDA` to the "options and features" section of `INSTALL` though?

README.CUDA
```diff
@@ -0,0 +1,40 @@
...
+
+Requires `POCL_CUDA_TOOLKIT_PATH` environment variable to be set to the root of
+the CUDA installation (containing the `nvvm` directory).
+
+The `POCL_CUDA_GPU_ARCH` environment variable can be set to override
+the target GPU architecture (e.g. `POCL_CUDA_GPU_ARCH=sm_35`), which
+may be necessary in cases where LLVM doesn't yet support the
+architecture (currently an issue for `sm_32` and `sm_53`).
+
+The `POCL_CUDA_NO_PTX_CACHE` environment variable can be set to `1` to
```
jrprice: It wasn't working before, but I've just fixed the issue, so I'll remove this CUDA specific variable now.

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
+/* pocl-ptx-gen.cc - PTX code generation functions
```
jrprice: Will do - I ran `clang-format` but I forgot to check style of variable/function names etc.

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -0,0 +1,691 @@
...
+  // Load bitcode
+  llvm::LLVMContext context;
+  llvm::Expected<std::unique_ptr<llvm::Module>> module =
+      parseBitcodeFile(buffer->get()->getMemBufferRef(), context);
+  if (!module) {
+    POCL_MSG_ERR("[CUDA] ptx-gen: failed to load bitcode\n");
+    return 1;
+  }
+
+  llvm::InitializeAllTargets();
```
jrprice: I'm sure there was a reason I added these, but it seems to be working without them now, so I'll remove them.

lib/kernel/cuda/get_global_id.c
jrprice: I think the problem is that the PTX builtins have `.` characters in the names, so can't be called from C.

lib/kernel/cuda/get_local_id.c
jrprice: The difference is the type of these external variables. In PTX these special registers are always 32-bits, so we need to use `uint` here instead of `size_t`. I suppose we could use the default `size_t` versions and then hack the LLVM IR later, but I think that would be messier?

lib/llvmopencl/TargetAddressSpaces.cc
jrprice: Yes, it definitely works without fake address spaces now since we require LLVM 4.0. I wasn't sure what to do with this code - should we leave it there, just remove it, or assert that we shouldn't get here at all if we're using the PTX backend?

jrprice:
Thanks for all the feedback. I've pushed a few updates, and I'll try and address the remaining items over the next few days.

> Is it possible / usual to have multiple installations of CUDA on a single linux machine ? if not, then i think we can CMake-discover and compile in the cuda path, no need to snprintf() it.

It's definitely possible to have more than one CUDA toolkit installed on a machine. I have 2-3 on all of my machines, and on many HPC clusters there will also be several versions available. Also, if pocl is distributed as a binary (e.g. a Linux package), the location of the CUDA toolkit at compile-time might be different to where it is installed on the end-user machine. Longer term I want to remove this runtime dependency completely as per the comment in the code.

I think detecting the CUDA location at compile-time and using this as the default path is a good idea though (instead of just using an empty string!), it definitely would make things easier for users in most situations. I'll try and sort something out soon.

lib/kernel/cuda/get_global_id.c
pjaaskel: Could you add get_global_id.ll and call from LLVM instead?

lib/llvmopencl/TargetAddressSpaces.cc
pjaaskel: Assert sounds sensible.

lib/kernel/cuda/get_global_id.c
jrprice: I've done something like this now, with one `.ll` that wraps all the NVVM intrinsics with functions that are callable from C.

jrprice:
OK, I think I've addressed all of the feedback now. I've added some CMake stuff for auto-detecting the CUDA toolkit path as @franz suggested.

Let me know if there is anything else.

lib/CL/devices/cuda/pocl-ptx-gen.h
```diff
@@ -31,10 +31,12 @@ extern "C"
 {
 #endif
 
-int pocl_ptx_gen(const char *bc_filename,
-                 const char *ptx_filename,
-                 const char *kernel_name,
-                 const char *gpu_arch);
+// Generate a PTX file from an LLVM bitcode file.
```
pjaaskel: Please use C style comments /* */ in C files.

lib/CL/pocl_cl.h
```diff
@@ -413,6 +413,7 @@ struct _cl_device_id {
      we need to generate work-item loops to execute all the work-items
      in the WG, otherwise the hardware spawns the WIs. */
   cl_bool spmd;
+  cl_bool workgroup_pass;
```
pjaaskel: Please add a comment to this one. The workgroup pass serves multiple purposes at the moment. It generates the launcher functions and also converts the work item id global variable placeholders to context struct accesses.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -0,0 +1,679 @@
+/* pocl-cuda.c - driver for CUDA devices
+
+   Copyright (c) 2016 James Price / University of Bristol
```
pjaaskel: Is the (c) year correct?

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -0,0 +1,679 @@
...
+  ops->alloc_mem_obj = pocl_cuda_alloc_mem_obj;
+  ops->free = pocl_cuda_free;
+  ops->compile_kernel = pocl_cuda_compile_kernel;
+  ops->read = pocl_cuda_read;
+  ops->read_rect = pocl_cuda_read_rect;
+  ops->write = pocl_cuda_write;
+  ops->write_rect = pocl_cuda_write_rect;
+  ops->copy = pocl_cuda_copy;
+  ops->copy_rect = pocl_cuda_copy_rect;
+  // ops->get_timer_value = pocl_cuda_get_timer_value;
```
pjaaskel: Please avoid commented out code and also C++ style comments. If you comment out code, #if 0 is the recommended option.

lib/CL/pocl_llvm_api.cc
```diff
@@ -1579,6 +1580,12 @@ kernel_library
     is_host = false;
   }
 #endif
+#ifdef BUILD_CUDA
+  if (triple.getArch() == Triple::nvptx || triple.getArch() == Triple::nvptx64){
```
pjaaskel: Space missing before {

jrprice:
OK, these comments should be addressed now too.

I've added an initial draft of the web posting too. Let me know if it's the kind of thing you're after - if it's too long I can take stuff out.

pjaaskel:
Looks great. Final local test here (without cuda though) and then I'll push & announce.

Oblomov:
I can do some local testing on actual CUDA, but I'll need some time for that.

jrprice:
@Oblomov Further testing on other platforms would definitely be very welcome. Do let us know how you get on.

@pjaaskel Thanks for merging!

Oblomov:
Ok, managed to build and got an abort() after a failure to open libdevice library file while running my clinfo. Not impressed ;-) Joking aside, I'll look into this further, and then either open tickets here or propose pull requests, depending on time and detail availability.

Issue #430, opened 2017-02-06
Random kernel cache dir for MPI programs


noma:
I'm using PoCL within an MPI program, i.e. I have multiple communicating processes of the same program. Each of these processes (or ranks) uses PoCL to compute stuff. This scenario seems to collide with the kernel cache mechanism, as multiple processes seem to try to use the same cache directory, which leads to different kinds of non-deterministic program crashes.

It works if I use at most one process per compute node and a node-local filesystem for the cache to fully avoid conflicts. However, for performance reasons, I need more than one process per node. So I was wondering if there already is some intended way handle this scenario. One work-around I can think of would be a wrapper script around the application which gives every process an individual `POCL_CACHE_DIR` value.

pjaaskel:
Process ID in the cache dir defeats the purpose of the cache, doesn't it? Does it work if you disable cache? The proper solution would be to have file locks in the cache updates.

noma:
As a user, I don't really care about the cache. I noticed PoCL has long compilation times compared with vendor-implementations, but compared to the job runtimes, its negligible. 

The problem seems to be that POCL needs files to interact with LLVM. So setting `POCL_KERNEL_CACHE=0` does not help. According to the documentation, this only triggers deleting cache files with `clReleaseProgram()`. So I guess as long as this cannot be prevented, different cache paths per process are mandatory to avoid conflicts between processes. Otherwise I see errors ranging from segfaults in some processes, to failed asserts, related to file locking.

I currently wrap my application with a small bash script:
```bash
#!/bin/bash
EXE=my_executable
export POCL_CACHE_DIR=$(mktemp -d)
$EXE $@
```

In an ideal world, a cache with file locking etc. would lead to building everything only once across all processes of the MPI job. But I am not sure how safe/stable the whole caching is when it comes to compiling the same source with different compiler options/defines on the different processes. I remember having had some trouble which seemed like the caching mechanism used outdated compilation results instead of rebuilding. But that was some time ago, and I was too lazy to trace it back, so I just deleted the cache as a work-around.

Long story short: it might be a good addition to PoCL to have some environment variable that triggers using unique temporary cache dirs for each process (in addition to disable caching). Or maybe even fix the caching to work in an MPI-like scenario with multiple processes sharing a cache on a local/global filesystem.

This is not super urgent or a show stopper, as I can work-around it. But it's not too straight forward to find the actual problem/solution if you run into this and aren't familiar with PoCL.

pjaaskel:
[...]

PR #466, opened 2017-04-25
more bugfixes

franz:
* POCL_DEBUG messages now have categories
* cache locking should be somewhat more reliable (issue #430)
* pocl_write_file() now uses new pocl_rename()
* enable clEnqueueMigrateMemObjects(), clEnqueueBarrierWithWaitlist()
* fix memory problems in llvm_codegen by using vfork()+exec() instead of system()
* add proper refcounting to samplers
* fix buffer fill for >8byte patterns


doc/sphinx/source/env_variables.rst
```diff
@@ -33,6 +33,11 @@ pocl.
  CL_INVALID_VALUE. If clock_gettime is available, messages
  will include a timestamp.
 
+ The old way (setting POCL_DEBUG to 1) has been updated to support categories.
```
pjaaskel: Why couldn't we just print out prefixes and use the grep to filter out the interesting lines? This seems quite a lot of code for logging and I don't think it is worth the effort to not print the uninteresting ones as when one is debugging it should not be performance critical.

include/CL/cl2.hpp
```diff
@@ -5809,6 +5809,10 @@ class Kernel : public detail::Wrapper<cl_kernel>
```
pjaaskel: Can you please upstream this?

lib/CL/clGetSamplerInfo.c
```diff
@@ -31,8 +31,24 @@ POname(clGetSamplerInfo)(cl_sampler          sampler ,
...
-  return CL_SUCCESS;
+  POCL_RETURN_ERROR_COND ((sampler == NULL), CL_INVALID_SAMPLER);
+
+  switch (param_name)
+    {
+    case CL_SAMPLER_REFERENCE_COUNT:
+      POCL_RETURN_GETINFO (cl_uint, sampler->pocl_refcount);
+    case CL_SAMPLER_CONTEXT:
+      POCL_RETURN_GETINFO (cl_context, sampler->context);
+
```
pjaaskel: Extra space.

lib/CL/devices/basic/basic.c
```diff
@@ -719,14 +719,21 @@ pocl_basic_fill_rect (void *data,
                 + buffer_slice_pitch * k, fill_pixel, pixel_size);
 }
 
+typedef unsigned long v2ul __attribute__ ((vector_size (16)));
```
pjaaskel: This might be counter productive. The vector datatypes are an extension (although supported both by gcc and Clang) pls I believe the loops should get easily vectorized automatically too.

pjaaskel: Also, do you now wrongly assume the data is aligned to the vector length. The accesses through vector pointers assume natural alignment, at least with Clang. It think the original switch..case optimization might be counter productive too (whoever implemented that). Did you measure the impact on various machines with autovectorizing compilers and CPUs with SIMD instructions?

pjaaskel: I mean not to say that you should measure and optimize this spot, I just have serious doubts that it makes sense to hand vectorize it here as it's not future proof and autovectorizers are surprisingly good these days.

franz: 1) If you want i can revert to nested for loops.

2) AFAICT we check the alignment, we allocate buffers aligned, and  clEnqueueFillBuffer() requires both size & offset to be multiple of sizeof(pattern).

3) I did implement the original switch/case, what do you find counterproductive about it ?

4) no i did not measure anything, the code with vectors is simpler / shorter. I saw  no reason to not use vectors.

doc/sphinx/source/env_variables.rst
franz: 1) POCL_DEBUG writes to stderr, not stdout (this might be useful sometimes), is there a way to redirect stderr only to grep ?

2) I've already written the code ;)

3) "program | grep" means program's stderr/stdout is a pipe, not a tty -> no colors. Colors make it (for me) a lot easier to find what i need in the debug output...

But you can use grep on prefixes after this commit, though the non-tty output is slightly different.

pjaaskel: OK. Should be fine. The flag system caught my eye as it sort of "breaks the modularity" bringing things as "hsa" and "tce" to a central piece of code. But this is not a big deal as it's only a logger.

lib/CL/devices/basic/basic.c
pjaaskel: If you did not measure any major speedups on any platform, the original loops version might overall ne better. The biggest improvement from this comes from the fact that you communicate to the compiler that the data is aligned (which it otherwise cannot tell). Anyways, proper vectorizers will specialize or add a prologue in that case. Thus, I'd use a clean for loop here instead or a libc function (even better), if it can be implemented with such.

PR #468, opened 2017-04-28
Add support for LLVM 5.0 (trunk)

jrprice:
Other than the usual minor API changes, the main change here is that the various `*->dump()` methods are no longer available in release builds, so I had to guard a bunch of uses of this throughout the LLVM passes.

There are only two regressions for me on Linux at present:
- `test_clGetKernelArgInfo` fails
- `EinsteinToolkit` hangs during program build


lib/llvmopencl/Flatten.cc
```diff
@@ -75,12 +76,19 @@ Flatten::runOnModule(Module &M)
...
           (KernelName == "" && pocl::Workgroup::isKernelToProcess(*f)))
         {
-          AttributeSet attrs;
-          f->removeAttributes(
-              AttributeSet::FunctionIndex, 
-              attrs.addAttribute
-              (M.getContext(), 
-               AttributeSet::FunctionIndex, Attribute::AlwaysInline));
+#if LLVM_OLDER_THAN_5_0
+        AttributeSet attrs;
```
pjaaskel: Is there an indentation problem here? Also pls rename 'attrs' to Attrs as per LLVM coding style.

lib/llvmopencl/Barrier.h
```diff
@@ -60,8 +62,11 @@ namespace pocl {
 
       llvm::Function *F = llvm::cast<llvm::Function>
         (M->getOrInsertFunction(BARRIER_FUNCTION_NAME,
-                                llvm::Type::getVoidTy(M->getContext()),
-                                NULL));
+                                llvm::Type::getVoidTy(M->getContext())
+#if LLVM_OLDER_THAN_5_0
```
pjaaskel: It might be cleaner (in terms of readability) to duplicate the whole function call instead of guarding the single parameter.

lib/llvmopencl/ParallelRegion.cc
```diff
@@ -197,17 +197,18 @@ ParallelRegion::chainAfter(ParallelRegion *region)
       tail = region->at(region->size() - 2);
       t = tail->getTerminator();
     }
-  if (t->getNumSuccessors() != 1)
-    {
+#ifndef NDEBUG
```
pjaaskel: It would be better use a macro that tells if LLVM is built in the debug mode? As NDEBUG is unrelated to the LLVM build here, right?

jrprice:
Updated to address your comments.

PR #469, opened 2017-04-28
Fix AutomaticLocals breaking constant expressions

### LINKED ISSUE #467: AutomaticLocals pass breaks when local ptr is used in const expression
A kernel that contains something like this:

    local char foo[3];
    results[0] = (ulong)foo; // cast local pointer to integer

Generates LLVM like this:

    @test.foo = internal global [3 x i8] undef, align 1
    ...
    store i64 ptrtoint ([3 x i8]* @test.foo to i64), i64* %results, align 8, !tbaa !6

The `AutomaticLocals` pass then converts `@test.foo` into a kernel argument and replaces uses of `@test.foo` with this new argument. The problem is, a function argument is not a constant, and so it cannot be used in a constant expression like the `ptrtoint` above.

With LLVM built in Debug mode, this shows up as an assertion:

    Assertion failed: (isa<X>(Val) && "cast<Ty>() argument of incompatible type!"), function cast, file ../../include/llvm/Support/Casting.h, line 236.

jrprice:
Replacing an automatic local with an argument doesn't work if one of the
users is a constant expressions, since function arguments aren't
constants. Instead, we first replace any constant expression users with
their equivalent instructions.

Fixes #467.

### DIFF
--- a/lib/CL/pocl_llvm_api.cc
+++ b/lib/CL/pocl_llvm_api.cc
@@ -1376,7 +1376,6 @@ static PassManager& kernel_compiler_passes
   passes.push_back("workitem-handler-chooser");
   passes.push_back("mem2reg");
   passes.push_back("domtree");
-  passes.push_back("break-constgeps");
   if (device->autolocals_to_args)
 	  passes.push_back("automatic-locals");
   passes.push_back("flatten");
--- a/lib/llvmopencl/AutomaticLocals.cc
+++ b/lib/llvmopencl/AutomaticLocals.cc
@@ -143,6 +143,17 @@ AutomaticLocals::processAutomaticLocals(Function *F) {
       Locals.push_back(&*i);
       // Add the parameters to the end of the function parameter list.
       Parameters.push_back(i->getType());
+
+      // Replace any constant expression users with an equivalent instruction.
+      // Otherwise, the IR breaks when we replace the local with an argument.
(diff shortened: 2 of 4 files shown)


pjaaskel:
BreakConstantGEPs was previously used for this. Could it be expanded instead of adding the fix inside AutomaticLocals?

pjaaskel:
Hmm. Looking at BreakConstantGEPs, its code probably could be now simplified if there's an LLVM API call to get the equivalent normal instruction now.

jrprice:
Is this automatic locals problem the only reason the BreakConstantGEPs pass exists? If so, wouldn't it be better to just handle this issue inside the AutomaticLocals pass and get rid of the BreakConstantGEPs pass completely? We don't need to convert *all* constant expressions to instructions, just the ones that use automatic locals. Also, we only want to do this if we are running the AutomaticLocals pass at all; if we're not then this is just wasted work.

pjaaskel:
I believe it's the only reason it exists. At least I cannot remember of another reason. In that case, yes, you are right it makes more sense to do it there and only for the autolocals that are affected by it. Perhaps a good patch is to disable BreakConstantGEPs for now (but not yet remove it) and replace it with your code, and we can see if any of the bots encounter problems.

jrprice:
OK, I've now done this. No regressions on my machines so far.

PR #473, opened 2017-05-08
Probe for CUDA devices

### LINKED ISSUE #472: CUDA support should not abort if device is not found or functional
Currently, `pocl_cuda_init` aborts if an issue if found during device initialization. This prevents any other OpenCL device (handled by either pocl or other platforms) from being used. The correct way of handling this should be to simply not expose the CUDA device in pocl if initialization failed.

Oblomov:
This fixes #472 and additionally introduces the possibility to use (or at least access) all available devices.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -111,19 +111,23 @@ pocl_cuda_init_device_ops (struct pocl_device_ops *ops)
   /* free_event_data */
 }
 
+/* The CUDA device to be initialized next. This is reset when probing,
+ * and incremented every time a device gets initialized.
+ * TODO this is a bit of a hack, which is also shared by HSA. It might
+ * be a good idea to consider extending the init() function to also
+ * get a progressive index, and use that instead */
+static int next_cuda_device = 0;
+
 void
 pocl_cuda_init (cl_device_id dev, const char *parameters)
 {
   CUresult result;
 
-  result = cuInit (0);
-  CUDA_CHECK (result, "cuInit");
-
   if (dev->data)
     return;
 
   pocl_cuda_device_data_t *data = malloc (sizeof (pocl_cuda_device_data_t));
-  result = cuDeviceGet (&data->device, 0);
+  result = cuDeviceGet (&data->device, next_cuda_device++);
   CUDA_CHECK (result, "cuDeviceGet");
 
   // Get specific device name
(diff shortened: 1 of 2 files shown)


pjaaskel:
@jrprice pls review

Oblomov:
Rebased on current master, and removed a spurious comment that crept in during my tests.

BTW, the first two commits are not really CUDA-specific, so you may want to cherry pick them regardless.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -111,19 +111,23 @@ pocl_cuda_init_device_ops (struct pocl_device_ops *ops)
   /* free_event_data */
 }
 
+/* The CUDA device to be initialized next. This is reset when probing,
+ * and incremented every time a device gets initialized.
+ * TODO this is a bit of a hack, which is also shared by HSA. It might
+ * be a good idea to consider extending the init() function to also
+ * get a progressive index, and use that instead */
+static int next_cuda_device = 0;
```
jrprice: Agree with the comment - this is bit hacky and it would be better if the index was passed to `init()`. I'm fine with this for now though if HSA also does this.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -246,15 +250,21 @@ pocl_cuda_init_device_infos (struct _cl_device_id *dev)
...
 pocl_cuda_probe (struct pocl_device_ops *ops)
 {
   int env_count = pocl_device_get_env_count (ops->device_name);
 
-  // TODO: Check how many CUDA device available (if any)
-
   if (env_count < 0)
-    return 1;
+    {
+      CUresult ret = cuInit(0);
```
jrprice: This needs to be before the conditional block, otherwise if we run with `POCL_DEVICES=CUDA` then `cuInit` is never called.

Also use GNU C style (space before `(`) - same for next 3 lines.

Oblomov:
> I notice that if we only have one device and run with `POCL_DEVICES="CUDA CUDA"` it will abort.

Arguably, erroring out is not a bad choice, since the user requested devices which are not available.

Other approaches possible:

* expose `min(requested, available)` CUDA devices;
* if there are less available devices than requested, propose the _same_ devices again (e.g. if there are two devices, and the user has `POCL_DEVICES="CUDA CUDA CUDA CUDA"`, then expose devices 0, 1, 0, 1).

jrprice:
The `pthread` backend does the latter (expose the CPU multiple times), but I think in this case erroring out is fine. It might be good to make the error more user-friendly though (and possible not abort?), currently we just get `CUDA_ERROR_INVALID_DEVICE: invalid device ordinal`.

Oblomov:
OK, pushed. Now we abort with a more sensible error message.

PR #475, opened 2017-05-14
[RFC] Cleanup init devices

Oblomov:
I dislike it when platforms `abort()` on error, since this prevents the invoking application from properly handling error conditions. This is particularly damning on initialization, which is actually one of the easiest situations to handle (for example by changing platform).

This patch series proposes a method to handle error condition during initialization (essentially, marking problematic devices as unavailable instead of simply killing everything). It is not complete, since I'm not particularly familiar with the TCE and HSA code and I'm not in the condition to test things.

### DIFF
--- a/lib/CL/devices/devices.c
+++ b/lib/CL/devices/devices.c
@@ -187,7 +187,7 @@ pocl_string_to_dirname(char *str)
     }
 }
 
-void 
+cl_int
 pocl_init_devices()
 {
   static unsigned int init_done = 0;
@@ -206,7 +206,7 @@ pocl_init_devices()
      infinite loop. */
 
   if (init_in_progress)
-      return;
+      return CL_SUCCESS; /* debatable, but what else can we do ? */
   init_in_progress = 1;
 
   if (init_done == 0)
@@ -215,7 +215,7 @@ pocl_init_devices()
   if (init_done) 
     {
       POCL_UNLOCK(pocl_init_lock);
-      return;
+      return pocl_num_devices ? CL_SUCCESS : CL_DEVICE_NOT_FOUND;
     }
 
   /* Set a global debug flag, so we don't have to call pocl_get_bool_option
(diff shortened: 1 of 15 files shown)


lib/CL/devices/pthread/pthread.c
```diff
@@ -195,21 +197,28 @@ pocl_pthread_init (unsigned j, cl_device_id device, const char* parameters)
...
+  if (d == NULL)
+    return CL_OUT_OF_HOST_MEMORY;
 
 #ifdef CUSTOM_BUFFER_ALLOCATOR
   if (mrm == NULL)
     {
       mrm = malloc (sizeof (mem_regions_management));
+      if (mrm == NULL)
+        {
+          free(d);
```
pjaaskel: Please add white space befor parenthesis.

lib/CL/devices/devices.c
```diff
@@ -206,7 +206,7 @@ pocl_init_devices()
      infinite loop. */
 
   if (init_in_progress)
-      return;
+      return CL_SUCCESS; /* debatable, but what else can we do ? */
```
pjaaskel: This is here because of recursive initialization that happens due to hwloc's OpenCL plugin: When we initialize hwloc, it initializes OpenCL which can be pocl, etc.

pjaaskel:
HSA fails to build. Please fix:
```
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c: In function ‘pocl_hsa_init_device_infos’:
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c:462:5: error: ‘j’ undeclared (first use in this function)
   assert (j < found_hsa_agents);
     ^
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c:462:5: note: each undeclared identifier is reported only once for each function it appears in
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c: At top level:
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c:603:1: error: conflicting types for ‘pocl_hsa_init’
 pocl_hsa_init (unsigned j, cl_device_id device, const char* parameters)
 ^
In file included from /home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.c:71:0:
/home/visit0r/src/pocl-public/lib/CL/devices/hsa/pocl-hsa.h:32:663: note: previous declaration of ‘pocl_hsa_init’ was here
 GEN_PROTOTYPES (hsa)
```

lib/CL/devices/devices.c
Oblomov: I know. My question is which would be the most correct return value in this case. I went with CL_SUCCESS, but maybe something else would be more apt.

Oblomov:
Ah, fixing the HSA build I think requires a change also to init_device_infos. Not a problem, I guess.

lib/CL/devices/devices.c
pjaaskel: I think success is fine here.

Issue #478, opened 2017-05-29
Missing retain/release for event dependencies


jrprice:
When a command is enqueued with a dependency on an event, the ref count of that event should be incremented, and decremented again once the depending command has completed (or at least finished checking its dependencies). If we don't do this, that event might be destroyed before the command scheduler gets a chance to look at the dependencies (which is happening with an asynchronous implementation of command scheduling I'm working on for the CUDA backend).

The fix I want to apply here is to simply increment the event ref count in `pocl_create_event_sync`, and then call `clReleaseEvent` in `pocl_broadcast` ([see this patch](https://gist.github.com/jrprice/4dca8971102bb0ba2d3d342db96f7f2e)).

This works when I do the equivalent in the CUDA backend (which doesn't use `pocl_broadcast` in my new version), but for the pthread backend this hangs since that `clReleaseEvent` tries to lock an event that is already locked by the same thread (I think).

Using `PTHREAD_MUTEX_RECURSIVE` to allow a thread to re-lock a mutex that it has already locked would potentially solve this, but might be overkill.

Is there a better solution to this?

franz:
The better solution is to scrap recursive processing in pthread / basic drivers, and convert all of it to queues. I doubt recursive mutex will make it more reliable.

jrprice:
It looks like this might not actually be an issue for basic/pthread anyway, due to how they handle dependencies. An event removes itself from the wait lists of any dependent commands upon completion (before being destroyed), so the commands themselves don't actually check their dependencies explicitly.

If this is the case, I can just deal with the issue entirely inside the CUDA backend without breaking anything for any other drivers.

pjaaskel:
@vkorhonen comments?

pjaaskel:
@franz please explain your idea a bit better: "convert all of it to queues"? Remember that the task scheduling framework has to support cross-device distributed (hostless) synchronization with devices participating in the execution that might be more or less black boxes to the command scheduler (which need the host to synchronize their parts).

Oblomov:
FWIW, I've had issues similar to those @jrprice mentioned even in other platforms (at least Intel's CPU proprietary one). I've gone through the OpenCL documentation a couple of times, and I see no explicit mentioning of retain/release being needed for the events in a waitlist. I must say I did find it a bit odd because the standard is typically very specific about which commands do an automatic retain/release.

vkorhonen:
[...]

PR #479, opened 2017-06-02
[CUDA] Asynchronous command submission

jrprice:
This patch makes the CUDA backend deal with command submission asynchronously. This enables commands to be overlapped when using multiple devices, or with multiple command queues targeting the same device (e.g. overlapping compute with data transfers). Event dependencies are now handled using CUDA events instead of blocking on the host.

The `events` category of the Khronos OpenCL conformance test suite now passes for the CUDA backend. I've also added a small section to the CUDA docs that list all the conformance test categories that pass.

There's a small handful of changes and bug fixes outside of the CUDA backend, but I don't think there's anything controversial.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -991,15 +991,8 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq)
   event_node *dep = NULL;
   LL_FOREACH (node->event->wait_list, dep)
     {
-      /* Try and lock dependency event */
-      int ret = 0;
-      while ((ret = pthread_mutex_trylock (&dep->event->pocl_lock)))
-        {
-          /* If it is in the process of completing, just skip it */
-          if (dep->event->status == CL_COMPLETE)
-              break;
-        }
-      if (ret)
+      /* If it is in the process of completing, just skip it */
+      if (dep->event->status <= CL_COMPLETE)
         continue;
 
       /* Add CUDA event dependency */
@@ -1028,8 +1021,6 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq)
 
           event_data->num_ext_events++;
         }
-
-      pthread_mutex_unlock (&dep->event->pocl_lock);
     }
 
   /* Wait on flag for external events */
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -928,13 +928,13 @@ pocl_cuda_submit_kernel (CUstream stream, _cl_command_run run,
           }
         case POCL_ARG_TYPE_IMAGE:
         case POCL_ARG_TYPE_SAMPLER:
-          POCL_ABORT ("Unhandled argument type for CUDA");
+          POCL_ABORT ("Unhandled argument type for CUDA\n");
           break;
         }
     }
 
   if (constantMemBytes > constant_mem_size)
-    POCL_ABORT ("[CUDA] Total constant buffer size %u exceeds %lu allocated",
+    POCL_ABORT ("[CUDA] Total constant buffer size %u exceeds %lu allocated\n",
                 constantMemBytes, constant_mem_size);
 
   unsigned arg_index = kernel->num_args;
@@ -1021,7 +1021,13 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq)
             }
         }
       else
-        event_data->num_ext_events++;
+        {
+          if (!((pocl_cuda_queue_data_t *)cq->data)->use_threads)
+            POCL_ABORT (
+                "Can't handle non-CUDA dependencies without queue threads\n");
+
+          event_data->num_ext_events++;
+        }
 
       pthread_mutex_unlock (&dep->event->pocl_lock);
     }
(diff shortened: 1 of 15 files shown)


lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -557,34 +660,42 @@ pocl_cuda_copy_rect (void *data, const void *__restrict const src_ptr,
...
+pocl_cuda_submit_map_mem (CUstream stream, void *buf_ptr, size_t offset,
+                          size_t size, void *host_ptr)
+{
+  assert (host_ptr != NULL);
 
-  void *ptr = malloc (size);
-  CUresult result = cuMemcpyDtoH (ptr, (CUdeviceptr) (buf_ptr + offset), size);
-  CUDA_CHECK (result, "cuMemcpyDtoH");
-  return ptr;
+  // TODO: Map instead of copy?
```
pjaaskel: Please use old C style comments.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -557,34 +660,42 @@ pocl_cuda_copy_rect (void *data, const void *__restrict const src_ptr,
...
+                            void *device_start_ptr, size_t offset, size_t size)
 {
   if (host_ptr)
     {
-      // TODO: offset?
-      CUresult result = cuMemcpyHtoD (
-          (CUdeviceptr) (device_start_ptr + offset), host_ptr, size);
-      CUDA_CHECK (result, "cuMemcpyHtoD");
-      free (host_ptr);
+      // TODO: Only copy back if mapped for writing
```
pjaaskel: Ditto.

jrprice:
Fixed comments - all C-style now in that file.

Also snuck in one more low-hanging conformance test fix.

pjaaskel:
There are some failing pthread tests (at least with llvm-trunk):
```
83:runtime/clCreateSubDevices
86:runtime/test_user_event
115:AMD_30_AdvancedConvolution
117:AMD_30_BinomialOption
118:AMD_30_BinomialOptionMultiGPU
119:AMD_30_BitonicSort
120:AMD_30_BlackScholes
121:AMD_30_BlackScholesDP
122:AMD_30_BoxFilter
124:AMD_30_ConcurrentKernel
125:AMD_30_DCT
128:AMD_30_DwtHaar1D
129:AMD_30_FastWalshTransform
130:AMD_30_FloydWarshall
139:AMD_30_LUDecomposition
140:AMD_30_MatrixMulImage
141:AMD_30_MatrixMultiplication
142:AMD_30_MatrixTranspose
144:AMD_30_MonteCarloAsian
145:AMD_30_MonteCarloAsianMultiGPU
147:AMD_30_PrefixSum
149:AMD_30_RadixSort
150:AMD_30_RecursiveGaussian
151:AMD_30_Reduction
153:AMD_30_SimpleConvolution
156:AMD_30_SobelFilter
157:AMD_30_StringSearch
158:AMD_30_Template
160:AMD_30_URNG
```

jrprice:
Should be fixed now.

I've also addressed the issues we were talking about #478 by adding a second background thread to handle command completion. This also greatly simplifies a lot of the code, and we no longer need to explicitly check the event wait list. The CUDA backend now behaves a lot more like the other device drivers, relying on the broadcast/notify approach to signal event completion.

vkorhonen:
It seems that notifications and broadcasting are doing what they supposed to. Except for notify seems to assume that notifier event is not from this particular cuda device (`assert (event_data->num_ext_events > 0);`). The original idea is that all events are notified, so the device may notify itself. I do not see any events being removed from the wait lists. I might have misunderstood something, because you should have hit that assert in many cases already, assuming my assumption is correct.

In function `pocl_cuda_submit_node` you might want to lock the dep event before checking its status and unlocking it after `event_data->num_ext_events++;` to ensure that increment happens before dep broadcast. That `num_ext_events` should probably be volatile since it is possibly updated by another driver thread.

Submit/finalize-threads are continuously running full speed. `pthread_cond_wait` could be used there to suspend the threads when there is nothing to do. Also wait_event could use condition variable (see pthread implementation).

I am not sure if mem migration works. I assume that `cuMemcpyDtoDAsync` copies data around inside a device and the purpose of the migrate is to transfer mem object from another device/host to this device. So you might want to use `pocl_migrate_mem_objects` which uses this device drivers copy function if src and the dst are in the same global memory or else it uses appropriate drivers APIs to read from src device to host memory and then write to dst device. 
Actually the same applies to all copy commands, it may be that src dev != dst dev. This is quite important to do correctly if someone needs to implement for example cpu<->gpu pipelined application.

jrprice:
Thanks for the feedback @vkorhonen.

I've fixed the issue regarding notifications from CUDA devices to themselves in PR #485. That PR also deals with the memory migration issue you raised, although it still won't work if one of the devices is using non-host memory.

> In function pocl_cuda_submit_node you might want to lock the dep event before checking its status and unlocking it after event_data->num_ext_events++; to ensure that increment happens before dep broadcast. That num_ext_events should probably be volatile since it is possibly updated by another driver thread.

IIRC this introduces a deadlock, but actually shouldn't be necessary. The dep event won't be able to broadcast to this node while we're in the process of incrementing `num_ext_events` because it would have to lock this event first. This was really awkward to reason about but I *think* it works.

> Submit/finalize-threads are continuously running full speed. pthread_cond_wait could be used there to suspend the threads when there is nothing to do. Also wait_event could use condition variable (see pthread implementation).

I've made a note of this and will give it a go when I've regained access to some systems where I can test this properly in a multi-device set-up.

pjaaskel:
Is this done?

PR #483, opened 2017-06-26
Field program->compiler_options is not free-ed in clReleaseProgram

hominhquan:
`program->compiler_options` is set to `modded_options` in `clBuildProgram`, when `modded_options` is result of a sequence of calloc and realloc in the `APPEND_TOKEN` macro.

### DIFF
--- a/lib/CL/clBuildProgram.c
+++ b/lib/CL/clBuildProgram.c
@@ -596,7 +596,6 @@ ERROR:
   POCL_MEM_FREE(unique_devlist);
   pocl_cache_release_lock(write_cache_lock);
 ERROR_CLEAN_OPTIONS:
-  POCL_MEM_FREE(modded_options);
   program->build_status = CL_BUILD_ERROR;
 
   POCL_UNLOCK_OBJ(program);
--- a/lib/CL/clReleaseProgram.c
+++ b/lib/CL/clReleaseProgram.c
@@ -97,10 +97,7 @@ POname(clReleaseProgram)(cl_program program) CL_API_SUFFIX__VERSION_1_0
         }
 
       POCL_MEM_FREE(program->build_hash);
-      if (program->compiler_options)
-        {
-          POCL_MEM_FREE(program->compiler_options);
-        }
+      POCL_MEM_FREE(program->compiler_options);
       POCL_MEM_FREE(program->llvm_irs);
       POCL_MEM_FREE(program);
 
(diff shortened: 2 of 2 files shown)


franz:
No need for the if condition, `free()` can take NULL pointers.

 `man 3 free` If ptr is NULL, no operation is performed.

otherwise LGTM.

hominhquan:
I removed the if-condition. However, somewhere in my head still tell me that there may be a system where free(NULL) is not totally safe. Some offensive programming guides recommend adding the if-condition to be safe.

pjaaskel:
Thanks for the PR,but it causes a double free with test_clBuildProgram. Please check.

pjaaskel:
Here's how to run the tests: http://portablecl.org/docs/html/development.html#testsuite

hominhquan:
Hi @pjaaskel, the double free come from the test with the failing build. 

In case of build fails, the label ERROR_CLEAN_OPTIONS in clBuildProgram.c frees up the `modded_options` without resetting program->compiler_options to NULL, which will obviously cause double-free in clReleaseProgram in the next test. 

Resetting `compiler_options` to NULL: 
`POCL_MEM_FREE(modded_options);`
`program->compiler_options = NULL;`
fixes the double-free. 

However, I'm confusing about free-ing the `modded_options` in case of building fails, because user can later query on this by `clGetProgramBuildInfo (CL_PROGRAM_BUILD_OPTIONS)` (his own options) and he will find `program->compiler_options` non-NULL pointed to something already free-ed => segfault.

We have two choices : 
1. In case of fails (ERROR_CLEAN_OPTIONS in clBuildProgram.c) : do not POCL_MEM_FREE(modded_options), but POCL_MEM_FREE(program->compiler_options) in clReleaseProgram.c

2. Or keeping POCL_MEM_FREE(modded_options) + resetting compiler_options to NULL, but user wil lose trace to his build-options. 

What's your opinion ?

pjaaskel:
I'm not sure if it's allowed to query the build info on a program of which build has failed, thus is this undefined by the specs?

hominhquan:
I think its is allowed, since, in spec, `clGetProgramBuildInfo` can return CL_BUILD_ERROR on the program when we query for CL_PROGRAM_BUILD_STATUS.

pjaaskel:
OK. Then your 1. sounds the more sensible option, right?

hominhquan:
Unless we have any specific reason for free-ing modded_options in clBuildProgram, I think 1. is the most appropriate.

PR #493, opened 2017-07-26
Next

franz:
* properly handle -cl-ext for LLVM 4+
* implement get_image_array_size
* cmake: disable tests that take too long on CPU, enable some kernel tests (convert_type) that should be enabled
* subdevices now pass conformance tests
* partially revert the flatten-only-globals pass since it was causing significant performance regressions. We still need the globals-only for images, so detect image reads/writes and don't flatten programs with those.
* refactor pocl llvm codegen / gen wg function. The way they were written, they were holding a write lock during entire gen_wg()+codegen(), which was causing some heavily threaded tests to timeout on acquire-lock. I've changed the code to build everything in memory, and write out all files just before linking.
* some performance tuning (focused on better scaling for multicore machines).

Some benchmarks on AMD SDK 3.0, relative to pocl 0.14 _ceteris partybus_ (8C/16T cpu, llvm 4.0, kernel 4.11) (+improvement, -regression):

BitonicSort +116%
FastWalshTransform -2%
FloydWarshall +350% kernel, +136% total
MatrixTranspose +1%
MatrixMultiplication +11%
NBody +42%
RadixSort +140%
SimpleConvolution Separable Filter +44%

CPU power management is too clever these days, even with performance governor, so results might be few % off. Worst regression seems to be FastWalshTransform (-2%) so far, but i haven't run the full AMD SDK tests yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,7 @@ message(STATUS "Hwloc_CFLAGS ${Hwloc_CFLAGS}")
 
 ######################################################################################
 
-if(NOT CACHELINE_SIZE)
+if(NOT HOST_CPU_CACHELINE_SIZE)
 
   set(CL_SIZE 0)
   if(UNIX)
@@ -217,15 +217,15 @@ if(NOT CACHELINE_SIZE)
         # getconf sometimes just returns zero
         if(NOT (CL_SIZE EQUAL 0))
           message(STATUS "L1D Cacheline size detected: ${CL_SIZE}")
-          set(CACHELINE_SIZE "${CL_SIZE}" CACHE STRING "L1D Cacheline size")
+          set(HOST_CPU_CACHELINE_SIZE "${CL_SIZE}" CACHE STRING "L1D Cacheline size")
         endif()
       endif()
     endif()
   endif()
 
   if(CL_SIZE EQUAL 0)
-    message(WARNING "Unable to detect cacheline size - assuming 64byte cacheline, override with -DCACHELINE_SIZE=<number>")
-    set(CACHELINE_SIZE "64" CACHE STRING "L1D Cacheline size")
+    message(WARNING "Unable to detect cacheline size - assuming 64byte cacheline, override with -DHOST_CPU_CACHELINE_SIZE=<number>")
+    set(HOST_CPU_CACHELINE_SIZE "64" CACHE STRING "L1D Cacheline size")
   endif()
 endif()
 
@@ -1232,4 +1232,4 @@ MESSAGE(STATUS "Kernel caching: ${KERNEL_CACHE_DEFAULT}")
 MESSAGE(STATUS "Kernel library CPU variants: ${KERNELLIB_HOST_CPU_VARIANTS}")
 MESSAGE(STATUS "Kernel library distro build: ${KERNELLIB_HOST_DISTRO_VARIANTS}")
 MESSAGE(STATUS "Use fake address space IDs: ${POCL_USE_FAKE_ADDR_SPACE_IDS}")
-MESSAGE(STATUS "L1d cacheline size: ${CACHELINE_SIZE}")
+MESSAGE(STATUS "L1d cacheline size: ${HOST_CPU_CACHELINE_SIZE}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,6 +201,35 @@ message(STATUS "Hwloc_LDFLAGS ${Hwloc_LDFLAGS}")
 message(STATUS "Hwloc_CFLAGS ${Hwloc_CFLAGS}")
 
 ######################################################################################
+
+if(NOT CACHELINE_SIZE)
+
+  set(CL_SIZE 0)
+  if(UNIX)
+    find_program(GETCONF "getconf")
+    if(GETCONF)
+      execute_process(COMMAND "getconf" "LEVEL1_DCACHE_LINESIZE"
+                      RESULT_VARIABLE RES OUTPUT_VARIABLE CL_SIZE)
+      if(RES)
+        message(WARNING "getconf exited with nonzero status!")
+        set(CL_SIZE 0)
+      else()
+        # getconf sometimes just returns zero
+        if(NOT (CL_SIZE EQUAL 0))
+          message(STATUS "L1D Cacheline size detected: ${CL_SIZE}")
+          set(CACHELINE_SIZE "${CL_SIZE}" CACHE STRING "L1D Cacheline size")
+        endif()
+      endif()
+    endif()
+  endif()
+
+  if(CL_SIZE EQUAL 0)
+    message(WARNING "Unable to detect cacheline size - assuming 64byte cacheline, override with -DCACHELINE_SIZE=<number>")
+    set(CACHELINE_SIZE "64" CACHE STRING "L1D Cacheline size")
+  endif()
+endif()
+
+######################################################################################
 #
 # Find executables to few tools required during build 
 #
@@ -1203,3 +1232,4 @@ MESSAGE(STATUS "Kernel caching: ${KERNEL_CACHE_DEFAULT}")
 MESSAGE(STATUS "Kernel library CPU variants: ${KERNELLIB_HOST_CPU_VARIANTS}")
 MESSAGE(STATUS "Kernel library distro build: ${KERNELLIB_HOST_DISTRO_VARIANTS}")
 MESSAGE(STATUS "Use fake address space IDs: ${POCL_USE_FAKE_ADDR_SPACE_IDS}")
+MESSAGE(STATUS "L1d cacheline size: ${CACHELINE_SIZE}")
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -12,7 +12,7 @@
 /* "Build with ICD" */
 #cmakedefine BUILD_ICD
 
-#define CACHELINE_SIZE @CACHELINE_SIZE@
+#define HOST_CPU_CACHELINE_SIZE @HOST_CPU_CACHELINE_SIZE@
 
 #cmakedefine CLANG_IS_PATCHED_FOR_SPIR_CC
 
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -12,6 +12,8 @@
 /* "Build with ICD" */
 #cmakedefine BUILD_ICD
 
+#define CACHELINE_SIZE 64
+
 #cmakedefine CLANG_IS_PATCHED_FOR_SPIR_CC
 
 #define CLANG "@CLANG@"
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -12,7 +12,7 @@
 /* "Build with ICD" */
 #cmakedefine BUILD_ICD
 
-#define CACHELINE_SIZE 64
+#define CACHELINE_SIZE @CACHELINE_SIZE@
 
 #cmakedefine CLANG_IS_PATCHED_FOR_SPIR_CC
 
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -632,7 +632,7 @@ pocl_broadcast (cl_event brc_event)
   event_node *tmp;
   while ((target = brc_event->notify_list))
     {
-      POCL_LOCK_OBJ (target->event);
+      pocl_lock_events_inorder (brc_event, target->event);
       /* remove event from wait list */
       LL_FOREACH (target->event->wait_list, tmp)
         {
@@ -645,14 +645,12 @@ pocl_broadcast (cl_event brc_event)
         }
       if (target->event->status == CL_SUBMITTED)
         {
-          POCL_UNLOCK_OBJ (target->event);
           target->event->command->device->ops->notify
             (target->event->command->device, target->event, brc_event);
         }
-      else 
-        POCL_UNLOCK_OBJ (target->event);
       
       LL_DELETE (brc_event->notify_list, target);
+      pocl_unlock_events_inorder (brc_event, target->event);
       pocl_mem_manager_free_event_node (target);
     }
 }
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -294,7 +294,7 @@ pocl_exec_command (_cl_command_node * volatile node)
 {
   unsigned i;
   /* because of POCL_UPDATE_EVENT_ */
-  cl_event *event = &(node->event);
+  cl_event event = node->event;
   switch (node->type)
     {
     case CL_COMMAND_READ_BUFFER:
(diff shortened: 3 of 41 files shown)


config.h.in.cmake
```diff
@@ -12,6 +12,8 @@
 /* "Build with ICD" */
 #cmakedefine BUILD_ICD
 
+#define CACHELINE_SIZE 64
```
pjaaskel: What is this? Needs comments. For which machine and what it's used for etc.

lib/CL/devices/common.c
```diff
@@ -74,87 +78,160 @@ char*
...
+  void* write_lock = NULL;
+
+  cl_program program = kernel->program;
+
+  int device_i = pocl_cl_device_to_index(program, device);
+  assert(device_i >= 0);
+
+  /* $/parallel.bc */
+  char parallel_bc_path[POCL_FILENAME_LENGTH];
+  pocl_cache_work_group_function_path(parallel_bc_path, program, device_i, kernel, local_x, local_y, local_z);
```
pjaaskel: Long line.

lib/CL/devices/common.c
```diff
@@ -74,87 +78,160 @@ char*
...
-  int file_name_alloc_size = 
-    min(POCL_FILENAME_LENGTH, strlen(tmpdir) + strlen(kernel->name) + 5);
-  char* module = (char*) malloc(file_name_alloc_size); 
-  /* To avoid corrupted .so files, create a tmp file first
-     and then rename it. */
-  char tmp_module[file_name_alloc_size + 4]; /* .tmp postfix */
-  int error;
+  /* $/kernel.so */
+  char final_binary_path[POCL_FILENAME_LENGTH];
+  pocl_cache_final_binary_path(final_binary_path, program, device_i, kernel, local_x, local_y, local_z);
```
pjaaskel: Here too.

lib/CL/devices/common.c
```diff
@@ -74,87 +78,160 @@ char*
...
-  assert (error >= 0);
+  /* $/kernel.so.o */
+  char objfile_path[POCL_FILENAME_LENGTH];
+  strcpy (objfile_path, final_binary_path);
+  strcat (objfile_path, ".o");
+  char *objfile = NULL;
+  size_t objfile_size = 0;
+
+  void* llvm_module = NULL;
+  error = pocl_llvm_generate_workgroup_function_nowrite( device, kernel,
```
pjaaskel: Check space after parenthesis.

lib/CL/devices/common.c
```diff
@@ -74,87 +78,160 @@ char*
...
 
-  memcpy (tmp_module, module, file_name_alloc_size);
-  strcat (tmp_module, ".tmp");
 
-  void* write_lock = pocl_cache_acquire_writer_lock(kernel->program, device);
-  assert(write_lock);
 
-  error = pocl_llvm_generate_workgroup_function (device, kernel,
+  /* write parallel.bc only if we want to leave compiler files*/
+  if (pocl_get_bool_option("POCL_LEAVE_KERNEL_COMPILER_TEMP_FILES", 0))
```
pjaaskel: check the white spacing in this chunk

lib/CL/devices/common.c
```diff
@@ -74,87 +78,160 @@ char*
...
+  if (error)
+    goto FINISH;
   assert (error == 0);
 
   /* Save space in kernel cache */
   if (!pocl_get_bool_option("POCL_LEAVE_KERNEL_COMPILER_TEMP_FILES", 0))
     {
-      pocl_remove(objfile);
-      pocl_remove(bytecode);
+      pocl_remove(objfile_path);
```
pjaaskel: pocl_remove (...

lib/CL/devices/pthread/pthread.c
```diff
@@ -117,6 +113,10 @@ struct data {
   mem_regions_management* mem_regions;
 #endif
 
+  /* Lock for command list related operations */
+  pthread_mutex_t cq_lock __attribute__ ((aligned (CACHELINE_SIZE)));
```
pjaaskel: OK, here CACHELINE_SIZE is used. I think that's too general. Sure 64B is a common choice, but it's not a global truth, if you know what I mean :) Better just have it as a specifically named macro PTHREAD_MUTEX_ALIGNMENT or similar if you have measured it speeds up the lock handling.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -214,19 +205,36 @@ pthread_scheduler_sleep()
 }
 
 #define POCL_PTHREAD_MAX_WGS 256
-static int get_wg_index_range (kernel_run_command *k, unsigned *start_index,
-                               unsigned *end_index, char *last_wgs)
+#define POCL_PTHREAD_MIN_WGS 32
```
pjaaskel: This macro needs a comment: what it's used for.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -214,19 +205,36 @@ pthread_scheduler_sleep()
...
+
+static int
```
pjaaskel: Also this function needs a comment.

lib/CL/pocl_llvm.h
```diff
@@ -73,6 +73,10 @@ int pocl_llvm_generate_workgroup_function(cl_device_id device,
                                           cl_kernel kernel, size_t local_x,
                                           size_t local_y, size_t local_z);
 
+int pocl_llvm_generate_workgroup_function_nowrite(cl_device_id device,
```
pjaaskel: nowrite (...

lib/CL/pocl_llvm_api.cc
```diff
@@ -1529,8 +1556,25 @@ static PassManager& kernel_compiler_passes
 	  passes.push_back("automatic-locals");
   if (SPMDDevice)
     passes.push_back("flatten-inline-all");
-  else
-    passes.push_back("flatten-globals");
+  else {
+    /* LLVM is unfortunately still not intelligent enough to flatten.
```
pjaaskel: By flatten you might mean 'inline' here. So you mean LLVM is not intelligently enough choosing what to inline if we do not perform the inline-everything (flatten) pass?

lib/CL/pocl_llvm_api.cc
```diff
@@ -1892,25 +1923,45 @@ int pocl_llvm_generate_workgroup_function(cl_device_id device,
...
+int pocl_llvm_generate_workgroup_function(cl_device_id device,
+                                          cl_kernel kernel, size_t local_x,
+                                          size_t local_y, size_t local_z) {
+
+  void *modp = NULL;
+
+  int device_i = pocl_cl_device_to_index(kernel->program, device);
+  assert(device_i >= 0);
+
+  int error = pocl_llvm_generate_workgroup_function_nowrite(device, kernel, local_x, local_y, local_z, &modp);
```
pjaaskel: Too long line.

lib/CL/pocl_util.c
```diff
@@ -64,6 +75,126 @@ typedef struct list_item
 } list_item;
 
 
+#ifdef __linux__
+#ifdef __x86_64__
```
pjaaskel: Have you tested with ARM CPUs recently, BTW?

config.h.in.cmake
franz: right, this should be detected via CMake. i'll look into it

lib/CL/devices/pthread/pthread_scheduler.c
franz: It's a few lines below in the text

lib/CL/pocl_util.c
franz: I have tested after the first commit in this PR (-cl-ext) and it did compile and run some tests (from -L internal), but not all work.

lib/CL/pocl_llvm_api.cc
franz: Yes, unfortunately. Or perhaps we're missing some pass. I've noticed this in NBody (IIRC), it had a not inlined _cl_sqrt() or some such function with only a half dozen instructions. Had a relatively big negative impact on NBody's performance.

lib/CL/devices/pthread/pthread.c
franz: Well, it's not an alignment by mutex size, it's really aligned by cacheline size :) but like i said, i'll add detection.

pjaaskel: Yes, I meant PTHREAD_OPTIMAL_MUTEX_ALIGNMENT ;) But are you sure it's worth the effort? Did you measure speedups?

franz: Yes, i have provided some numbers at the PR's "main" comment.

pjaaskel: All of the speedups were due to lock alignment really?

franz: Not all, i'd guess maybe 3/4 of it. Rest is likely from `get_wg_index_range()` tuning.

I'd say the most critical (WRT alignment) are `struct scheduler_data`, which is shared across all driver threads, and `struct pool_thread_data` which is allocated one per thread - but it was allocated via calloc() so every driver thread got something not cacheline aligned. Both these structs are heavily used by driver threads.

franz: Also not helping is the amount of volatiles (though i understand why some are required), since every volatile means fetching from memory = fighting other driver threads for cachelines :) so perhaps surprising but yes alignment does the most speedup here.

pjaaskel: OK. Surprising indeed. Volatiles should not cause accesses to main memory / cache flushs, only avoid storing it to registers, AFAIK. I think one still can cache them, cache coherence HW takes care of the consistency.

franz: > Volatiles should not cause accesses to main memory / cache

AFAICT, they may cause quite a lot of it. If e.g. we have a `struct { volatile int i; pthread_mutex_t lock; }` then reading `i` in a loop will cause an L3/memory fetch everytime some other thread does a pthread lock, because they share a cacheline and it's invalidated by the lock. So volatiles can in fact be very expensive.

pjaaskel: "Flushs" (flushes) was an important word you missed in your quote. Surely if you need to share a variable you go to lower levels of the cache, but not necessarily to the main (external) memory. 

What you are describing is the general problem of "false sharing" which has nothing to do with volatile as such, although volatile here might cause unneccessarily not keeping the i in register if you guard its updates with a critical section anyways, and makes the false sharing problem worse. 

If you don't need to share int i or you guard it with a mutex (the lock) then it should not be marked volatile, but allow compiler to utilize registers for temporary accesses _inside the critical section_. The false sharing might itself be a problem here which can be cured by padding, but dropping volatile means you don't need to share i, you use atomic updates, or guard it with the mutex (which ensures the variables are flushed from registers after the section), or you have a race.

franz: > makes the false sharing problem worse.

Yes that's exactly what i meant to say. :)

 > If you don't need to share int i or you guard it with a mutex (the lock) then it should not be marked volatile

Yes. For some reason i haven't found out, it seems there's _a lot_ of variables in structs marked volatile, even ones not guarded with a mutex. I think this _might_ have something to do with the concept of inter-device handling of events. TBH this would explain volatile in structs like cl_event, but it doesn't explain why so many pthread driver internals are volatile.

franz: OK, i have added a bit of CMake code to detect cacheline size on linux, with fallback to 64. It's also possible to define at CMake run.

franz:
OK i think everything should be handled now, all tests pass on my buildbots.

CMakeLists.txt
```diff
@@ -201,6 +201,35 @@ message(STATUS "Hwloc_LDFLAGS ${Hwloc_LDFLAGS}")
 message(STATUS "Hwloc_CFLAGS ${Hwloc_CFLAGS}")
 
 ######################################################################################
+
+if(NOT CACHELINE_SIZE)
```
pjaaskel: Can you rename this to HOST_CPU_CACHELINE_SIZE?

lib/CL/devices/pthread/pocl-pthread_utils.h
```diff
@@ -4,15 +4,27 @@
 #include "pocl_cl.h"
 
 /* locking macros */
-#define PTHREAD_LOCK(__lock , __counter)        \
-  do {                                          \
-    if ((__counter))                            \
-      ++(*((unsigned*)(__counter)));            \
-  }while (pthread_mutex_trylock((__lock)))
+#define PTHREAD_LOCK_COUNT(__lock, __counter)                                 \
```
pjaaskel: Is this used somewhere? What it's for?

lib/CL/pocl_llvm_api.cc
```diff
@@ -1527,10 +1561,37 @@ static PassManager& kernel_compiler_passes
...
+
   if (SPMDDevice)
     passes.push_back("flatten-inline-all");
-  else
-    passes.push_back("flatten-globals");
+  else {
+    /* LLVM is either still not intelligent enough to flatten,
+     * or we're missing some pass.
+     *
+     * The only reason we've reintroduced "global-only" flatten was that
```
pjaaskel: What do you mean global-only here? Could we simply rely on LLVM inliner nowadays? That is, flatten (this used to mean "inline everything") only those that have to be (functions which take in the context struct basically), but let the rest of the inlining decisions to LLVM. This should work fine for CPUs and DSPs. For GPUs we might want to inline more aggressively.

lib/CL/pocl_util.c
```diff
@@ -290,6 +292,35 @@ pocl_aligned_free (void *ptr)
 }
 #endif
 
+void
+pocl_lock_events_inorder (cl_event ev1, cl_event ev2)
```
pjaaskel: The purpose of this function is still undocumented.

lib/CL/pocl_llvm_api.cc
franz: > only those that have to be (functions which take in the context struct basically)

That's what i meant by "global-only".

> Could we simply rely on LLVM inliner nowadays?

The performance will take a hit (in some cases significant, see the entire comment), so i'd prefer to play around with LLVM inliner **after** we do the release. The way it's set up with this commit, it's at least as good as pocl 0.14 and it also works with read/write_image{f,i,ui}.

CMakeLists.txt
franz: will do

lib/CL/devices/pthread/pocl-pthread_utils.h
franz: possibly unused, i guess Ville used it for debugging. not sure, will look

lib/CL/pocl_llvm_api.cc
pjaaskel: Yup, makes sense to postpone optimization of inlining to post release forking.

PR #499, opened 2017-08-03
[CUDA] Handle free_ptr for clSVMFree

jrprice:


hahnjo:
I can confirm that this fixes the crash we have been dicussing offline 👍

Issue #488, opened 2017-07-06
Undefined symbols cause symbol lookup crash


ChrisCummins:
The following kernel should produce a compile time error for `__morldNormalided ` but instead compiles fine, then crashes the program during kernel execution:

```
__kernel void entry(__global int* a) {
  int b = __morldNormalided(a);
}
```

Host code and steps to reproduce: http://paste.ubuntu.com/25031508/

Error:
```
symbol lookup error: <pocl-cache>/entry.so: undefined symbol __morldNormalided
```

Expected output: `CL_BUILD_PROGRAM_FAILURE` during compilation.

franz:
This should be now fixed via ab37825d8eaaf547e0d56774864e056f7af9e60f

(reopen if you still find it broken)

PR #503, opened 2017-08-09
Next

franz:
* bunch of minor fixes (for errors reported by piglit & stuff)
* ARM now at least builds (bunch of tests fail)
* pocl_llvm_api.cc cleanup (move most of one-time initialization to InitializeLLVM() instead of having it all over )
* fp16 vload/vstore implementation using F16C instruction set with fallback to software. ATM vload & vstore pass tests for float <-> half but not yet double <-> half.
* move linking of kernel library from generate_wg_func (=parallel.bc) to pocl_llvm_{build,link}_program (=program.bc). Should save some time when building multiple kernels from 1 program, also allows pocl to check undefined symbols at clBuild/LinkProgram time. Bump pocl_binary version (b/c old binaries expect to be linked to kernel lib after loading).
* fix issue #488 by checking undefined symbols when linking program.bc with kernel library. Impact on build/link times should be minimal.


lib/CL/devices/devices.c
```diff
@@ -431,9 +431,11 @@ pocl_init_devices()
             pocl_devices[dev_index].available = 0;
           }
 
-          if (dev_index == 0)
-            pocl_devices[dev_index].type |= CL_DEVICE_TYPE_DEFAULT;
-
+          if (pocl_get_bool_option ("POCL_WORKAROUND_DEVTYPE", 0) == 0)
```
pjaaskel: The option should be more descriptive and documented. POCL_USE_DEFAULT_DEVICE_TYPE_FLAG or similar.

lib/CL/devices/topology/pocl_topology.c
```diff
@@ -52,7 +102,14 @@ pocl_topology_detect_device_info(cl_device_id device)
     return ret;
   }
 
-  hwloc_topology_set_flags (pocl_topology, HWLOC_TOPOLOGY_FLAG_WHOLE_IO);
+  // hwloc_topology_set_flags (pocl_topology, HWLOC_TOPOLOGY_FLAG_WHOLE_IO);
```
pjaaskel: Commented out code.

lib/CL/pocl_binary.c
```diff
@@ -52,7 +52,10 @@
 #define POCLCC_STRING_ID_LENGTH 8
 /* changes for version 2: added program.bc right after header */
 /* changes for version 3: added flush_denorms flag into header */
-#define POCLCC_VERSION 3
```
pjaaskel: I don't think we need to gracefully handle previous pocl binary versions before we have 1.0 out. But OK.

lib/CL/pocl_llvm_api.cc
```diff
@@ -137,7 +142,112 @@ static llvm::sys::Mutex kernelCompilerLock;
 /* Global pocl device to be used by passes if needed */
 cl_device_id currentPoclDevice = NULL;
 
-static void InitializeLLVM();
+static std::map<cl_device_id, llvm::TargetMachine *> target_machines;
```
pjaaskel: This file is C++, so please follow the C++ convention in new edits.

lib/CL/pocl_llvm_api.cc
```diff
@@ -137,7 +142,112 @@ static llvm::sys::Mutex kernelCompilerLock;
...
+static std::map<cl_device_id, PassManager *> kernel_passes_noflatten;
```
pjaaskel: Please describe the need for two pass managers. (to me and in a comment)

lib/CL/clSetKernelArg.c
```diff
@@ -51,20 +51,49 @@ POname(clSetKernelArg)(cl_kernel kernel,
...
+  if (pi->type == POCL_ARG_TYPE_POINTER || pi->type == POCL_ARG_TYPE_IMAGE
+      || pi->type == POCL_ARG_TYPE_SAMPLER)
+    POCL_RETURN_ERROR_ON (((!pi->is_local) && (arg_size != sizeof (cl_mem))),
+                          CL_INVALID_ARG_SIZE,
+                          "Arg %u is pointer/buffer/image, but arg_size is "
+                          "not sizeof(cl_mem)\n",
+                          arg_index);
+  else if (pi->type_size)
+    {
+      size_t as = arg_size;
```
pjaaskel: What is this handling for?

lib/CL/pocl_llvm_api.cc
```diff
@@ -1226,11 +1503,41 @@ int pocl_llvm_get_kernel_metadata(cl_program program,
                llvm::dyn_cast<ConstantAsMetadata>(
                  ReqdWGSize->getOperand(2))->getValue()))->getLimitedValue();
   }
+/* TODO
```
pjaaskel: Commented out code.

lib/CL/pocl_llvm_api.cc
```diff
@@ -1459,61 +1759,40 @@ static bool moduleReferencesImages(llvm::Module *M) {
...
+  PassRegistry *Registry = nullptr;
 
-  if (kernel_compiler_passes.find(device) != 
-      kernel_compiler_passes.end())
-    {
-      return *kernel_compiler_passes[device];
+  bool refs_images = moduleReferencesImages(input);
+
+  std::map<cl_device_id, PassManager *> &kernel_pass_map =
+      (refs_images ? kernel_passes_noflatten : kernel_passes_flatten);
```
pjaaskel: OK. It might be worth figuring out a sensible inline scheme as the workaround gets somehow "out of hands".

pjaaskel: Can you try the no-flatten by default now? I think LLVM should take care of "sensible" inlining in standard opts.

tests/tce/multi_AS_copy/host.cpp
```diff
@@ -44,7 +44,7 @@
 #include "poclu.h"
 
 static char
-kernelSourceCode[] =
+kernelSourceCodeTCE[] =
```
pjaaskel: Please try if printf() works OK with TCE too. LLVM should optimize it at least to putstring which is OK too.

lib/CL/pocl_llvm_api.cc
franz: This is because the inline method (all vs globals only) is decided per-module not per-device, depending on whether the module uses images. Once we get rid of our inline-all hack and let LLVM do the inlining, we can use just one pass manager list per device.

lib/CL/pocl_llvm_api.cc
franz: OK

lib/CL/clSetKernelArg.c
franz: Piglit checks if the clSetKernelArg accepts incorrect arg_size for an argument. This check is rather crude and obviously doesn't work for structs, but it does make piglit pass. Do you think it's unnecessary ?

lib/CL/pocl_binary.c
franz: Yep, not really a must but it's nicer to get a warning than a crash :)

lib/CL/devices/devices.c
franz: This is actually a workaround, we **do** set the CL_DEVICE_TYPE_DEFAULT by default :) this bool option disables ORing the CL_DEVICE_TYPE_DEFAULT, so software that treats device_type as enum (checks with equality) instead of bitfield, will work.

I've added some docs and renamed the option to POCL_WORKAROUND_DEVICE_TYPE_BUG.

pjaaskel: Yes, I know. But your new name is not still descriptive. "What bug?" people ask. You basically add reasons for the env / history to its name which is not interesting in longer term :)

POCL_NO_DEFAULT_DEVICE_TYPE_FLAG  (that explicitly tells what it does) or similar or the one I proposed, but with the default of 1. I'm not sure if it's mandatory according to the specs to use/return the default device flag actually. Might be not as I've seen several codes with this "bug".

franz: > "What bug?" people ask. 

The device type bug. It's literallly spelled out there. There isn't all that much bugs related to device-type :)

> POCL_NO_DEFAULT_DEVICE_TYPE_FLAG (that explicitly tells what it does)
Yep it's explicit. It just has one tiny problem, nobody will ever remember it. 

> I'm not sure if it's mandatory according to the specs to use/return the default device flag actually. 

It doesn't seem to be. We could remove it, but then we need to special case CL_DEVICE_TYPE_DEFAULT for clCreateContextFromType(). I can also remove the commit entirely (it's really just a matter of convenience).

pjaaskel: I'd just use POCL_NO_DEFAULT_DEVICE_TYPE_FLAG as its name and be done with it. There might be other bugs related to device-type in the future and we cannot assume this env workarounds them all ;-)

franz:
Rebased to master.

The FP16 library was causing troubles for both TCE and HSA, so i've decided to reimplement their code in OpenCL and drop the original code, wasn't worth the trouble.

TCE tests / examples now use printf() instead of putchar().

SPIR binaries are now expected to be fully linked.

lib/CL/clLinkProgram.c
```diff
@@ -47,7 +47,25 @@ CL_API_SUFFIX__VERSION_1_2
...
   POCL_GOTO_ERROR_COND ((input_programs == NULL), CL_INVALID_VALUE);
 
+  POCL_GOTO_ERROR_COND ((num_devices > 0 && device_list == NULL),
+                        CL_INVALID_VALUE);
+  POCL_GOTO_ERROR_COND ((num_devices == 0 && device_list != NULL),
+                        CL_INVALID_VALUE);
+
   /* DEVICE LIST */
+
+  // check binary types
```
pjaaskel: Please avoid the C++ comment.

lib/CL/clSetKernelArg.c
pjaaskel: It's OK. Better would be to check if LLVM stores type size info as MD and do it properly, but if it's not checked by the conformance suite, it's not important for now.

lib/CL/clUnloadPlatformCompiler.c
```diff
@@ -22,11 +22,25 @@
  **/
 
 #include "pocl_cl.h"
+#include "pocl_llvm.h"
 
 CL_API_ENTRY cl_int CL_API_CALL
 POname(clUnloadPlatformCompiler)(cl_platform_id platform)
 CL_API_SUFFIX__VERSION_1_2
 {
+#if defined(BUILD_ICD) && defined(OCS_AVAILABLE)
```
pjaaskel: This needs comments. Why only with BUILD_ICD?

lib/llvmopencl/linker.cpp
```diff
@@ -304,66 +305,97 @@ stringref_cmp(llvm::StringRef a, llvm::StringRef b)
...
+  // this one is a handled with a special pocl LLVM pass
+  StringRef pocl_sampler_handler("__translate_sampler_initializer");
+  // ignore undefined llvm intrinsics
+  StringRef llvm_intrins("llvm.");
   std::list<llvm::StringRef>::iterator di,de;
   for (di = declared.begin(), de = declared.end();
        di != de; di++) {
-      copy_func_callgraph(*di, lib, krn, vvm);
+      llvm::StringRef r = *di;
+#ifdef TCE_AVAILABLE
```
pjaaskel: Is this still needed? A bit ugly now that I see it here.

lib/CL/clUnloadPlatformCompiler.c
franz: Unless i'm misreading the code in clGetPlatformIDs, calling clGetPlatformIDs() only returns a valid platform when BUILD_ICD is #defined. (which TBH might be a bug)

lib/llvmopencl/linker.cpp
franz: removed

lib/CL/clUnloadPlatformCompiler.c
pjaaskel: If I understood correctly, an OpenCL implementation can provide multiple platforms even without ICD and if not, it should return the single platform ID if BUILD_ICD is not on.

lib/llvmopencl/WorkitemLoops.cc
```diff
@@ -93,6 +93,12 @@ WorkitemLoops::getAnalysisUsage(AnalysisUsage &AU) const
 
 }
 
+bool WorkitemLoops::doFinalization(llvm::Module &M) {
+  // todo what should it return
+  cleanParallelRegions();
```
pjaaskel: Why is it needed in the doFinalization()?

franz: IIRC i tried deleting original_parallel_regions at the end of runOnFunction() or ProcessFunction(), but i got segfaults. I'm not familiar enough with the code to know where is it safe to delete, but this works (at least valgrind says so). Do you have a better place in mind?

franz: Moved the cleanup to runOnFunction(), all tests seem to be passing now.

franz:
Had to shuffle around some CMake code to workaround the WILL_FAIL fail :) The two other commits are fixing some regressions caused by kernel lib linking.

PR #514, opened 2017-08-16
Next

franz:
Fixes a bunch of issues for 1.0/1.2 milestones


config.h.in.cmake
```diff
@@ -151,6 +151,9 @@
 /* Disable cl_khr_fp16 because fp16 is not supported */
 #cmakedefine _CL_DISABLE_HALF
 
+/* Disable cl_khr_fp64 because fp64 is not supported */
```
pjaaskel: Does it disable it for all devices? What about heterogeneous platforms where host does not support fp64 but the co-processor (GPU) does?

franz: No, this variable is only for host CPU (perhaps i should rename it) and only used in pthread/basic drivers. 

The other drivers each have their own supported extension lists in CMake (used @ building the kernel library). In driver code, CUDA and HSA currently  hardcode double support to fully supported, TCE hardcodes it to `CL_FP_ROUND_TO_NEAREST | CL_FP_INF_NAN`

franz: _CL_DISABLE_HALF is also detected on CPU only.

inducer:
Confirmed fixed. Thanks for working on this!

PR #533, opened 2017-09-27
CUDA attribute retrieval clean up and fixes

Oblomov:
This series cleans up CUDA attribute retrieval, and fixes a couple of minor issues.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -192,32 +192,33 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
   /* Get other device properties */
   if (ret != CL_INVALID_DEVICE)
     {
-      cuDeviceGetAttribute ((int *)&dev->max_work_group_size,
-                            CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_BLOCK,
-                            data->device);
-      cuDeviceGetAttribute ((int *)(dev->max_work_item_sizes + 0),
-                            CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_X, data->device);
-      cuDeviceGetAttribute ((int *)(dev->max_work_item_sizes + 1),
-                            CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Y, data->device);
-      cuDeviceGetAttribute ((int *)(dev->max_work_item_sizes + 2),
-                            CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Z, data->device);
-      cuDeviceGetAttribute (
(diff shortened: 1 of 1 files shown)


lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -192,34 +192,37 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
...
+      GET_CU_PROP (MAX_BLOCK_DIM_Y, dev->max_work_item_sizes[1]);
+      GET_CU_PROP (MAX_BLOCK_DIM_Z, dev->max_work_item_sizes[2]);
+      GET_CU_PROP (MAX_SHARED_MEMORY_PER_BLOCK, dev->local_mem_size);
+      GET_CU_PROP (MULTIPROCESSOR_COUNT, dev->max_compute_units);
+      GET_CU_PROP (ECC_ENABLED, dev->error_correction_support);
+      GET_CU_PROP (INTEGRATED, dev->host_unified_memory);
+      GET_CU_PROP (TOTAL_CONSTANT_MEMORY, dev->max_constant_buffer_size);
+      GET_CU_PROP (CLOCK_RATE, dev->max_clock_frequency);
       dev->max_clock_frequency /= 1000;
+      GET_CU_PROP (TEXTURE_ALIGNMENT, dev->mem_base_addr_align);
```
jrprice: I'm not sure about this one. Do we know that the texture alignment is always the same as the alignment for regular buffers?

The hardcoded 2048 bits came from the [CUDA programming guide](http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses), which states:

> Any address of a variable residing in global memory or returned by one of the memory allocation routines from the driver or runtime API is always aligned to at least 256 bytes.

Oblomov: The motivation for this change is that NVIDIA's own OpenCL platform reports 4096 bits on devices with CC 2.x or higher, while it used to report 2048 on devices with CC 1.x.

The value reported matches the one used by the CUDA driver when rounding up the horizontal pitch for 2D allocations (`cudaMallocPitch` and the driver API equivalent), which is also the one used for texture alignment (because in CUDA you can bind a 2D texture to a linear buffer too, provided the pitch is correct, i.e. a multiple of this value).

It's quite possible that NVIDIA may have misinterpreted the field meaning, or simply decided to abuse it to communicate a value that would otherwise not be possible to communicate within the OpenCL core, but it's also possible that they forgot to update that particular sentence when support for CC 1.x was dropped (which was the hardware with the 256-byte requirements).

In the end I decided to use this value for consistency with the NVIDIA proprietary platform.

jrprice: OK, that's fine with me then (it doesn't break any of the conformance tests).

Issue #537, opened 2017-10-12
pocl_binaries are not functionnal


rjodin:
Hi everyone,

I updated my version of POCL, and my test-suite fails.
It fails on a test where I compile POCL without LLVM, and I try to run a kernel created from binary.

From what I see, we added the "program.bc" inside the binary. But it seems that you forget to take it into consideration when deserializing.
I found 2 functions in "pocl_binary.c" that seems not to take care of it:
-  pocl_binary_get_kernel_metadata
-  pocl_binary_get_kernel_count

In both of them, we are reading the header and then we try to get information of kernels. But with the current way of serializing the binary, the "program.bc" is just between the header and the kernels. So we would need to increment the buffer accordingly.

And then, I have another issue related.
When my code arrives in "clCreateKernel" (with POCL compiled without LLVM) at line 85 (53ef5e89aa99b5e85ad512ff46aa8f38b36acf4f today's master), it goes into "program->binaries" as now we deserialize "program.bc" in the cache directory (we write "program->binaries" in clCreateProgramWithBinary line 164). But as I do not have LLVM, it does nothing. It would have needed to go into "program->pocl_binaries".

If you agree with all this, I will propose a patch for this issue.

pjaaskel:
Please do. I think offline compile compilation (aka no-LLVM) build is not tested regularly by my test suite so it's no wonder if it fails. If you have tests to share, I can add it to mine so it doesn't regress.

rjodin:
We already have a test that test "clCreateProgramWithBinary" in POCL, but the test uses online compilation.
I don't know how to add a test that will just be run on a build of POCL without compiler. And for this test, we would need to give a static binary that would need to be rebuild each time we are changing the way we serialize/deserialize the binary. But I agree that it is important the improve our testing set on this feature.
How could I make a test that will be run only on a build of POCL without compiler?

pjaaskel:
I don't know if there's any mechanism for that yet. Indeed it's important to have it in the test suite to avoid accidental breakage like this.

franz:
I have one buildbot which is setup to build LLVM-less pocl builds to keep that setup buildable, but it does run any tests with it.

@rjodin

> How could I make a test that will be run only on a build of POCL without compiler?

With current pocl you'll need two buildroots (from the same source) and two tests, one that produces a binary and another that loads it. The test(s) can be not/build on the OCS_AVAILABLE cmake variable The only problem with automating this, is how to tell buildbot about a dependency in two buildroots.

[...]

PR #538, opened 2017-10-12
fix issue #537 (pocl_binaries are not functionnal)

rjodin:
We need to skip program.bc in a pocl_binaries when we want to have
information on kernels from the start of a pocl_binaries.
When running POCL offline, we do not need to take care of
program->binaries (even if they are present) as we do not have a
compiler available.

### DIFF
--- a/lib/CL/clCreateKernel.c
+++ b/lib/CL/clCreateKernel.c
@@ -83,10 +83,10 @@ POname(clCreateKernel)(cl_program program,
 
   for (device_i = 0; device_i < program->num_devices; ++device_i)
     {
+#ifdef OCS_AVAILABLE
       if (program->binaries[device_i] &&
           pocl_cache_device_cachedir_exists(program, device_i))
         {
-#ifdef OCS_AVAILABLE
           pocl_llvm_get_kernel_metadata (program, kernel, device_i,
                                          kernel_name, &errcode);
           cl_device_id device = program->devices[device_i];
@@ -118,12 +118,14 @@ POname(clCreateKernel)(cl_program program,
 
               device->ops->compile_kernel (&cmd, kernel, device);
             }
-#endif
         }
       /* If the program was created with a pocl binary, we won't be able to
          get the metadata for the cl_kernel from an IR file, so we call pocl
          binary function to initialize the cl_kernel data */
       else if (program->pocl_binaries[device_i])
+#else
+      if (program->pocl_binaries[device_i])
+#endif
         {
           errcode
             = pocl_binary_get_kernel_metadata (program->pocl_binaries[device_i],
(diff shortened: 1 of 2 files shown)


pjaaskel:
Please replace C++ style comments with C style /* */.  Overall, can we always assume there's a program.bc in the binary and that the files are stored in that exact order? It seems a bit fragile to me. E.g. if it was built from a SPIRV (when it's supported), there might not be a program.bc.

There was a plan to store also temp files (e.g. .cl) to the binaries (for transferring "debug data"), that is, when the "leave compiler temp files" env was set it would collect also those files to the binary from the cache dir. I cannot remember (and am too busy to check right now) if this is the case now or does it pick only certain files such as the program.bc to the container?

rjodin:
I agree, assuming there's a program.bc in that exact order is a bit fragile. We would need to have something more structured to avoid issue like that.
At the moment, "program.bc" is always serialize. I don't what's happening if it is not present in the directory, I suppose that the serialization fails...

From what I see, temp files are not present in the final binary.

franz:
@pjaaskel WRT 
> if it was built from a SPIRV (when it's supported), there might not be a program.bc.

.. there would be, because we need to translate SPIR to normal LLVM bitcode before linking it to kernel library (and program.bc is written **after** the kernel lib linking). At least that's how SPIR would work if it was implemented properly :)

WRT 
> Overall, can we always assume there's a program.bc in the binary

.. i think yes, there is always a program.bc, i don't remember pocl code ever to delete it (parallel.bc and <kernel>.so.o were deleted with LEAVE_TEMP_FILES unset).

If there's some good reason for having pocl binaries without program.bc, we *could* do it, but it will need a bit more work; a binary without program.bc would crash/fail on compiler conformance test, so there's some code that needs to be ifdef-ed out.

I'm not sure what order are we talking about, there's always only one program.bc per pocl binary, and always stored right after header, IIRC... so even a program with no kernels should be properly (de)serialized. In fact i'm about 90% sure this is tested by compiler conformance test.

WRT temp files, IIRC the code recursively serializes the kernel directories (but not the program directory). So everything that's in kernel dirs get serialized, but program.bc is serialized by an extra call b/c it's in program dir (outside kernel dirs).

franz:
Looking at the PR from 10000 feet, it looks OK. Also thanks for the fix :)

I'll run the PR through the compiler test and if it passes i'm good.

pjaaskel:
> If there's some good reason for having pocl binaries without program.bc, we could do it, but it will need a bit more work; a binary without program.bc would crash/fail on compiler conformance test, so there's some code that needs to be ifdef-ed out.

Obfuscation. Space saving. I'm not saying they are important needs now (at least for us here), but I think we should not add code that blindly trusts the file is there.

franz:
Well, the code that assumes it exists is the shortest possible _conformant_ code, and that's what we have now. It's the special cases like possibly-not-having program.bc that need adding extra code.

pjaaskel:
That's true.

pjaaskel:
@franz did it pass, OK to merge?

franz:
`PASSED 55 of 55 tests.`

LGTM.

PR #539, opened 2017-10-18
use SLEEF for (a part of) kernel library

franz:
implement some kernel library via SLEEF, plus a few fixes in generic kernel lib / vecmathlib.


jrprice:
Pastebins for AVX-512 compilation failure below.

CMake: https://pastebin.com/shBFVL7p
Make: https://pastebin.com/uncZFGyq

    /nfs/home/james/pocl-avx512/lib/kernel/sleef/arch/helperavx512f.h:200:3: error: expected expression
      return _mm512_cmp_epi64_mask(_mm512_cvtepi32_epi64(x), _mm512_cvtepi32_epi64(y), _MM_CMPINT_EQ);
      ^

franz:
Doh, that was a dumb typo. Retry please :)

jrprice:
Builds now, and passes all the core tests apart from `kernel/test_convert_type_16` (which doesn't fail with master). Here's the LLVM errors: https://pastebin.com/z4G5Ph63

franz:
Rebased & updated with upstream patches. @pjaaskel the AVX512 issue is fixed by the use-LLVM-inline commit, but currently i have it in the PR for upcoming release. Do you want it in this PR ?

franz:
Rebased, fixed build on ARM and i386, added a missing function (lgamma_r).

Created a bug for tracking the AVX512 issue #555 (which is sadly not fixed by use-LLVM-inline as i thought).

lib/kernel/sleef/arch/helperneon32.h
```diff
@@ -236,3 +236,9 @@ static INLINE void vsscatter2_v_p_i_i_vf(float *ptr, int offset, int step, vfloa
   vst1_f32((float *)(ptr+(offset + step * 0)*2), vget_low_f32(v));
   vst1_f32((float *)(ptr+(offset + step * 1)*2), vget_high_f32(v));
 }
+
+// TODO
```
pjaaskel: Should these be here?

lib/kernel/sleef/arch/helpervecext.h
```diff
@@ -853,3 +853,16 @@ static INLINE void vscatter2_v_p_i_i_vq(Sleef_quad *ptr, int offset, int step, v
 
 static INLINE void vsscatter2_v_p_i_i_vq(Sleef_quad *ptr, int offset, int step, vquad v) { vscatter2_v_p_i_i_vq(ptr, offset, step, v); }
 #endif
+
+
+// TODO
```
pjaaskel: Is it nice to have such dummy functions committed?

cmake/bitcode_rules.cmake
```diff
@@ -127,7 +127,7 @@ function(compile_sleef_c_to_bc EXT FILENAME SUBDIR BCLIST)
         "-I" "${CMAKE_SOURCE_DIR}/lib/kernel/sleef/arch"
         "-I" "${CMAKE_SOURCE_DIR}/lib/kernel/sleef/libm"
         "-I" "${CMAKE_SOURCE_DIR}/lib/kernel/sleef/include"
-        "-xc" "-o" "${BC_FILE}" "-c" "${FULL_F_PATH}"
+        "-O1" "-xc" "-o" "${BC_FILE}" "-c" "${FULL_F_PATH}"
```
pjaaskel: IMO this is a Clang bug or at least questionable behavior. Please check if this fixed upstream. It's common to run opts for the .bc produced by Clang with the 'opt' tool, it's weird that Clang would restrict the inlining that can be applied with opt. I think it hasn't been like this in the older versions of Clang.

lib/kernel/sleef/arch/helpervecext.h
franz: well, SLEEF is done in a way where you have the main files with algorithms using helpers (e.g. ``vadd_vf_vf_vf``, and based on the CPU type they include a "helper header" which translates helpers to actual intrinsics (``vadd_vf_vf_vf`` -> ``_mm256_add_ps`` on AVX2, etc). So i must include the a function in all helper headers for the main file to compile everywhere, even if it's just a dummy function on ARM currently.

cmake/bitcode_rules.cmake
franz: Looking at LLVM trunk, it's still there:

``./tools/clang/lib/Frontend/CompilerInvocation.cpp``  around line 454:

```
  // At O0 we want to fully disable inlining outside of cases marked with
  // 'alwaysinline' that are required for correctness.
  Opts.setInlining((Opts.OptimizationLevel == 0)
                       ? CodeGenOptions::OnlyAlwaysInlining
                       : CodeGenOptions::NormalInlining);
```

Git blame:

```
commit d0ca1639669b4ad12088c1b4d84312b7091f1666
Author: Chandler Carruth <chandlerc@gmail.com>
Date:   Fri Dec 23 01:24:49 2016 +0000

    Cleanup the handling of noinline function attributes, -fno-inline,
    -fno-inline-functions, -O0, and optnone.
```

So you're right that this hasn't been the case with older versions of Clang. We could probably workaround this somewhere in pocl-llvm-API but the -O1 seems to be the simplest solution.

PR #547, opened 2017-10-30
pocl_binary: add reqd_wg_size informations into pocl_binary_kernel structure

rjodin:


pjaaskel:
This LGTM, but doesn't pass the test suite for me. These fail:

```
46:regression/test_program_from_binary_with_local_1_1_1_REPL
60:regression/test_program_from_binary_with_local_1_1_1_LOOPS
71:runtime/clCreateProgramWithBinary
85:runtime/test_enqueue_kernel_from_binary
```

rjodin:
My mistake.
I tried the feature with some printf, and I broke my code while removing them.
It should be functional now.
(This issue was with the macros and the "for" loops, missing paranthesis)

PR #557, opened 2017-11-07
User events handling

franz:
Already described at pocl-devel, so only shortly:

This PR attempts to fix the remaining issues with user events.
 
In particular, pocl_broadcast() now calls notify on events also in
CL_QUEUED state, and also calls notify on events which depend on failed
*user* events. Some problems remain:
 
1) i've updated all drivers except CUDA which seems to have its own way
of handling events.
 
2) after any event fails, the state is "implementation defined". Code in this PR just sets all dependent events to failed.

### DIFF
--- a/lib/CL/clCreateUserEvent.c
+++ b/lib/CL/clCreateUserEvent.c
@@ -20,6 +20,11 @@ POname(clCreateUserEvent)(cl_context     context ,
       event->pocl_refcount = 1;
       event->status = CL_SUBMITTED;
       event->context = context;
+      pocl_user_event_data *p = malloc (sizeof (pocl_user_event_data));
+      assert (p);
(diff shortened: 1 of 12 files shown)


franz:
@jrprice any ideas what needs to be done for the CUDA driver ?

jrprice:
There's just one tweak needed to get the `events` conformance tests passing with CUDA again:
https://github.com/jrprice/pocl/commit/f7b3a16e51298a3f3437d96f4c559724e3a03519

Longer term, I should probably rework the handling of events to be more like the other drivers for dependencies that come from user events and commands submitted to other drivers, but I don't have time to look into this right now.

franz:
@jrprice Thanks, i've added your patch. Perhaps the entire event handling could use a little cleanup, but definitely not for this release.

Issue #516, opened 2017-08-22
WI-loops don't consider alignment for context arrays [was: vload3 is broken]


franz:
SIGSEGV on VMOVAPS with unaligned address.

Oblomov:
Yeah, `vload*` should be implemented as element-by-element fetch since the whole point of them is to work without any alignment constraint (e.g. packed `float3` structures).

franz:
@Oblomov: actually it has been using element-by-element for a while now, that's not the issue.

The problem seems to be not with the vload3 itself, but the vload3 test case. There is a piece of code with a small loop:

```
    __private char sPrivateStorage[ 255 ];
    int tid = get_global_id( 0 );

    for( int i = 0; i < 255; i++ )
    {
        sPrivateStorage[ i ] = src[ i ];
    }
```

and LLVM seems to optimize it in some strange way, to copy the first 128 bytes via AVX and copy the remaining 127bytes one-by-one (AFAICT). For some reason the AVX instructions add the size of sPrivateStorage (255) to the stack pointer (rsp). So `vmovaps XMMWORD PTR [rsp+r11*1+0x20],xmm2` should be `vmovaps XMMWORD PTR [rsp+0x20],xmm2` AFAICT.

git bisect says it's broken by commit 69ea56f3c0bbf01ecee7517dbd0f ( kernel compiler: Apply standard optimations early). @pjaaskel can you take a look ?

pjaaskel:
If there is a WI, loop, it should increment the sp with the WI ID times the private array size to find its copy. Are you sure it is not that? Do you see a problem in the bitcode, which should be very simple in this case?

franz:
> it should increment the sp with the WI ID times the private array size to find its copy. 

And that is the problem. The private array is defined as:

`  %sPrivateStorage = alloca [255 x i8], align 16`

The standard-opts pass then vectorizes the small for loop and uses 16-aligned stores to this array. The WI pass then uses 255 * WI-ID as index, which obviously breaks the 16-aligned store instructions. So WI pass needs to consider both alignment and size, not just size.

pjaaskel:
Yes, that's a clear bug if it doesn't.

pjaaskel:
And it's a good sign -- autovec finally starting to vectorize pocl's output more aggressively.

PR #560, opened 2017-11-13
various small fixes

franz:
* docker files
* sanitizer builds are usable again
* fix memleaks found by leak sanitizer
* fix a few issues


lib/llvmopencl/WorkitemLoops.cc
```diff
@@ -965,7 +965,20 @@ WorkitemLoops::GetContextArray(llvm::Instruction *instruction)
             allocType =
               ArrayType::get(elementType->getArrayElementType(), alignedSize);
           } else if (isa<StructType>(elementType)) {
-          // TODO handle structs properly
+            StructType *old_struct = dyn_cast<StructType>(elementType);
```
pjaaskel: Can you remind me why is this code needed and what it does exactly? Doesn't ArrayType::get() alignment argument work for structs for some reason? I think we discussed this but I forgot what was the exact problem. It should be added to comments here. You create a new type with a padding array and the old struct's elements, but what if the struct has child structs recursively etc.?

doc/sphinx/source/env_variables.rst
```diff
@@ -5,6 +5,10 @@ The behavior of pocl can be controlled with multiple environment variables
 listed below. The variables are helpful both when using and when developing
 pocl.
 
+- **POCL_AFFINITY**
+ If set to 1, each thread of the pthread CPU driver sets its affinity to
```
pjaaskel: What are the benefits of this setting? Can you add some more documentation so people know when and why this should be used.

lib/CL/clCreateKernel.c
```diff
@@ -178,6 +178,7 @@ POname(clCreateKernel)(cl_program program,
       POCL_MEM_FREE(kernel->reqd_wg_size);
       POCL_MEM_FREE(kernel->dyn_arguments);
       POCL_MEM_FREE(kernel->arg_info);
+      POCL_MEM_FREE (kernel->name);
```
pjaaskel: Can you fix the argument spacing style for the others around it too?

doc/sphinx/source/env_variables.rst
franz: performance-wise probably none, but it allows to easily verify that splitting pthread into subdevices works properly (and it's just a few lines).

lib/CL/clCreateKernel.c
franz: ok

Docker/ArchLinux/default
```diff
@@ -0,0 +1,9 @@
+FROM pritunl/archlinux:latest
```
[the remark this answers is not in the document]
franz: No, docker doesn't care. I'll move it to tools

lib/llvmopencl/WorkitemLoops.cc
franz: > Can you remind me why is this code needed and what it does exactly? 

It re-aligns array (and now struct) types in WILoop's Context array. It's explained in a comment:

```
      /* 3D context array. In case the elementType itself is an array or struct,
       * we must take into account it could be alloca-ed with alignment and loads
       * or stores might use vectorized instructions expecting proper alignment.
       * Because of that, we cannot simply allocate x*y*z*(size), we must
       * enlarge the array type to fit the alignment. */
```
We've been fixing this in issue #516 already, but i only fixed arrays.

Perhaps the proper fix would be to tell LLVM to emit unaligned vector instructions, but i have no idea how to do that.

> ArrayType::get() alignment argument work for structs

There's no alignment argument in ArrayType::get(); ``alignedSize`` is the array size argument. This line of code creates a new ArrayType (with more elements).

> but what if the struct has child structs recursively etc.?

Then it may or may not break, depending on how LLVM calculates the sizes.

Right now even simple structs break on this, so i think it's an improvement. I agree padding is not the best way to fix this, but i haven't found a simple way to tell LLVM to create structs with alignments.

pjaaskel: OK, I see. IIRC in LLVM the alignment info is attached to the loads and stores accessing the data, not to the variable, so there might not be another way.

PR #564, opened 2017-11-17
* make SLEEF xfmaxf(), xfminf(), xfmax() and xfmin() aware of platfor…

blu:
…ms with minmax-number-only capabilities via macro SLEEF_*_MINMAXNUM_AVAILABLE

* add minmax-number-only capabilities to SLEEF arch AArch64


franz:
LGTM. Tested on A53

PR #571, opened 2017-11-23
Various stuff

franz:
* pthread driver now respects subdevices 
* conformance_suite_micro - yet smaller version of CTS, runs ~20 minutes on slow hardware
* some small fixes


lib/CL/devices/basic/basic.c
```diff
@@ -244,6 +244,8 @@ pocl_basic_init_device_infos(unsigned j, struct _cl_device_id* dev)
   dev->platform = 0;
 
   dev->parent_device = NULL;
+  dev->core_count = 0;
+  dev->core_start = 0;
   // basic does not support partitioning
```
pjaaskel: Please convert this comment to /* */

lib/CL/devices/basic/basic.c
```diff
@@ -244,6 +244,8 @@ pocl_basic_init_device_infos(unsigned j, struct _cl_device_id* dev)
   dev->platform = 0;
 
   dev->parent_device = NULL;
+  dev->core_count = 0;
```
pjaaskel: Why 0? Basic should be 1?

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -174,13 +175,35 @@ work_group_scheduler (kernel_run_command *k,
 static void finalize_kernel_command (thread_data *thread_data,
                               kernel_run_command *k);
 
-int pthread_scheduler_get_work (thread_data *td, _cl_command_node **cmd_ptr)
+static int
+shall_we_run_this (thread_data *td, cl_device_id subd, void *cmd)
```
pjaaskel: This function requires a basic comment of what is its purpose.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -174,13 +175,35 @@ work_group_scheduler (kernel_run_command *k,
...
+{
+  if (subd && subd->parent_device)
+    {
+      // subdevice, let's see if we're supposed to pick up work
```
pjaaskel: Please convert to /* */.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -174,13 +175,35 @@ work_group_scheduler (kernel_run_command *k,
...
 {
   _cl_command_node *cmd;
   kernel_run_command *run_cmd;
+
   // execute kernel if available
   pthread_spin_lock (&scheduler.wq_lock_fast);
-  if ((run_cmd = scheduler.kernel_queue))
+  run_cmd = scheduler.kernel_queue;
+
+  // execute kernel if available
```
pjaaskel: /* */

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -422,6 +449,7 @@ pocl_pthread_driver_thread (void *p)
    * force a first FTZ setup */
   td->current_ftz = 213;
   td->num_threads = scheduler.num_threads;
+  td->last_cmd_ignored = NULL;
 
 #ifdef __linux__
   if (pocl_get_bool_option ("POCL_AFFINITY", 0))
```
pjaaskel: How does the subdevice splitting work if you don't pin the threads to cores to ensure work gets scheduled only to the cores in the sub device? Kernel's scheduler can migrate the threads to any core AFAIK and you don't have control for the subdevice compute units then? This is fuzzy.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -197,26 +220,28 @@ int pthread_scheduler_get_work (thread_data *td, _cl_command_node **cmd_ptr)
...
 static void
-pthread_scheduler_sleep()
+pthread_scheduler_sleep (thread_data *td)
 {
   struct timespec time_to_wait = {0, 0};
   time_to_wait.tv_sec = time(NULL) + 5;
 
   pthread_spin_lock (&scheduler.wq_lock_fast);
-  if (scheduler.work_queue == NULL && scheduler.kernel_queue == 0)
+  if ((scheduler.work_queue == NULL && scheduler.kernel_queue == NULL)
```
pjaaskel: This needs comments.

examples/EinsteinToolkit/EinsteinToolkit.c
```diff
@@ -526,8 +524,22 @@ void setup(const char* program_source1, const char* program_source2)
...
+  main_device_id = device_ids[0];
+
+  if (use_subdev)
+    {
+      const cl_device_partition_property props[]
+          = { CL_DEVICE_PARTITION_EQUALLY, 2, 0 };
+      cl_device_id subdevs[128];
+      cl_uint retval;
+      int err
+          = clCreateSubDevices (main_device_id, props, 2, subdevs, &retval);
```
pjaaskel: How do you verify that the sub devices is working correctly, that is, it's restricting its execution to only certain CUs? I think that's not trivial to do automatically.

lib/CL/devices/basic/basic.c
franz: It's only valid for subdevices.

examples/EinsteinToolkit/EinsteinToolkit.c
franz: Actually it is. Currently it's not done automatically, but measuring execution time for 1 CU vs 2 CUs should do the trick.

lib/CL/devices/pthread/pthread_scheduler.c
franz: > How does the subdevice splitting work if you don't pin the threads to cores

Affinity has nothing to do with how it works, except it makes it easier to manually check that it works.

> Kernel's scheduler can migrate the threads to any core AFAIK

Of course, and it doesn't matter. Each pthread gets an index at creation; each subdevice gets its (core_start, core_count) values set up in clCreateSubDevices; then ``shall_we_run_this`` decides whether a particular pthread runs a command, based on its index fitting within command->device->[start, start+count]; if it does not, it ignores the command.

lib/CL/devices/basic/basic.c
pjaaskel: Please add a comment about it here.

lib/CL/devices/pthread/pthread_scheduler.c
pjaaskel: OK. So it restricts the threads the commands can go into, thus setting an upper limit to the number of cores it occupies. But what stops the threads getting scheduled to the same cores? E.g. if one wants to partition an N core machine so half of the cores are dedicated to another part and another to other part, how does it work if you only restrict the threads the tasks are pushed to, but not where the threads are scheduled?

examples/EinsteinToolkit/EinsteinToolkit.c
franz: yes, that's true

lib/CL/devices/pthread/pthread_scheduler.c
franz: > if you only restrict the threads the tasks are pushed to, but not where the threads are scheduled?

The tasks are not pushed to; they are pulled by each thread. If you're asking what stops threads getting scheduled to random CPU cores - nothing, except when you set POCL_AFFINITY=1 env var. But the code does not depend on that env var to work correctly, at worst you'll see worse performance b/c the threads will be randomly scheduled across CPU.

pjaaskel: OK, got it. Like discussed, let's leave improving this for post 1.0 (or other contributors).

Issue #525, opened 2017-09-03
LLVM/Clang 5.0 POCL CUDA errors in ptx-gen..


oscarbg:
Hi,
just compiled Pocl with CUDA support using LLVM/Clang 5.0rc3 from (http://prereleases.llvm.org/5.0.0/rc3/clang+llvm-5.0.0-rc3-x86_64-linux-gnu-debian8.tar.xz)
and got errors on simple example2 which before (LLVM4.0) was working.. 

seems error is:
```
module flag identifiers must be unique (or of 'require' type)
!"wchar_size"
[CUDA] ptx-gen: module verification failed
```

full log:
 ./example2
```
[2017-08-03 08:59:41.362466978]POCL: in fn cl_int pocl_cuda_init(unsigned int, cl_device_id, const char *) at line 261:
  |   GENERAL |  [CUDA] GPU architecture = sm_52
[2017-08-03 08:59:41.140342764337684]POCL: in fn cl_command_queue POclCreateCommandQueue(cl_context, cl_device_id, cl_command_queue_properties, cl_int *) at line 42:
  |   GENERAL |  Create Command queue on device 0
[2017-08-03 08:59:41.140342765167403]POCL: in fn cl_int compile_and_link_program(int, int, cl_program, cl_uint, const cl_device_id *, const char *, cl_uint, const cl_program *, const char **, cl_uint, const cl_program *, void (*)(cl_program, void *), void *) at line 485:
  |   GENERAL |  building program with options (null)
[2017-08-03 08:59:41.140342765174239]POCL: in fn cl_int compile_and_link_program(int, int, cl_program, cl_uint, const cl_device_id *, const char *, cl_uint, const cl_program *, const char **, cl_uint, const cl_program *, void (*)(cl_program, void *), void *) at line 531:
  |   GENERAL |  building from sources for device 0
[2017-08-03 08:59:41.671759072]POCL: in fn void linkLibDevice(llvm::Module *, const char *, const char *) at line 587:
  |   GENERAL |  loading libdevice from '/usr/local/cuda/nvvm/libdevice/libdevice.compute_50.10.bc'
[2017-08-03 08:59:41.140343064631154]POCL: in fn int pocl_ptx_gen(const char *, const char *, const char *, const char *, const char *, int) at line 99:
  |   GENERAL |  NVVM module:
[...]

PR #572, opened 2017-11-24
[CUDA] Add POCL_CUDA_VERIFY_MODULE environment variable

jrprice:
Make module verifiation opt-in, since the pocl linker currently
generates invalid metadata.

This is a workaround for #525, to make the CUDA backend work with LLVM 5+.

### DIFF
--- a/lib/CL/devices/cuda/pocl-ptx-gen.cc
+++ b/lib/CL/devices/cuda/pocl-ptx-gen.cc
@@ -99,12 +99,15 @@ int pocl_ptx_gen(const char *BitcodeFilename, const char *PTXFilename,
     POCL_MSG_PRINT_INFO("NVVM module:\n%s\n", ModuleString.c_str());
   }
 
-  // Verify module.
   std::string Error;
-  llvm::raw_string_ostream Errs(Error);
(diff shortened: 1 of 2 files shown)


lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -99,12 +99,15 @@ int pocl_ptx_gen(const char *BitcodeFilename, const char *PTXFilename,
...
 
-  // Verify module.
   std::string Error;
-  llvm::raw_string_ostream Errs(Error);
-  if (llvm::verifyModule(*Module->get(), &Errs)) {
-    POCL_MSG_ERR("\n%s\n", Error.c_str());
-    POCL_ABORT("[CUDA] ptx-gen: module verification failed\n");
+
+  // Verify module.
+  if (pocl_get_bool_option("POCL_CUDA_VERIFY_MODULE", 0)) {
```
pjaaskel: Please add this to the documents too.

jrprice: Good point, added this and snuck in one extra doc update too.

Issue #554, opened 2017-11-02
clCreateSubDevices with type CL_DEVICE_PARTITION_BY_COUNTS doesn't respect supplied core count


arghdos:
Migrated from mailing list [see here](https://sourceforge.net/p/pocl/mailman/message/36100200/) for details and example.

franz:
@arghdos PR #571 has the fix (there's also a test for it)

arghdos:
@franz thanks!  Giving this a test shows an increasing runtime (for a problem size of 10^7 in the example from the mailing list), with increasing cores:

>platform: pocl  num cores: 1    elapsed time: 2.722041e+03
platform: pocl  num cores: 2    elapsed time: 2.964494e+03
platform: pocl  num cores: 4    elapsed time: 3.134813e+03
platform: pocl  num cores: 8    elapsed time: 5.078341e+03
platform: pocl  num cores: 16   elapsed time: 4.990547e+03
platform: pocl  num cores: 32   elapsed time: 7.407428e+03
platform: Intel num cores: 1    elapsed time: 2.003667e+03
platform: Intel num cores: 2    elapsed time: 2.509890e+03
platform: Intel num cores: 4    elapsed time: 1.000509e+03
platform: Intel num cores: 8    elapsed time: 8.530500e+02
platform: Intel num cores: 16   elapsed time: 7.983490e+02
platform: Intel num cores: 32   elapsed time: 7.267400e+02

so there is likely more optimization to do in the scheduler, but the CPU usage shown by `top` does correspond to the expected number of cores.

Perhaps we should close this one and open another issue for scheduler performance?  I'd also be willing to take a look at that if you'd like

arghdos:
Also of potentially of note, the SubDev test is failing with a segfault on this system:

>Command: "/home/ncurtis/pocl/build/examples/EinsteinToolkit/EinsteinToolkit" "s"
Directory: /home/ncurtis/pocl/build/examples/EinsteinToolkit
"EinsteinToolkit_SubDev" start time: Nov 24 12:48 EST
Output:
----------------------------------------------------------
<end of output>
Test time =   0.89 sec
----------------------------------------------------------
Test Failed.
"EinsteinToolkit_SubDev" end time: Nov 24 12:48 EST
"EinsteinToolkit_SubDev" time elapsed: 00:00:00

pjaaskel:
> I'd also be willing to take a look at that if you'd like

[...]

PR #576, opened 2017-11-24
Fix num devices check

arghdos:
Migrated from #554 -- note, the segfault went away after a clean rebuild.  Instead I just had a failing test.

---

The relevant part of the spec is:

>CL_INVALID_VALUE if out_devices is not NULL and **num_devices is less than the number of sub-devices** created by the partition scheme.

instead, it was checking for exact equality (`clCreateSubDevices.c::126`).

On my 40 core server, this manifested as an error in the `EinsteinToolkit_subDev` test, as it would split `40 / 2 -> 20 != 2` passed in `clCreateSubDevices (main_device_id, props, 2, subdevs, &retval);`

Instead that should be the size of the full sub device array (128).

### DIFF
--- a/examples/EinsteinToolkit/EinsteinToolkit.c
+++ b/examples/EinsteinToolkit/EinsteinToolkit.c
@@ -533,7 +533,7 @@ void setup(const char* program_source1, const char* program_source2)
       cl_device_id subdevs[128];
       cl_uint retval;
       int err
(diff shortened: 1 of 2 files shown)


pjaaskel:
Looks OK. What's your real name and email which I can add to the CREDITS?

arghdos:
@pjaaskel Nick Curtis -- nicholas.curtis@uconn.edu

PR #606, opened 2018-01-18
pre 1.1 fixes

franz:
few fixes for LLVM 6.0 + random stuff
... with recent LLVM 6 pull, all tests now pass.

### DIFF
--- a/include/pocl_cache.h
+++ b/include/pocl_cache.h
@@ -51,7 +51,7 @@ void pocl_cache_cleanup_cachedir(cl_program program);
 int pocl_cl_device_to_index(cl_program   program,
                             cl_device_id device);
 
-void pocl_cache_tempname (char *path_template, const char *suffix, int *fd);
+int pocl_cache_tempname (char *path_template, const char *suffix, int *fd);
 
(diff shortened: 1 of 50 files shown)


include/pocl_cache.h
```diff
@@ -48,16 +48,6 @@ pocl_cache_create_program_cachedir(cl_program program, unsigned device_i,
```
pjaaskel: Based on your commit description: What happens if thread A manages to build the kernel and start running it while thread B is simultaneously building the same kernel and then renames it, overwriting the .so loaded and being executed by A?

pjaaskel:
474 - conformance_printf (Failed) with LLVM 5.0 after pulling this in.

include/pocl_cache.h
franz: Overwriting the .so should be safe, but i just realized we might be loading the same (or two different b/c of rewrite) kernel.so files with dlopen(). I'll add a commit to fix up dlcache handling a bit.

franz:
OK, rebased. Fixed printf by dropping the offending commit, added a comit to fix dlhandle_cache.

PR #618, opened 2018-02-16
probably last feature PR before 1.1

franz:
small fixes + SPIR support


oscarbg:
Hi,I see you say some very exp. SPIR-V support but this depends on functional llvm-spirv converter. The Khronos official llvm-spirv is currently unusable for pocl..
can detail how to obtain or build a llvm-spirv converter suitable for pocl then?

doc/sphinx/source/env_variables.rst
```diff
@@ -161,6 +161,11 @@ pocl.
  Defaults to 1. If set to 0, pocl will not install the SIGFPE handler.
  See :ref:`sigfpe-handler`
 
+- **POCL_SYNC_CACHE**
```
pjaaskel: Is this really necessary? I mean 1) is the fdatasync() needed in some scenarios? 2) is there a reason not to call it by default? The cold cache case is slow anyways and it should affect that only. I'd like to reduce the number of envs we have instead of adding them unless really necessary.

doc/sphinx/source/features.rst
```diff
@@ -27,3 +33,23 @@ All 1.2 API call should be implemented. The list of unimplemented
...
 https://github.com/pocl/pocl/blob/master/lib/CL/clGetPlatformIDs.c
 
+SPIR support
+------------
+
+There is some experimental support available for SPIR, with LLVM 5 and newer.
+There is some even more experimental support available for SPIR-V, but this
+depends on functional llvm-spirv converter. The Khronos official llvm-spirv
+is currently unusable for pocl (because it produces LLVM 3.6 bitcode and pocl
+requires LLVM 5+ for SPIR, and LLVM 5 refuses to load LLVM 3.6 bitcode).
```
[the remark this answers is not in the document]
franz: oh, i had no idea. That tool could be very useful for this case.

doc/sphinx/source/env_variables.rst
franz: Well, it's not that necessary, but depending on the filesystem, a write+close+rename might result in an empty file in case of problems, and this lowers the chances of that. Good point about env vars though. I'll take another look at it.

franz:
@oscarbg the only one i found is this one: https://gitlab.collabora.com/tomeu/llvm-spirv.git (ignore the README there, it's outdated) but even that has some problems with llvm bitcode <-> spir-v conversion. Requires LLVM 5+ to build.

oscarbg:
@franz thanks good to know..

franz:
Updated PR, fixed the cache sync (it's now without env var) and added a few more workarounds:

* ``conformance_micro_compiler`` test was failing because ``cl_khr_spir`` is not in approved extension list (disabled the test for now)
* TCE + LLVM 6 one test fails (disabled for now)
* disabled metadata-less SPIR test (it seems Clang now always generates metadata)

lib/llvmopencl/LLVMFileUtils.cc
```diff
@@ -298,6 +300,9 @@ int pocl_write_file(const char *path, const char *content, uint64_t count,
     if (write(fd, content, (ssize_t)count) < (ssize_t)count)
       return errno ? -errno : -1;
 
+    if (fdatasync(fd))
```
pjaaskel: My man page says:
```
      On POSIX systems on which fdatasync() is available, _POSIX_SYNCHRONIZED_IO is defined in <unistd.h> to a value greater  than  0.   (See
       also sysconf(3).)
```
Can you use that macro to add a bit more portability?

franz: Good point, will do for next PR.

oscarbg:
@franz seems https://github.com/pierremoreau/llvm-spirv/ "branch integrate_with_mesa" is a newer branch with than  https://gitlab.collabora.com/tomeu/llvm-spirv.git 
seen here:
"Introducing SPIR-V support to clover":
https://lists.freedesktop.org/archives/mesa-dev/2018-February/186333.html
"llvm-spirv can be currently grabbed from
https://github.com/pierremoreau/llvm-spirv, on the branch integrate_with_mesa."
maybe more stable than other branch now than CLOVER devs are using it to support SPIR-V..
could mention in release notes of final 1.1 if you give a try under some **EXPERIMENTAL** warning..

franz:
@oscarbg thanks, i've also found it. Yes I could mention in release notes, but i need at least one SPIR-V bitcode that the tool can convert (currently it fails even with very simple ones).

oscarbg:
Hi @franz,
Have you noticed new branch https://github.com/KhronosGroup/SPIRV-LLVM/tree/khronos/spirv-6.0 so seems “official” and seems worked by Intel devs who are pushing hard for OpenCL 2.2 support this year on Intel GPUs.. can you give a shot for POCL Spir-v support?
Thanks..

franz:
Hello @oscarbg 

Thanks, i didn't know about that branch. I have built it, a quick test shows that its ``llvm-spirv`` seems to be working. I will update the documentation to mention experimental SPIR-V support.

PR #619, opened 2018-02-21
last feature PR before 1.1

franz:
kernel compilation speedup + various fixes


lib/CL/devices/pthread/pthread.c
```diff
@@ -476,7 +476,7 @@ void pocl_pthread_update_event (cl_device_id device, cl_event event, cl_int stat
       if (event->queue->properties & CL_QUEUE_PROFILING_ENABLE)
         event->time_end = device->ops->get_timer_value(device->data);
 
-
+      POCL_DEBUG_EVENT_TIME (event, "NDRange Kernel        ");
```
pjaaskel: Why the white space? Looks weird here.

franz: string-justify... all timing info messages have the same length, originally intended for nicer output when run with POCL_DEBUG=timing

jrprice:
FYI this PR crashes with a multi-device CG benchmark that I use to test the async stuff in the CUDA backend - will try and investigate.

franz:
@jrprice FWIW the only changes directly related to CUDA are the module verification (i enabled it by default) or the event-wait. Wildly guessing it's the module verification

CMakeLists.txt
```diff
@@ -387,18 +387,63 @@ endmacro()
 
 if(UNIX)
   include(CheckCSourceCompiles)
+
+  # don't allow implicit function declarations
+  set(CMAKE_REQUIRED_FLAGS "-std=c99")
```
jrprice: This causes the `HAVE_CLOCK_GETTIME` check to fail on my system (with various versions of GCC).

Using `-std=gnu99` instead or adding `-D_POSIX_C_SOURCE=199309L` fixes it.

franz: There should be a ``#define _DEFAULT_SOURCE`` in the HAVE_CLOCK_GETTIME test with commit ff42b03 - according to gc manual for _DEFAULT_SOURCE, "the effect is to enable features from the 2008 edition of POSIX". Weird thing is, i can't see that commit (and a bunch of others) in this PR.

jrprice: `_DEFAULT_SOURCE` is there, but it was only introduced in glibc 2.19, and my glibc is a couple of revs older than that (using RHEL 7.3).

franz: @jrprice OK thanks, i forgot that _DEFAULT_SOURCE is quite recent.

lib/CL/devices/pthread/pthread.c
pjaaskel: Isn't this better be done at the callee side of this macro? No biggie, just looks a bit funny.

franz:
dropped the patch causing problems for CUDA, plus slightly tuned the portability check macros

franz:
After spending some time with ThreadSanitizer + CUDA, i've decided to postpone all CUDA patches (except module verification) until after pocl 1.1 release. It'll require more work than makes sense to do before release.

CMakeLists.txt
Oblomov: Setting `_POSIX_C_SOURCE` to 200809L should work too.

PR #627, opened 2018-03-05
support hwloc 2

jeffhammond:
This is an attempt to address issue #358.  It is likely not ready to be merged. 

- Source changes are simple substitution based upon https://www.open-mpi.org/projects/hwloc/doc/v2.0.0/a00327.php
- API version is detected from hwloc.h in-place, rather than using CMake.
- There is no test for header-library incompatibility.  This is demonstrated on https://www.open-mpi.org/projects/hwloc/doc/v2.0.0/a00326.php#faq_upgrade and can be implemented.

### DIFF
--- a/lib/CL/devices/topology/pocl_topology.c
+++ b/lib/CL/devices/topology/pocl_topology.c
@@ -101,6 +101,15 @@ pocl_topology_detect_device_info(cl_device_id device)
     return ret;
   }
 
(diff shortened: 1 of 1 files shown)


franz:
WRT "API version is detected from hwloc.h in-place, rather than using CMake." - i think that's fine.

I have created a new PR #629 which has your commit + the test for header-library incompatibility, let's continue there.

PR #637, opened 2018-04-11
Allow SPIR -> PTX conversion in llvm_build

PeterTh:
The scenario here is that we are using POCL to provide a OpenCL device that supports SPIR on Nvidia hardware that only natively supports PTX. 
This is useful for higher-level libraries/APIs building on top of OpenCL that emit SPIR.

Without this change, this use case fails with "Invalid SPIR binary triple". With it, we have successfully run several ComputeCPP tests on Nvidia hardware using the POCL CUDA backend.

### DIFF
--- a/lib/CL/pocl_llvm_build.cc
+++ b/lib/CL/pocl_llvm_build.cc
@@ -623,6 +623,8 @@ int pocl_llvm_link_program(cl_program program, unsigned device_i,
         POCL_MSG_ERR("Invalid SPIR binary triple\n");
         return CL_LINK_PROGRAM_FAILURE;
       }
+    } else if(triple.getArch() == Triple::nvptx || triple.getArch() == Triple::nvptx64) {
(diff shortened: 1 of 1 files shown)


pjaaskel:
Sorry for the slow response. Why don't you add nvptx to the same branch as x86 is? That is, force set the datalayout to silence LLVM error due to different datalayouts? @franz do you recall if that incompatible datalayout error was an assertion or just spurious error printout? If an assert then your nvptx special case likely works only if LLVM is compiled without assertions.

franz:
Since our group now needs SPIR working with TCE backend, i've started working on better SPIR support. Good news is that CUDA and the rest will get a free ride.

Turns out that one big missing piece of SPIR support is address space number translation. Simple CUDA SPIR examples work by accident because the global and private AS have the same numbers in SPIR bitcode. I plan to get all CL1.2 address spaces working with SPIR in my next PR.

@PeterTh hate to do this to first-time contributors, but i'll be closing this PR. Hopefully the good news makes up for it

PR #640, opened 2018-04-30
Device-side printf()

franz:
implements device-side printf.

I added an external project (MIT license) for %e / %f / %g conversions, but by default use libc via snprintf.


lib/kernel/tinyprintf.c
```diff
@@ -1,8 +1,9 @@
...
-File: tinyprintf.c
+/* OpenCL built-in library: tiny_printf
 
 Copyright (C) 2004  Kustaa Nyholm
 
+Copyright (C) 2018  Michal Babej
+
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU Lesser General Public
 License as published by the Free Software Foundation; either
```
pjaaskel: The LGPL might be a problem here. I think an interpretation is that it's OK only to link to a LGPL lib without infecting other codes with the license. Can we find a more liberal replacement?

lib/llvmopencl/Workgroup.cc
```diff
@@ -236,6 +258,12 @@ Workgroup::runOnModule(Module &M)
       }
   }
 
+    // remove functions which were replaced by their printf-augmented variants
```
pjaaskel: Please do not commit commented out code.

lib/kernel/tinyprintf.c
franz: Actually, it's dual-licensed. http://www.sparetimelabs.com/tinyprintf/tinyprintf.php

> Note, as of 8.5.2012 the license has been changed to BSD so you are free to use the LGPL licensed from this page or the BSD licensed from the link above.

pjaaskel: Ah, right. I think you already mentioned that. Then please attach the BSD license to the file and remove the LGPL.

PR #641, opened 2018-05-15
Improved SPIR support for TCE + CUDA

franz:
+ a bunch of small fixes. There are a few bugs left in the SPIR support but AFAIK no regressions.


lib/CL/devices/basic/basic.c
```diff
@@ -293,6 +293,10 @@ pocl_basic_init_device_infos(unsigned j, struct _cl_device_id* dev)
 
   dev->device_side_printf = 1;
   dev->printf_buffer_size = 16 * 1024 * 1024;
+  dev->spir_compatible_triple = SPIR_HOST_TARGET_TRIPLE;
+  dev->spir64_datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128";
```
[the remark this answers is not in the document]
franz: SPIR does have a fixed DL, but we need to forcibly change the module DL to device-specific DL before linking, otherwise you get LLVM assertion errors.

lib/CL/pocl_cl.h
```diff
@@ -504,6 +504,9 @@ struct _cl_device_id {
   size_t printf_buffer_size;
   char *short_name;
   char *long_name;
+  const char* spir_compatible_triple;
```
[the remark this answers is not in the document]
franz: OK

lib/CL/devices/tce/ttasim/ttasim.cc
```diff
@@ -147,7 +147,22 @@ pocl_ttasim_init_device_infos(unsigned j, struct _cl_device_id* dev)
   dev->local_mem_type = CL_GLOBAL;
   dev->error_correction_support = CL_FALSE;
   dev->host_unified_memory = CL_FALSE;
-  dev->endian_little = CL_FALSE;
+
+  // TODO this should be read automatically from ADF file
+  if (pocl_get_bool_option("POCL_TCELE", 0)) {
```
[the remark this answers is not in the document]
franz: Yeah i've noticed machine->littleEndian(), but AFAICT it is setup quite late, not in the constructor. I'll take another look at this

lib/CL/pocl_llvm_build.cc
```diff
@@ -588,6 +591,32 @@ int pocl_llvm_build_program(cl_program program,
   return CL_SUCCESS;
 }
 
+#if defined(ENABLE_SPIR) && defined(TCE_AVAILABLE)
```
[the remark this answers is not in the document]
franz: OK.

lib/CL/pocl_llvm_build.cc
```diff
@@ -615,16 +644,32 @@ int pocl_llvm_link_program(cl_program program, unsigned device_i,
...
+    Triple::ArchType t = triple.getArch();
+    const std::string mod_t(linked_module->getTargetTriple());
+    if (device->spir_compatible_triple &&
+        (t == Triple::x86 || t == Triple::x86_64 ||
+#ifdef TCE_AVAILABLE
+         t == Triple::tce || t == Triple::tcele ||
+#endif
+         t == Triple::nvptx || t == Triple::nvptx64)) {
+
+      if (StringRef(mod_t).equals(StringRef(device->spir_compatible_triple))) {
```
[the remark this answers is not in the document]
franz: > What's the SPIR compatible triple, what is its purpose?

To refuse SPIR binaries with different pointer size than the device's kernel library. With this code, on 64bit linux host CPU and NVidia only accept 64bit SPIR, while TCE accepts 32bit SPIR everywhere.

It *may* be possible to link 32bit SPIR to 64bit kernel library, but i haven't tried it and it seems  more trouble than it's worth.

lib/CL/devices/basic/basic.c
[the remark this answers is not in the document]
pjaaskel: OK, why it's called spir64 datalayout then? You are forcing it to the target-specific data layout, so it is actually target specific. The above looks like x86-64 basic layout string.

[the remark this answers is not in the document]
pjaaskel: Could we ask the datalayout string from the generated bitcode library?

lib/CL/devices/basic/basic.c
```diff
@@ -293,6 +293,8 @@ pocl_basic_init_device_infos(unsigned j, struct _cl_device_id* dev)
 
   dev->device_side_printf = 1;
   dev->printf_buffer_size = 16 * 1024 * 1024;
+  dev->spir_compatible_triple = SPIR_HOST_TARGET_TRIPLE;
```
pjaaskel: I still don't understand the point of this. Now you seem to use it mainly as a boolean to check whether the target supports SPIR input. A bool makes more sense here as we need to force set for the target triple in any case to match the bitcode lib's target. Later on we can add a pass to ensure alignment is explicit in all loads and stores in the SPIR input, if they are not mandated to be.

lib/CL/devices/tce/tce_common.cc
```diff
@@ -71,11 +71,14 @@ TCEDevice::TCEDevice(cl_device_id dev, const char* adfName) :
   POCL_INIT_LOCK(tce_compile_lock);
   dev->address_bits = 32;
   dev->autolocals_to_args = 1;
+  /* This assumes TCE is always Big-endian;
```
pjaaskel: Please mirror this assumption now as we consider big endian TCE almost like in deprecated status now. All new TTAs are LE basically.

lib/poclu/misc.c
```diff
@@ -53,21 +55,20 @@ poclu_create_any_context ()
 
 cl_int
 poclu_get_any_device (cl_context *context, cl_device_id *device,
-                      cl_command_queue *queue)
+                      cl_command_queue *queue, cl_platform_id *platform)
```
pjaaskel: It might be nice to allow passing NULL at least to the platform parameter as it's often not needed after initializing the context. Then it would create one internally to populate the other inputs. Lots of test changes in this patch just add an unused temp variable for this call.

examples/example2/CMakeLists.txt
```diff
@@ -23,28 +23,27 @@
...
 endif(MSVC)
 add_executable("example2" example2.c example2.cl)
 
-# example1_LDADD = @OPENCL_LIBS@ ../../lib/poclu/libpoclu.la
 target_link_libraries("example2" ${POCLU_LINK_OPTIONS})
 
-add_test("examples/example2_matrix_transpose" "example2")
+add_test("examples/example2" "example2")
+
+if (ENABLE_SPIR AND NOT (KERNELLIB_HOST_DISTRO_VARIANTS))
```
pjaaskel: Why is "KERNELLIB_HOST_DISTRO_VARIANTS"? Do you mean that SPIR is not functioning if host distro variants are built of the kernel lib? Why? If so, then would it be correct to disallow ENABLE_SPIR in case that is set?

franz:
WRT "does this test spirv input somehow somewhere?" - nope currently it does not.

lib/poclu/misc.c
franz: allright

examples/example2/CMakeLists.txt
franz: Hmm you're right, i should disable SPIR for distro builds for the time being. I haven't figured out yet why, but one of the LLVM passes wreaks havoc on SPIR-turned-to-LLVM-IR binaries.

lib/CL/devices/basic/basic.c
pjaaskel: OK, then just 2 bools? ;) Still not seeing the point of storing a datalayout string here if we force set to the bitcode lib anyhow.

pjaaskel: Or just reuse the 32b property already available in cl_device properties.

franz: We're not storing the datalayout string here, we're storing a SPIR triple only, but i'll try to simplify it.

pjaaskel: Ah right. But yes, better have option "spir_supported" and "spirv_supported" and the bitness read from the ocl device's bitness?

PR #649, opened 2018-08-06
[CUDA] Fix build for LLVM 7.0

jrprice:
The `addPassesToEmitFile` function has an additional argument.

Recommend cherry picking to the 1.2 release.


franz:
applied in both branches

PR #653, opened 2018-08-07
[CUDA] Map LLVM math intrinsics to libdevice calls

jrprice:
Just adds additional entries for the `llvm.*.f{32,64}` intrinsics to the libdevice function map.

Recommend cherry picking to the 1.2 release.


jrprice:
Merged in a4f1f4b.

PR #657, opened 2018-08-09
[CUDA] Implement proper async work-group copies

jrprice:
Use a blocked copy, with a barrier in wait_group_events().

This gets the `basic` conformance test suite passing with CUDA again.

### DIFF
--- a/lib/kernel/cuda/CMakeLists.txt
+++ b/lib/kernel/cuda/CMakeLists.txt
@@ -32,6 +32,7 @@ foreach(FILE atomics.cl)
 endforeach()
 
 foreach(FILE
+  async_work_group_copy.cl async_work_group_strided_copy.cl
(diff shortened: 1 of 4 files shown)


jrprice:
Yes, I meant the `basic` test category from the OpenCL CTS.

They're not particularly CUDA specific, but I don't know if they're best for all devices. Is the default serial copy better for the pthread backend for example? Also, the default implementation of `wait_group_events` used to have a barrier like this implementation, but it was removed in 3ce08a5, apparently because it caused some issues for CFG analysis (I guess for the pthread backend).

franz:
> Is the default serial copy better for the pthread backend for example?

Good question. It could be better for the HSA backend though.

Shall i also apply this to release branch?

jrprice:
> Shall i also apply this to release branch?

Sure, I'm fine with that. Shouldn't cause any regressions since this was just broken for CUDA before this patch.

franz:
Applied in both branches, will be in RC2

PR #658, opened 2018-08-24
Native ISA compilation on top of HSA etc

pjaaskel:


franz:
@pjaaskel the CTS results seems OK, so i believe everything should be fixed

PR #667, opened 2018-10-08
Ensure to not use CPU autodetection

bmwiedemann:
Ensure to not use CPU autodetection
when building distro packages.

Without this patch, config.h's `OCL_KERNEL_TARGET_CPU` got this value and
lib/CL/devices/basic/basic.c
lib/CL/pocl_llvm_utils.cc
lib/CL/pocl_llvm_build.cc
used the value, creating variations in build results.

Note: I tested that it builds reproducibly on different CPUs, but not that it works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -400,10 +400,9 @@ if (OCS_AVAILABLE)
 
   if(NOT DEFINED HOST_DEVICE_BUILD_HASH)
     if(KERNELLIB_HOST_CPU_VARIANTS STREQUAL "distro")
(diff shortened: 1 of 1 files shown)


franz:
Hello, are you trying to get reproducible builds ? Your patch kind-of works, but i've tried to fix it properly in git master. Please test and let me know if it works.

bmwiedemann:
Hi, yes, I'm working on reproducible builds for openSUSE, but even without that, we want to have distribution packages that do not depend on the build CPU type.

Can you please link to the relevant commit?

bmwiedemann:
Tested successfully. So this PR can be closed.

PR #675, opened 2018-11-07
Fix a few more things for multiple device use case

franz:


lib/CL/clEnqueueMapBuffer.c
```diff
@@ -82,6 +82,9 @@ POname(clEnqueueMapBuffer)(cl_command_queue command_queue,
 
   POCL_CHECK_DEV_IN_CMDQ;
 
+  // TODO are refcounts OK with Map ?
```
pjaaskel: Please fix the comment style to /* */.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -926,7 +926,7 @@ pocl_cuda_submit_kernel (CUstream stream, _cl_command_run run,
                 /* Get device pointer */
                 cl_mem mem = *(void **)arguments[i].value;
                 CUdeviceptr src
-                    = (CUdeviceptr)mem->device_ptrs[device->dev_id].mem_ptr;
+                    = (CUdeviceptr)mem->device_ptrs[device->dev_id].mem_ptr + arguments[i].offset;
```
pjaaskel: Quite long lines here?

lib/CL/pocl_build.c
```diff
@@ -616,21 +616,29 @@ compile_and_link_program(int compile_program,
...
+              error = pocl_llvm_build_program (
+                  program, device_i, program->compiler_options,
+                  program_bc_path, num_input_headers, input_headers,
+                  header_include_names, (create_library ? 0 : link_program));
+              POCL_GOTO_ERROR_ON ((error != 0), build_error_code,
+                                  "pocl_llvm_build_program() failed\n");
+            }
+          else
 #endif
+            do
```
pjaaskel: This seems like spaced indent (which is not according to the GNU's crazy mixed space+tab), but I think we should not enforce that craziness. I need to create a style file to .emacs which adheres otherwise to the GNU C style but with this.

lib/CL/devices/cuda/pocl-cuda.c
franz: yes, i forgot to run clang-format here

PR #686, opened 2019-01-20
pthread: broadcast event updates with pthread_cond_broadcast()

sdh4:
pocl_pthread_update_event() notifies event waiters about changes of status. It should use pthread_cond_broadcast() rather than pthread_cond_signal() so all waiters get notified in case multiple threads are waiting on the event.


pjaaskel:
Can you tell me your real name and email so I can add it to CREDITS? We like to list all contributors there (even minor ones).

sdh4:
Don't think this really qualifies as much of a contribution, but in any case:
Steve Holland... sdh4@iastate.edu

pjaaskel:
Well it is a huge contribution compared to the average pocl user 😅

PR #699, opened 2019-03-08
Add option to disable installing of OpenCL headers

justxi:
Currently only installing or auto detection of OpenCL headers is possible. 
With this patch it is possible to install (cmake -DINSTALL_OPENCL_HEADERS=1/YES), not install (cmake -DINSTALL_OPENCL_HEADERS=0/NO) or auto detect (cmake). 
With no option set, it behaves like before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -837,7 +837,7 @@ message(STATUS "Run tests with ICD: ${TESTS_USE_ICD}")
 if(INSTALL_OPENCL_HEADERS)
   message(STATUS "Install POCL's OpenCL headers: ${INSTALL_OPENCL_HEADERS}")
 elseif(DEFINED INSTALL_OPENCL_HEADERS AND NOT INSTALL_OPENCL_HEADERS)
-  message(STATUS "Not install POCL's OpenCL headers: ${INSTALL_OPENCL_HEADERS}")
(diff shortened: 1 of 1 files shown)


keryell:
That seems like a useful feature.

justxi:
It would be very useful when writing an ebuild to install on Gentoo Linux, because the opencl headers are installed by other packages.

CMakeLists.txt
```diff
@@ -834,8 +834,10 @@ message(STATUS "Run tests with ICD: ${TESTS_USE_ICD}")
 
 ######################################################################################
 
-if(DEFINED INSTALL_OPENCL_HEADERS)
+if(INSTALL_OPENCL_HEADERS)
   message(STATUS "Install POCL's OpenCL headers: ${INSTALL_OPENCL_HEADERS}")
+elseif(DEFINED INSTALL_OPENCL_HEADERS AND NOT INSTALL_OPENCL_HEADERS)
+  message(STATUS "Not install POCL's OpenCL headers: ${INSTALL_OPENCL_HEADERS}")
```
pjaaskel: Can you change the message simply to "Not installing OpenCL headers."

justxi: Done.

Issue #702, opened 2019-03-18
CUDA fails with LLVM 8.0


franz:
CUDA backend seems to have one new failure with LLVM 8.0:

```
21/70 Test  #16: kernel/test_rotate ................................................................***Failed  Error regular expression found in output. Regex=[FAIL]  9.57 sec
CMake Error at /home/buildslave/basedir/quick_80_cuda/source/cmake/run_test.cmake:34 (message):
  FAIL: Test exited with nonzero code (Child aborted):
  /home/buildslave/basedir/quick_80_cuda/build/tests/kernel/kernel;test_rotate


  STDOUT:

  Running test test_rotate...

  FAIL: rotate type=

  STDERR:

  CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered
```

pjaaskel:
Is this due to LLVM 8.0, that is, it works with pocl master and LLVM 7.0?

franz:
Yes, it works with git master + LLVM 7.0

pjaaskel:
@jrprice any insight?

pjaaskel:
This is possibly an LLVM regression, so we won't block pocl 1.3 due to it. If you @jrprice have time to check into it, great, but otherwise we just document it as an XFAIL.

jrprice:
I've reproduced the error on one of my machines, so I'll try and take a look soon. I agree this shouldn't block the 1.3 release.

jrprice:
I've dug into this a bit today, just recording progress here. The crash was coming from the `printf` calls - this is fixed by #707. The fact that we're getting to the `printf` calls when we weren't before still implies an LLVM regression (i.e. we're actually getting the wrong results from the `rotate` builtin). It looks like a `llvm.fshl` intrinsic was added in 8.0, so probably an issue with how this is being lowered to PTX.

jrprice:
Reported LLVM bug:
https://bugs.llvm.org/show_bug.cgi?id=41347

pjaaskel:
Thanks, so indeed an LLVM regression. I'll keep the ticket open regardless if people search for this.

pjaaskel:
It works with the latest LLVMs, so not keeping this open. If someone cares for older LLVMs+CUDA, a PR is welcome.

PR #707, opened 2019-04-02
[CUDA] Improve printf handling

jrprice:
This fixes a bug that was causing kernels to crash when printing strings using `%s` (as seen in #702), and adds a couple of other improvements that helps when printing errors from pocl's internal tests.

NOTE: #702 is still an issue, it just now actually tells you the test failed rather than crashing


pjaaskel:
Is this a regression fix, that is, should we apply it to the release branch also?

jrprice:
It's not a regression, this issue is present in the previous releases but we were just never hitting `%s` formats until the LLVM regression in #702.

pjaaskel:
Since we need RC2 anyways due to a regression fix, I pulled this in. Please try to test the upcoming RC2 with CUDA. We might release before the weekend.

PR #710, opened 2019-04-10
fix name of Khronos

moeryomenko:


moeryomenko:
@keryell 
Changed "Khronos" to "Khronos Group"

pjaaskel:
Thanks. Can you limit the changes to the mentioned fix? Now it includes also random white space fixes, which we prefer to commit gradually only while editing the code around the spots.

PR #719, opened 2019-05-01
Misc. updates

pjaaskel:


lib/CL/devices/devices.c
```diff
@@ -416,23 +413,25 @@ pocl_init_devices ()
      In case libhwloc has the OpenCL plugin installed, it initializes
      it and it leads to initializing pocl again which leads to an
      infinite loop. */
-
+  POCL_LOCK (pocl_init_lock);
```
franz: If you want to avoid the problem described in few lines above, you cannot lock here; it would result in a deadlock. IIRC the original fix we agreed upon, was to move the unlock call after `pocl_reinit_devices()` - did you find it insufficient ?

If we still care about the old hwloc bug, the lines until 430 or so need to be restored to original; if we don't care about that bug anymore, this code still needs a bit of fixing - it doesn't make sense to check for `init_in_progress` if you're holding the init lock.

PR #734, opened 2019-05-16
Device-side allocation of local memory to stack

pjaaskel:
Allows devices to set "device_alloca_locals" to 1, which causes then Workgroup.cc to generate code that allocates local memory from stack (using the alloca LLVM IR instruction) instead of expecting it to be preallocated by the host launcher. In that case the hidden local args contain the size of the local buffer to allocate instead of the buffer pointer.

### DIFF
--- a/include/pocl.h
+++ b/include/pocl.h
@@ -44,6 +44,12 @@
 
 #define POCL_FILENAME_LENGTH 1024
 
+#define SPIR_ADDRESS_SPACE_PRIVATE 0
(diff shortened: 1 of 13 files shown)


include/pocl.h
```diff
@@ -44,6 +44,12 @@
 
 #define POCL_FILENAME_LENGTH 1024
 
+#define SPIR_ADDRESS_SPACE_PRIVATE 0
```
franz: These are already defined in `include/pocl_spir.h`

PR #737, opened 2019-05-22
Minor bugfixes & cleanups

franz:


tests/workgroup/for_bug.cl
```diff
@@ -43,7 +43,7 @@ test_kernel (void)
    barrier tail replication.
 */
 
-  for (i = 16; i >= 0; i--) {
+  for (i = 16; i > 0; i--) {
```
pjaaskel: Wouldn't it be more correct to just XFAIL this test for ARM32?

franz: I believe it wouldn't. From the description in the testcase:

> This produces a crash with 'repl' and an infinite loop with 'wiloops'.

The testcase is run with `POCL_WORK_GROUP_METHOD=workitemloops` and it didn't crash or go into an infinite loop, instead it produced wrong output. Also, i'm not changing the relation between the `for` and the `if` conditions in the test. So this looks like an entirely different bug triggered by something new in LLVM 7.0+ (it worked on ARM32 until 6.0).

pjaaskel: Fair enough.

pjaaskel:
Can you add at least the minimum LLVM version bump to CHANGES?

franz:
Added

PR #742, opened 2019-06-03
Remove support for old LLVM versions

franz:
We've already disabled them in CMake but this removes also the code.


include/pocl_file_util.h
```diff
@@ -48,8 +48,6 @@ int pocl_rename(const char *oldpath, const char *newpath);
 
 int pocl_exists(const char* path);
 
-int pocl_filesize(const char* path, uint64_t* res);
```
pjaaskel: Why do you remove this function? If it's not called at all now, maybe we can keep it around for later needs for some time?

include/_kernel.h
```diff
@@ -209,18 +163,7 @@
...
-#undef CLANG_HAS_IMAGE_AS
-#define IMG_WO_AQ
-#define IMG_RO_AQ
-#define IMG_RW_AQ
-
-#else
-
-/* 3.9 supports both read and write access qualifiers */
-
+// TODO this is always defined
```
pjaaskel: Address or remove this TODO?

pjaaskel: At the very least pls change the comment to /* */ ;)

lib/CL/pocl_llvm_build.cc
```diff
@@ -698,10 +669,7 @@ int pocl_llvm_link_program(cl_program program, unsigned device_i,
     llvm::Module *p = (llvm::Module *)cur_llvm_irs[i];
     assert(p);
 
-#ifdef LLVM_OLDER_THAN_3_8
-    if (Linker::LinkModules(mod, llvm::CloneModule(p))) {
-      delete mod;
-#elif LLVM_OLDER_THAN_7_0
+#if LLVM_OLDER_THAN_7_0
```
pjaaskel: #ifdef

lib/kernel/svm_atomics_x86_64.ll
```diff
@@ -1,8072 +0,0 @@
-; ModuleID = 'svm_atomics_host.cl.bc'
```
pjaaskel: Why this file is deleted?

include/pocl.h
```diff
@@ -103,14 +103,15 @@ typedef struct
   void *wg; /* The work group function ptr. Device specific. */
   cl_kernel kernel;
   /* The launch data that can be passed to the kernel execution environment. */
-  struct pocl_context pc;
   struct pocl_argument *arguments;
-  /* Can be used to store/cache arbitrary device-specific data. */
+  struct pocl_context pc;
```
pjaaskel: Why is this movement done?

include/pocl_context.h
```diff
@@ -31,25 +31,25 @@
 #include "pocl_types.h"
 
 struct pocl_context32 {
-  uint work_dim;
```
pjaaskel: Why?

lib/CL/devices/devices.c
```diff
@@ -75,6 +75,7 @@
 /* the enabled devices */
 static struct _cl_device_id* pocl_devices = NULL;
 unsigned int pocl_num_devices = 0;
+unsigned int pocl_implicit_finish = 0;
```
pjaaskel: Did this have significant performance impact? If so, it'd be better to optimize pocl_is_option_set() instead?

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -442,18 +372,99 @@ pocl_pthread_prepare_kernel (void *data, _cl_command_node *cmd)
...
+    cl_device_id subd = cmd->device;
+    if (shall_we_run_this (td, subd))
+      return cmd;
+  }
+
+  return NULL;
+}
+
+static int
+pthread_scheduler_get_work (thread_data *td)
```
pjaaskel: This function needs comments on why it does what it does in a way that it avoids "troubles with subdevices". This won't be remembered after 6 months without documentation.

lib/CL/devices/bufalloc.c
```diff
@@ -104,7 +104,7 @@ chunk_slack (chunk_info_t* chunk, size_t size, size_t* last_chunk_size)
   size_t end_addr = aligned_start_addr + size;
   if(last_chunk_size)
     *last_chunk_size = end_chunk - end_addr;
-  return end_chunk > end_addr;
```
pjaaskel: Why this change? What is the broken logic in the original?

lib/CL/clReleaseContext.c
```diff
@@ -64,17 +66,42 @@ POname(clReleaseContext)(cl_context context) CL_API_SUFFIX__VERSION_1_0
...
+  otherwise abort with an error.
+
+  Useful for various sanitizer checks.
+*/
+
+#ifndef ENABLE_ICD
+void
+pocl_check_uninit_devices ()
+{
+  /* This is a sleep with rather arbitrary amount, but it's required.
```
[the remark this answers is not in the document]
franz: There is (something wrong). I'll add more comments in commit/code

include/poclu.h
```diff
@@ -38,6 +38,12 @@
 extern "C" {
 #endif
 
+#ifdef ENABLE_DEVICE_UNINIT
+void pocl_check_uninit_devices();
+#else
+#define pocl_check_uninit_devices() NULL
```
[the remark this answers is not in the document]
franz: > I feel this commit adds a lot code to hide

This particular code adds a function that's normally compiled to NULL and is only compiled to a hidden pocl function used by tests, when compiling without ICD (for sanitizer tests).

lib/CL/devices/pthread/pthread_scheduler.c
franz: OK

lib/CL/devices/devices.c
franz: It popped up in cachegrind IIRC.

lib/CL/devices/bufalloc.c
franz: In a situation where 1) you have X bytes of (aligned) free device memory in `region->last_chunk` 2) you call  `chunk_slack(region->last_chunk, X, NULL);` the original code returns 0 (false). IOW at most X-1 free bytes are usable.

examples/PyOpenCL/build.sh
```diff
@@ -4,4 +4,4 @@ TS_BASEDIR="$1"
 TS_BUILDDIR="$2"
 TS_SRCDIR="$3"
 
-source "${TS_BUILDDIR}/bin/activate" && cd "${TS_SRCDIR}" && python3 setup.py build
```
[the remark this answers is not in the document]
franz: From https://docs.python.org/3/library/venv.html

> You don’t specifically need to activate an environment; activation just prepends the virtual environment’s binary directory to your path, so that “python” invokes the virtual environment’s Python interpreter and you can run installed scripts without having to use their full path. However, **all scripts installed in a virtual environment should be runnable without activating it, and run with the virtual environment’s Python automatically.**

1) slightly shorter code 2) explicit paths instead of env variables

lib/CL/devices/devices.c
franz: ... i don't remember the exact impact, but i can measure again. Yes it's possible `pocl_is_option_set()` would be a better place

include/poclu.h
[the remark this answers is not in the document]
pjaaskel: (I was referring to the whole commit since there was no way to comment a commit)

lib/CL/devices/bufalloc.c
pjaaskel: OK. Hard to remember anymore.

examples/PyOpenCL/build.sh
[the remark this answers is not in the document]
pjaaskel: ok

lib/CL/devices/devices.c
pjaaskel: Somehow difficult to believe that it'd be a significant bottleneck, but yep pocl_is_option_set() should have then a caching in place. Is this implicit finish really needed for something still? I'd consider dropping the whole feature.

lib/kernel/svm_atomics_x86_64.ll
franz: There were multiple versions of "atomics" file for various older LLVM versions, because of changes around atomics in bitcode or builtins, i can't remember now. Those LLVM versions can be seen in lines removed from `lib/kernel/host/CMakeLists.txt` in same commit.

lib/CL/devices/devices.c
franz: It's not necessary, AFAICT it was always purely for debugging. I have never used it, and i'm not aware of anyone using it.

include/pocl_context.h
franz: This was a micro-optimization done as part of larger work (which was already merged by you earlier, somewhere in the specialization work IIRC). So essentially a "leftover".

include/_kernel.h
franz: Yep this i forgot, will address

franz:
Rebased on master, fixes applied, removed `POCL_IMPLICIT_FINISH`, also removed `POCL_USE_FAKE_ADDR_SPACE_IDs` and `TargetAddressSpaces.cc`.

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -442,18 +374,121 @@ pocl_pthread_prepare_kernel (void *data, _cl_command_node *cmd)
...
+  cmds A2, A3, B2 are pushed to C
+  B finishes processing B1, checks queue head, A2 isn't for it, goes to sleep
+  A finishes processing A1, processes A2 + A3 but ignores B2, it's not for it
+  application calls clFinish to wait for queue
+
+  ...now B is sleeping and not possible to wake up
+  (since no new commands can arrive) and there's a B2 command
+  which will never be processed.
+
+  it's possible to workaround but it's cleaner to just check the whole queue.
```
pjaaskel: Wouldn't it be cleaner to have a queue per (sub)device?

PR #746, opened 2019-07-19
SPIR and SPIR-V improvements

franz:
plus a few minor fixes


CHANGES
```diff
@@ -9,6 +9,9 @@ Highlights
   CL_DEVICE_TYPE_CUSTOM hardware accelerators which implement a memory
   mapped control interface.
 
+- Improved SPIR and SPIR-V support. clCreateProgramWithIL() implemented,
```
pjaaskel: Did SPIR-V work previously? Better advertise it more if this is the first release where it's usable.

lib/CL/devices/basic/basic.c
```diff
@@ -368,6 +368,7 @@ pocl_init_cpu_device_infos (cl_device_id dev)
   dev->llvm_cpu = get_llvm_cpu_name ();
 #endif
 
+  dev->spirv_version = "SPIR-V_1.2";
```
pjaaskel: Why does the spirv_version include the prefix "SPIR-V_" ?

lib/CL/pocl_cl.h
```diff
@@ -794,6 +794,8 @@ struct _cl_device_id {
   cl_uint dev_queue_max_size;
   cl_command_queue_properties on_dev_queue_props;
   cl_command_queue_properties on_host_queue_props;
+  /* OpenCL 2.1 */
```
pjaaskel: What does this comment want to say? That SPIR-V is a requirement since 2.1, I think?

lib/CL/pocl_file_util.c
```diff
@@ -107,20 +107,6 @@ pocl_exists(const char* path)
   return !access(path, R_OK);
 }
 
-int
```
pjaaskel: Is it necessary to remove this function even though we don't use it currently?

lib/kernel/SPIR/generate_spir_layer.py
```diff
@@ -0,0 +1,306 @@
+#!/usr/bin/python3
+#
+# A script to generate functions (SPIR-mangled with SPIR AS) that will call
```
pjaaskel: Like we discussed, I think this is good enough for now. Later we might consider integrating this functionality to linker.cpp so it maps the SPIR-V AS0 ones on the fly to the target library ones.

lib/kernel/spir_layer.ll
```diff
@@ -0,0 +1,2668 @@
+; ModuleID = 'spir_layer.bc'
```
pjaaskel: Better name would be spir_wrapper.ll, also the script would be more descriptive as generate_spir_wrapper.py.

lib/llvmopencl/LLVMFileUtils.cc
```diff
@@ -240,44 +233,48 @@ int pocl_mk_tempname(char *output, const char *prefix, const char *suffix,
...
     Twine p(path);
 
-    *content = NULL;
-
-    int errcode = pocl_filesize(path, filesize);
-    if (errcode)
-      return errcode;
-
-    size_t fsize = (size_t)(*filesize);
+    /* files in /proc return zero size, while
```
pjaaskel: Did you run the new clang formatter? These comments seem to not fill the line.

lib/CL/devices/basic/basic.c
franz: The `clGetDeviceInfo` spec says so: 

> Returns a space-separated list of IL version strings of the form <IL_Prefix>_<Major_Version>.<Minor_Version>. For OpenCL 2.1, "SPIR-V" is a required IL prefix.

CHANGES
franz: It did, sort of. It wasn't very usable (like SPIR) because of mangling.

lib/kernel/spir_layer.ll
franz: OK, will rename

lib/llvmopencl/LLVMFileUtils.cc
franz: Yes, i did. It doesn't do anything on that commit.

lib/CL/pocl_cl.h
franz: it's "OpenCL 2.1 related properties follow", same as OpenCL 2.0 comment a few lines above

lib/CL/pocl_file_util.c
franz: Not really, but it's somewhat dangerous to use, since the reported size isn't correct on some filesystems.

lib/kernel/SPIR/generate_spir_layer.py
franz: Yes. It should be also fairly easy to extend it to support TCE and CUDA, both use SPIR AS numbers.

CHANGES
pjaaskel: Ok, let's make it explicit.

lib/CL/devices/basic/basic.c
pjaaskel: ok, didn't know it was standardized

PR #779, opened 2019-09-27
Link against libclang-cpp.so when linking with shared libraries

tstellar:
libclang-cpp.so is a new shared object in clang-9.0.0 which
contains all the clang components in one library.  Doing this will
save us from having to update cmake when there is a new clang component
and may provide better performance in LTO builds of clang.

### DIFF
--- a/cmake/LLVM.cmake
+++ b/cmake/LLVM.cmake
@@ -219,10 +219,16 @@ string(REPLACE " -pedantic" "" LLVM_CXXFLAGS "${LLVM_CXXFLAGS}")
 string(REGEX REPLACE "-W[^ ]*" "" LLVM_CXXFLAGS "${LLVM_CXXFLAGS}")
 
 # Llvm-config does not include clang libs
-set(CLANG_LIBNAMES clangCodeGen clangFrontendTool clangFrontend clangDriver clangSerialization
-    clangParse clangSema clangRewrite clangRewriteFrontend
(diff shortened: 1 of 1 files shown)


tstellar:
I've tested this with the build configuration we use for Fedora, but need help testing on other configurations.

franz:
Hi,

Is libclang-cpp always built, in all LLVM build configurations (static and dynamic)  ? Asking because your PR enables it unconditionally (for LLVM >= 9).

tstellar:
Yes, it's always built.

tstellar:
I actually just ran into this same problem too.  This happens when clang is liked without the --as-needed flag.  I think this will need a change to clang's cmake scripts in order to fix properly.

franz:
@tstellar we can pull this PR, with a little change to the `if` condition. It doesn't work with LLVM 9.0 but that's OK since we already have that handled. If you can get the `--as-needed` fix into next LLVM point release (9.0.1) then we need `if (LLVM_VERSION_FULL VERSION_GREATER 9 AND ..)`, otherwise `if (LLVM_VERSION_FULL VERSION_GREATER_EQUAL 10 AND ..)` would work.

franz:
@tstellar i have tested with ``release/10.x`` branch, and it's still broken. Can you get a fix into Clang CMake scripts for release 10 ?

franz:
Fixed & merged

PR #780, opened 2019-09-27
Packages

KOLANICH:


KOLANICH:
Any progress on merging this?

CMakeLists.txt
```diff
@@ -23,9 +23,10 @@
 #
 #=============================================================================
 
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15.2 FATAL_ERROR)
```
franz: This would make pocl unbuildable on every major Linux distribution. At the very least we want it working on the latest Ubuntu LTS (18.04) and Debian (10).

pjaaskel: Why bump this up? Which features requires a newer cmake?

CMakeLists.txt
```diff
@@ -1431,6 +1432,26 @@ add_subdirectory("examples")
...
+set(CPACK_PACKAGE_CONTACT "https://github.com/pocl/pocl")
+set(CPACK_PACKAGE_CHECKSUM "SHA512")
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_SOURCE_DIR}/COPYING")
+set(CPACK_GENERATOR "DEB")
+
+string(REPLACE "." ";" LLVM_VERSION_PARSED "${LLVM_VERSION}")
+list(GET LLVM_VERSION_PARSED 0 LLVM_VERSION_MAJOR)
+list(GET LLVM_VERSION_PARSED 1 LLVM_VERSION_MINOR)
+
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "clang-${LLVM_VERSION_MAJOR} (>= ${LLVM_VERSION_MAJOR}),libgcc1,libhwloc5 (>= ${Hwloc_VERSION}),libllvm${LLVM_VERSION_MAJOR} (>= ${LLVM_VERSION_MAJOR}),libc6")
```
franz: Hwloc is optional, and also LLVM is optional, so the dependencies here should be ifdef-ed on ENABLE_HWLOC and OCS_AVAILABLE cmake variables.

CMakeLists.txt
KOLANICH: No deep reason. I just have not tested on other versions. Feel free to downgrade to the earliest version you know it builds properly.

CMakeLists.txt
KOLANICH: Thanks. Gonna fix.

CMakeLists.txt
```diff
@@ -23,16 +23,27 @@
 #
 #=============================================================================
 
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5.2 FATAL_ERROR)
```
pjaaskel: This is still to fix (at least). Why bump the version up?

CMakeLists.txt
pjaaskel: I think the opposite way: I think it's better to keep it as is until we know there is a feature we need from higher versions.

CMakeLists.txt
KOLANICH: <details >
<summary>Versions not less than this one are widely available according to repology data:</summary>

[](https://repology.org/project/cmake/versions)

</details>

CMakeLists.txt
pjaaskel: I think the "official policy" in pocl we set place was to support the latest Debian and latest Ubuntu LTS with community support, so this includes Ubuntu 16.04, but not anymore 14.04. 

Ubuntu 16.04 ships 3.5.1 and Debian 10 ships 3.13.4. So from that perspective we could bump that to 3.5.1. But again, why? If we don't know of any issue with 2.8.12, why make the life (possibly) harder for some older distro which is stuck with Cmake?

KOLANICH: I don't remember why. Probably something related to some modules used. I can downgrade the required version though, but I am not sure that it will work on their distros. IMHO  if they have an old CMake, they should just upgrade it. Kitware ships prebuilt binaries. I personnaly have set up a deb repo with a bleeding edge CMake and use it.

pjaaskel: > I don't remember why. 

OK, then please just revert it so we can move forward with this PR.

KOLANICH: d94c3a5e647e64876c692a470ba62e5b703735d0

CMakeLists.txt
KOLANICH: d94c3a5e647e64876c692a470ba62e5b703735d0

CMakeLists.txt
pjaaskel: Clang/LLVM? I think the main problem there would be possibly mismatching versions between the builtin IR lib and the Clang/LLVM it happens to find at runtime.

KOLANICH: No, hwloc and ocs.

KOLANICH:
Is it going to be merged?

pjaaskel:
```
CMake Error at lib/CMakeLists.txt:39 (list):
  list does not recognize sub-command JOIN
```
This is with cmake 3.10.2 of Ubuntu 18.04 LTS. Please avoid JOIN here.

KOLANICH:
@pjaaskel

KOLANICH:
@pjaaskel

KOLANICH:
1Thanks.

Issue #796, opened 2019-12-07
Poor codegen from integral vector conversion on ARM64


blu:
Hi,

Doing a conversion of the type `convert_intN(shortN)` emits poor target code on arm64. Example:

```
__kernel
void monokernel(
    __global const ushort4* const src_a,
    __global const ushort4* const src_b,
    __global int8* const dst)
{
    const int8 m = convert_int8((ushort8)(0) == (ushort8)(*src_a, *src_b));
    *dst = m;
}
```

```
    ldr    d0, [x9]
    ldr    d1, [x10]
    mov    v0.d[1], v1.d[0]
    cmeq   v0.8h, v0.8h, #0
    str    s0, [sp, #4]
    ldrsh  w17, [sp, #6]
    ldrsh  w18, [sp, #4]
    st1    {v0.s}[1], [x14]
    ldrsh  w0, [sp, #2]
    ext    v0.16b, v0.16b, v0.16b, #8
    ldrsh  w1, [sp]
    str    s0, [sp, #12]
    ldrsh  w2, [sp, #14]
    fmov   s1, w18
    ldrsh  w18, [sp, #12]
    mov    v1.s[1], w17
    fmov   s2, w18
    mov    v2.s[1], w2
    st1    {v0.s}[1], [x15]
    mov    v1.s[2], w1
    ldrsh  w17, [sp, #8]
    mov    v2.s[2], w17
    mov    v1.s[3], w0
    ldrsh  w17, [sp, #10]
    mov    v2.s[3], w17
    stp    q1, q2, [x11]
```

Result from the vector comparison gets spilled onto the stack and subsequently read into GPR regs via `ldrsh`, those GPRs are later packed into SIMD regs `v1` and `v2`, and eventually written into `dst` buffer. This given that arm64 has native integral conversion vector ops.

Issue reproducible with both `release_1_3` and `release_1_4` branches.

blu:
How can I examine the IR that produced this codegen?

pjaaskel:
Check: http://portablecl.org/docs/html/env_variables.html envs POCL_CACHE_DIR and POCL_LEAVE_KERNEL_COMPILER_TEMP_FILES are what you are looking for. parallel.bc is the final multi-WI WG IR before passing to code gen after all IR/pocl optimizations/transformations.

blu:
The problem is already present in the .ll (don't mind the excessive args)

```
__kernel __attribute__ ((vec_type_hint(float4)))
void monokernel(
[...]

PR #797, opened 2019-12-11
refactor vector conversions for better autovectorizing codegen

blu:
Change replaces the original lo/hi decomposition of `convert_typeN(sourceN)` for a decomposition of the kind `(typeN)((type) source.s0, (type) source.s1, ... )`, which emits `sext`/ `zext` llvm IR opcodes (for integral types), which, in their turn, autovectorize nicely on ISAs with corresponding ops. Change addresses issue #796.


pjaaskel:
Can you add a CHANGES entry (preferably with some speedup estimates from this) and yourself to CREDITS? Thanks.

blu:
@pjaaskel I've updated CHANGES section for 1.5. CREDITS untouched as I'm already present.

PR #798, opened 2019-12-13
Also split if the resulting group size is equal to the minimum group size

new2f7:
clEnqueueNDRangeKernel(): When hitting a `global_work_size := global_x * global_y * global_z` that equals `4 * preferred_wg_multiple * max_compute_units`, the program will now run in parallel (new behaviour) rather than not running in parallel (old behaviour).
- If `global_work_size` is smaller, the program will not run in parallel (unchanged).
- If `global_work_size` is bigger, the program will run in parallel (unchanged).

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -69,3 +69,4 @@ Maxim Eremenko <moeryomenko@gmail.com>
 Andrew Gozillon <andrew.gozillon@yahoo.com>
 Jan Solanti <jan.solanti@tuni.fi>
 Stefan Brüns <stefan.bruens@rwth-aachen.de>
+Tobias Baumann <t.baumann@agdsn.de>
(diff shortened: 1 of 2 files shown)


pjaaskel:
Good catch. Can you add this to CHANGES (a mention of known perf. impact on some benchmark of interest would be optimal) and yourself to CREDITS? Thanks.

new2f7:
I updated the description of this PR to clarify my proposed modification. In my opinion, this modification is a small bug fix that only applies to an edge case, nothing worth mentioning in CHANGES (cf. "notable bug fixes"). However, I can add something like the first sentence from the description to CHANGES if you prefer.

Regarding performance impacts: I used the [hexciton_benchmark](https://github.com/noma/hexciton_benchmark/) (proxy benchmark derived from a [HEOM](https://en.wikipedia.org/wiki/Hierarchical_equations_of_motion) implementation) with a custom size that hits the described edge case above, see [this pdf](https://www2.informatik.hu-berlin.de/~baumanty/all.pdf). I originally encountered this problem when conducting measurements with an
unpublished micro benchmark, see [this pdf](https://www2.informatik.hu-berlin.de/~baumanty/split.pdf).

PR #804, opened 2020-01-31
Fix a crash when a null kernel object is passed to clEnqueueNDRangeKernel

alberto-cerato:
It looks like the 'kernel' parameter is used before is checked for nullness.
If you pass NULL as kernel argument to clEnqueueNDRangeKernel, the POCL library crashes.
This PR simply moves the nullness check before 'kernel' is actually used.

p.s.
I am not an OpenCL expert and new to this project, so maybe there is a reason I cannot see for this.


alberto-cerato:
Hello @pjaaskel,
is there anything else I could/should do from my side for this to be merged?

Also, I assumed 'master' to be the right branch to merge this PR, but maybe that is not the case.

PR #805, opened 2020-02-07
LLVM11 support

KOLANICH:
1. fixed the mess in macrodefs making adding support of new versions a nightmare
2. brought support to llvm and clang 11

### DIFF
--- a/cmake/LLVM.cmake
+++ b/cmake/LLVM.cmake
@@ -37,6 +37,7 @@ else()
   # search for any version
   find_program(LLVM_CONFIG
     NAMES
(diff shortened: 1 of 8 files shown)


franz:
@KOLANICH 

> it still doesn't compile because of nasty template problems each is half of screen high

Possibly fixed by commit 76ce2f9f2fa8c24ead879 ? i've run into that earlier, it can produce a huge amount of perplexing errors...

BTW this PR will have to wait until we branch pocl 1.5 for the LLVM 10 release, since we don't want to claim LLVM 11 compatibility in the release branch.

KOLANICH:
`master` is not a release branch.

KOLANICH:
Thanks, that works!

PR #806, opened 2020-03-19
Fix uneven distribution of work groups to threads

new2f7:
### Problem ([L208](https://github.com/pocl/pocl/blob/master/lib/CL/devices/pthread/pthread_scheduler.c#L208) and [L210](https://github.com/pocl/pocl/blob/master/lib/CL/devices/pthread/pthread_scheduler.c#L210))
Integer division discards fractional part. To avoid result `0`, `1` is added (poor man's `ceil()`). If `num_threads` divides `k->remaining_wgs` evenly, `1` is added anyway and the result is to high. This results in an uneven distribution of work groups to threads. An example: There are 4 work groups and 4 threads. For the first thread, `num_threads` divides `k->remaining_wgs` evenly (`4/4=1`), giving 2 work groups to the first thread. Thread 2 and 3 get 1 work group each. There is nothing left to do for the 4th thread which idles, while the first thread has double work.
```
if (k->remaining_wgs <= (scaled_max_wgs * num_threads))
  max_wgs = min (scaled_min_wgs, (1 + k->remaining_wgs / num_threads));
else
  max_wgs = min (scaled_max_wgs, (1 + k->remaining_wgs / num_threads));
```
### Solution
Floating point division and `ceil()`:
```
int my_wgs = ceil ((float) k->remaining_wgs / (float) num_threads);
if (k->remaining_wgs <= (scaled_max_wgs * num_threads))
  max_wgs = min (scaled_min_wgs, my_wgs);
else
  max_wgs = min (scaled_max_wgs, my_wgs);
```
See [before.png](https://www2.informatik.hu-berlin.de/~baumanty/before.png) and [after.png](https://www2.informatik.hu-berlin.de/~baumanty/after.png).

### DIFF
--- a/lib/CL/devices/pthread/pthread_scheduler.c
+++ b/lib/CL/devices/pthread/pthread_scheduler.c
@@ -190,6 +190,7 @@ get_wg_index_range (kernel_run_command *k, unsigned *start_index,
   const unsigned scaled_max_wgs = POCL_PTHREAD_MAX_WGS * num_threads;
   const unsigned scaled_min_wgs = POCL_PTHREAD_MIN_WGS * num_threads;
 
+  unsigned limit;
   unsigned max_wgs;
   POCL_FAST_LOCK (k->lock);
   if (k->remaining_wgs == 0)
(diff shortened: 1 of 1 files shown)


lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -205,10 +207,12 @@ get_wg_index_range (kernel_run_command *k, unsigned *start_index,
    * num_threads, otherwise fallback to smaller workgroups.
    */
   if (k->remaining_wgs <= (scaled_max_wgs * num_threads))
-    max_wgs = min (scaled_min_wgs, (1 + k->remaining_wgs / num_threads));
+    limit = scaled_min_wgs;
   else
-    max_wgs = min (scaled_max_wgs, (1 + k->remaining_wgs / num_threads));
+    limit = scaled_max_wgs;
 
+  const unsigned wgs_per_thread = ceil ((float) k->remaining_wgs / (float) num_threads);
```
noma: I think a manual `ceil()` using integer-arithmetic is a cleaner solution here:

https://stackoverflow.com/questions/2745074/fast-ceiling-of-an-integer-division-in-c-c

new2f7:
I will close this PR and open a new one because it now contains wrong commits.

new2f7:
I closed this PR and will open a new one because this PR now contains wrong commits.

Issue #613, opened 2018-02-08
POCL_VECTORIZER_REMARKS broken?


pjaaskel:

franz:
Definitely doesn't do anything, but asserts that prevent invalid options are not triggered. So likely we're missing some option or somehing...

PR #842, opened 2020-06-13
Print loop vertorizer remarks when using poclcc

Kazhuu:
If POCL_VECTORIZER_REMARKS environment variable is set, print vectorizer remarks. Fixes issue #613


pjaaskel:
Can you make a mention of this fix in CHANGES and and yourself to the list of CREDITS? Thanks.

PR #848, opened 2020-07-07
Added some hardening.

KOLANICH:


KOLANICH:
>Can you add documentation to the sphinx documentation on this build mode and what it adds?

I don't think it makes any sense. IMHO the stuff mentioned there shouldn't be explicitly mentioned. IMHO it should be the defaults of all the compilers, and if someone wants to opt out, he needs to do it explicitly.

cmake/Hardening.cmake
```diff
@@ -0,0 +1,205 @@
...
+	
+	set(FLAGS_HARDENED ${FLAGS_UNHARDENED})
+	list(APPEND FLAGS_HARDENED ${HARDENING_${property}})
+	string(REPLACE ";" " " FLAGS_HARDENED "${FLAGS_HARDENED}")
+	#message(STATUS "${target} PROPERTIES ${property} ${FLAGS_HARDENED}")
+	set_target_properties(${target} PROPERTIES ${property} "${FLAGS_HARDENED}")
+endfunction(processFlagsList)
+
+function(setupPIC target)
+	set_property(TARGET ${target} PROPERTY POSITION_INDEPENDENT_CODE ON) # FUCK, doesn't work
```
pjaaskel, line 48: Is that profanity really necessary here?

cmake/Hardening.cmake
```diff
@@ -0,0 +1,205 @@
...
+			"-Wl,-flto"
+		)
+		if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+			list(APPEND HARDENING_LINKER_FLAGS
+				"-Wl,--export-all-symbols"
+				"-Wl,--nxcompat"
+				"-Wl,--dynamicbase"
+			)
+			if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+			#	list(APPEND HARDENING_LINKER_FLAGS "-Wl,--image-base,0x140000000") # doesn't work for this project
```
pjaaskel, line 171: Can you remove the unneeded commented out codes?

cmake/Hardening.cmake
KOLANICH: No, I just copied that file from my tool.

cmake/Hardening.cmake
KOLANICH: I probably should uncomment it. It was not working for my Qt-based project, but may work on this one. Needs testing on a Windows machine.

CMakeLists.txt
```diff
@@ -355,6 +355,13 @@ endif()
 ######################################################################################
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+option(HARDENING_ENABLE "Enable hardening against various attacks. May worsen performance" OFF)
+if(HARDENING_ENABLE)
```
pjaaskel, line 359: Can you change this to ENABLE_HARDENING (other flags seem to start with ENABLE) and add a sentence to CHANGES about this feature? Thanks.

pjaaskel (changes requested):

CMakeLists.txt
KOLANICH: I don't think it makes any sense. `HARDENING_` prefix has been chosen because prefixes form collapsible groups in cmake-gui.

pjaaskel: Ok, I rarely use ccmake, didn't know. So the rest of the option names are wrong?

KOLANICH: What do you mean under `wrong`? The introduced options begin from `HARDENING_`. The old options haven't been touched.

pjaaskel: I just meant that if PoCL's old options are named ENABLE_XX they were added without knowing of such a semi-official (?) CMake config variable naming standard that is detected by the ccmake UI. So nevermind this comment, but the CHANGES is still valid.

CHANGES
```diff
@@ -37,6 +37,18 @@ Notable Bug Fixes
 - Fix an issue in which the loop vectorizer produced code with invalid
   memory reads (issue #757).
 
+Security
+--------
+
+Added a script to automatically apply hardening flags present in some modern
+compilers. Applying these flags may worsen performance of the compilation a
```
pjaaskel, line 44: I still think this is long enough to go to the user manual.  Similarly to this:  https://github.com/pocl/pocl/pull/859/commits/124a04f957e69fdfdc29eb1aa6f9b7305ecf5402 

In CHANGES a sentence like "Added a build parameter HARDENING_ENABLE that applies hardening flags present in some modern compilers to produce a more secure libpocl.so with the trade-off in performance." suffices.

KOLANICH: done.

PR #850, opened 2020-07-09
WI loops: ensure loads in parallel loops are safe to execute unconditionally

mattwala:
This change checks that loads in parallel loops are safe to execute
unconditionally before adding parallel loop metadata to the
instruction. LLVM makes this assumption. Without this change,
it's possible code can be generated that reads out of bounds
by the loop vectorizer.

See issue #757

### DIFF
--- a/lib/llvmopencl/ParallelRegion.cc
+++ b/lib/llvmopencl/ParallelRegion.cc
@@ -441,6 +441,12 @@ ParallelRegion::Verify()
   return true;
 }
 
+#ifdef LLVM_OLDER_THAN_8_0
+#define PARALLEL_MD_NAME "llvm.mem.parallel_loop_access"
+#else
+#define PARALLEL_MD_NAME "llvm.access.group"
+#endif
+
 /**
  * Adds metadata to all the memory instructions to denote
  * they originate from a parallel loop.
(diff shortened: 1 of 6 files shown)


lib/llvmopencl/ParallelRegion.cc
```diff
@@ -441,47 +441,61 @@ ParallelRegion::Verify()
...
+ *     !2 distinct !{}
  *
- * !0 { metadata !1, metadata !2}
- * !1 { metadata !1 }
- * !2 { metadata !2 }
+ * Parallel loop metadata on memory reads also implies that
+ * if-conversion (i.e., speculative execution within a loop iteration)
+ * is safe. Given an instruction reading from memory,
+ * IsLoadUnconditionallySafe should return whether it is safe under
+ * (unconditional, unpredicated) speculative execution.
```
pjaaskel, line 475: Can you add a link to the LLVM bug here?

pjaaskel:
Also this fix is worth mentioning in CHANGES.

tests/regression/CMakeLists.txt
```diff
@@ -263,7 +265,8 @@ set_tests_properties("regression/setting_a_buffer_argument_to_NULL_causes_a_segf
   "regression/struct_kernel_arguments" "regression/vector_kernel_arguments"
   "regression/autolocals_in_constexprs" "regression/test_issue_231"
   "regression/test_issue_445" "regression/test_issue_553"
-  "regression/test_issue_577" "regression/test_flatten_barrier_subs"
+  "regression/test_issue_577" "regression/test_issue_757"
```
inducer, line 268: How hard would it be to make the test suite check that this is valgrind clean?

mattwala: I don't know the best way to do this. CTest has a mode in which tests can be run through memcheck. Maybe that could be done automatically, but I'm fluent enough in CMake to figure out how to do that. We'd have to think about how to handle extraneous errors because running the test binary through valgrind still gives a "Conditional jump or move depends on uninitialised value" in LLVM when compiling.

lib/llvmopencl/ParallelRegion.cc
mattwala: ac4468e6cf68c882ed16b995d44d86df08b45ec9

mattwala:
> Also this fix is worth mentioning in CHANGES.

fe199143598d9812cec2c699cead223ac36ff014

PR #869, opened 2020-09-16
Syncronization to previous events only in case of out-of-order queue

Kazhuu:
Using pocl with [PolyBench gramschmidt](https://github.com/cavazos-lab/PolyBench-ACC/blob/master/OpenCL/linear-algebra/solvers/gramschmidt/gramschmidt.cl) kernel. Execution time is very high compared to normal CPU execution. Pocl execution takes around 44 seconds compared to CPU which only takes less than a second. Reason is use of barriers in the command queue which are synchronized to all previous commands in the queue. This change only does the synchronization if the command queue is out-of-order queue. Otherwise it defaults to binding to the last event when queue is in-order queue.

I moved last event binding to `pocl_command_enqueue()` function because I thought that where it should be with other event synchronizations.

### DIFF
--- a/lib/CL/pocl_util.c
+++ b/lib/CL/pocl_util.c
@@ -510,18 +510,6 @@ cl_int pocl_create_command (_cl_command_node **cmd,
   (*cmd)->device = command_queue->device;
   (*cmd)->event->command = (*cmd);
 
-  /* in case of in-order queue, synchronize to previously enqueued command
-     if available */
-  if (!(command_queue->properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE))
-    {
-      POCL_LOCK_OBJ (command_queue);
-      if (command_queue->last_event.event)
-        {
-          pocl_create_event_sync ((*cmd)->event,
-                                  command_queue->last_event.event);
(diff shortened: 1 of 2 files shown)


pjaaskel:
Nice improvement! Can you add a note of this to CHANGES (with a mention of the speedup in this case) and yourself to CREDITS?

Kazhuu:
I was already in CREDITS so omitted that one.

lib/CL/pocl_util.c
```diff
@@ -546,8 +534,22 @@ void pocl_command_enqueue (cl_command_queue command_queue,
...
+     if available */
+  if (!(command_queue->properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE))
+    {
+      if (command_queue->last_event.event)
+        {
+          pocl_create_event_sync (node->event,
+                                  command_queue->last_event.event);
+        }
+    }
+  /* Command queue is out-of-order queue. If commend type is a barrier, then
```
franz, line 547: "commend" -> typo

Kazhuu: Fixed

PR #870, opened 2020-09-17
Fix event timing data for CUDA

matthiasdiener:
In contrast to `start` and `end`, `time_submit` is not based on a CUDA event
and sometimes represents a time stamp that is *after*`start` and `end`, leading to bogus event timings.

See e.g. https://github.com/illinois-ceesd/mirgecom/issues/71#issuecomment-678298984


matthiasdiener:
CC @inducer @isuruf - this might be worth adding to pocl-feedstock.

inducer:
@matthiasdiener Agree. Could you start a PR into https://github.com/conda-forge/pocl-feedstock/?

pjaaskel:
@isuruf @jrprice does it look OK to merge?

isuruf (approved):

PR #886, opened 2020-11-20
Vendor ID for 3.0

Oblomov:
OpenCL 3.0 has formalized the device vendor_id field, which should be the PCI vendor ID of the device if possible, and a value registered with Khronos (and automatically kept in sync cross-APIs) otherwise. This patchset tries to get the correct PCI vendor for both x86 and ARM CPUs based on the PCI root bus and/or the specific vendor, and prepares the ground (with a simple define that currently maps to the same value as before) for the Khronos registration as fall back.

### DIFF
--- a/lib/CL/devices/basic/basic.c
+++ b/lib/CL/devices/basic/basic.c
@@ -194,12 +194,11 @@ pocl_basic_init (unsigned j, cl_device_id device, const char* parameters)
 
   /* in case hwloc doesn't provide a PCI ID, let's generate
      a vendor id that hopefully is unique across vendors. */
-  const char *magic = "pocl";
+  /* TODO this should be replaced with the appropriate
+   * CL_KHRONOS_VENDOR_ID_POCL when it's registered */
+#define POCL_OPENCL_VENDOR_ID 0x6c636f70 /* pocl */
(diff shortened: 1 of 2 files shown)


pjaaskel (commented):
Thanks for this. Can you check what the clang-format script says about this change? To me there seem to be some suspicious lines.

Also, I'm not sure should pocl apply for the vendor id -- who is the vendor then? Which organization? I'll ask Khronos.

Oblomov:
Sorry for not checking the patches first, I'll give clang-format-script a run.

For the vendor ID I don't know. Mesa has one too, so I assume FLOSS projects can have one too 8-)

Oblomov:
The format checks end up requiring some whitespace changes even in places where I just put stuff in the middle —should I proceed and accept the changes anyway?

Oblomov:
I think this fixed the obvious whitespace issues, but I think the formatter still needs some tuning ;-)

pjaaskel:
Yes, it the script is new and would use tuning from someone who knows the quirks of clang-format. https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkVendorId.html I see Mesa there, yes, I'll ask Khronos how we could get pocl there too.

PR #891, opened 2020-12-13
Double event wait

Oblomov:
This adds a test (and a fix for a failure in the CUDA driver) for when events are are waited-upon multiple times.

PR #892, opened 2020-12-13
Conditional event free test

Oblomov:
This changes test_event_free so that it doesn't run the image-related tests when the device doesn't support it (e.g. CUDA) and consequently labels the test as valid for the CUDA device.

It's built on top of #891 but it doesn't actually depend on it. If necessary I can rebase (which needs to be done manually because the changes in `CMakeList.txt` are right next to each other).


franz:
Rebased this on `release_1_6` branch and pushed.

PR #910, opened 2021-01-25
pocl-cuda: implement support for cl_nv_device_attribute_query

matthiasdiener:
WIP - looking for feedback.


znmeb:
I can test this on Jetson Nano and AGX Xavier - will the CUDA tests show anything? Will `clinfo`?

matthiasdiener:
> I can test this on Jetson Nano and AGX Xavier - will the CUDA tests show anything? Will `clinfo`?

Thank you! I'm not sure about the CUDA tests, but `clinfo` should show something. 

You could also try code like this (`0x4008 == CL_DEVICE_PCI_BUS_ID_NV`):

```
clGetDeviceInfo(devices, 0x4008, 8, &buf_uint, NULL)
```

matthiasdiener:
here is a diff of the output of `clinfo` between pocl-1.6 and this branch:

```patch 
--- pocl-1.6
+++ this branch
@@ -1,16 +1,18 @@
   Device Name                                     Tesla V100-SXM2-16GB
   Device Vendor                                   NVIDIA Corporation
   Device Vendor ID                                0x10de
   Device Version                                  OpenCL 1.2 pocl HSTR: CUDA-sm_70
-  Driver Version                                  1.6
+  Driver Version                                  1.7-pre cl_nv_device_attribute_query-ext-0-g37e699e5
   Device OpenCL C Version                         OpenCL C 1.2 pocl
   Device Type                                     GPU
+  Device Topology (NV)                            PCI-E, 03:00.0
   Device Profile                                  FULL_PROFILE
   Device Available                                Yes
   Compiler Available                              Yes
   Linker Available                                Yes
   Max compute units                               80
   Max clock frequency                             1530MHz
+  Compute Capability (NV)                         7.0
   Device Partition                                (core)
     Max number of sub-devices                     1
     Supported partition types                     None
@@ -19,6 +21,7 @@
   Max work item sizes                             1024x1024x64
   Max work group size                             1024
   Preferred work group size multiple              32
+  Warp size (NV)                                  32
   Preferred / native vector sizes
     char                                                 1 / 1
     short                                                1 / 1
@@ -50,12 +53,14 @@
   Error Correction support                        Yes
   Max memory allocation                           4236378112 (3.945GiB)
   Unified memory for Host and Device              No
+  Integrated memory (NV)                          No
   Minimum alignment for any data type             128 bytes
   Alignment of base address                       4096 bits (512 bytes)
   Global Memory cache type                        None
   Image support                                   No
   Local memory type                               Local
   Local memory size                               49152 (48KiB)
+  Registers per block (NV)                        65536
   Max number of constant args                     8
   Max constant buffer size                        65536 (64KiB)
   Max size of kernel argument                     1024
@@ -67,6 +72,10 @@
   Execution capabilities
     Run OpenCL kernels                            Yes
     Run native kernels                            No
+    Kernel execution timeout (NV)                 No
+  Concurrent copy and kernel execution (NV)       Yes
+    Number of async copy engines                  4
   printf() buffer size                            16777216 (16MiB)
   Built-in kernels                                (n/a)
-  Device Extensions                               cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_fp64 cl_khr_int64_base_atomics cl_khr_int64_extended_atomics
+  Device Extensions                               cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_fp64 cl_khr_int64_base_atomics cl_khr_int64_extended_atomics cl_nv_device_attribute_query
```

pjaaskel (changes requested):
LGTM. Can you add the feature to CHANGES?

lib/CL/clGetDeviceInfo.c
```diff
@@ -194,7 +196,29 @@ POname(clGetDeviceInfo)(cl_device_id   device,
...
+  case CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV       :
+  case CL_DEVICE_REGISTERS_PER_BLOCK_NV            :
+  case CL_DEVICE_WARP_SIZE_NV                      :
+  case CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV   :
+  case CL_DEVICE_PCI_BUS_ID_NV                     :
+  case CL_DEVICE_PCI_SLOT_ID_NV                    :
+  case CL_DEVICE_PCI_DOMAIN_ID_NV                  :
+    {
+      int res;
+      res = pocl_cuda_handle_cl_nv_device_attribute_query(device, param_name);
```
pjaaskel, line 210: This calls CUDA driver functions unconditionally. Doesn't this break if the CUDA driver is not enabled&built? Did you test without CUDA?

pjaaskel (changes requested):

lib/CL/clGetDeviceInfo.c
matthiasdiener: I tested with and without CUDA, and it seems to have worked like I would expect. I have made the extension conditional on `BUILD_CUDA` to make it a bit cleaner though.

matthiasdiener: `clinfo -a/-A` appear to work fine with and without CUDA.

pjaaskel:
Thanks. 

Now that I look at this again, I think it would be cleaner if the vendor-specific device attribute handling was delegated to the driver layer so that the drivers can handle driver-specific extensions internally without needing to add driver-specific calls or attributes to the generic clGetDeviceInfo() implementation. That is, before returning an error for an unknown device info in clGetDeviceInfo() it would call a function in the driver layer which can be then overridden in the cuda side to detect the cases (and would be NULL by default for the drivers). Can you do this type of refactoring to make it cleaner and easier in the future to add device/vendor-specific device infos? 

Also I tested this yesterday and PyOpenCL started to fail. I didn't have a chance to debug further, but can you try with the PyOpenCL enabled and check if you can see it?

matthiasdiener:
> Now that I look at this again, I think it would be cleaner if the vendor-specific device attribute handling was delegated to the driver layer so that the drivers can handle driver-specific extensions internally without needing to add driver-specific calls or attributes to the generic clGetDeviceInfo() implementation. That is, before returning an error for an unknown device info in clGetDeviceInfo() it would call a function in the driver layer which can be then overridden in the cuda side to detect the cases (and would be NULL by default for the drivers). Can you do this type of refactoring to make it cleaner and easier in the future to add device/vendor-specific device infos?

Ok, should this be a part of `pocl_device_ops` (maybe calling it `ops->attribute_query`)?

> 
> Also I tested this yesterday and PyOpenCL started to fail. I didn't have a chance to debug further, but can you try with the PyOpenCL enabled and check if you can see it?

Hmm, I tested this with pyopencl and it seemed to work fine, but I'll check again.

pjaaskel:
> Ok, should this be a part of `pocl_device_ops` (maybe calling it `ops->attribute_query`)?

Yes, in the device ops, but make it more generic so it can be used to handle any vendor-specific clGetDeviceInfo() query?
E.g. get_device_info_ext() or such?

> Hmm, I tested this with pyopencl and it seemed to work fine, but I'll check again.

Did you check with the PoCL's pyopencl test suite?

matthiasdiener:
> > Ok, should this be a part of `pocl_device_ops` (maybe calling it `ops->attribute_query`)?
> 
> Yes, in the device ops, but make it more generic so it can be used to handle any vendor-specific clGetDeviceInfo() query?
> E.g. get_device_info_ext() or such?

I refactored this in the most recent commit like you suggested.


> > Hmm, I tested this with pyopencl and it seemed to work fine, but I'll check again.
> 
> Did you check with the PoCL's pyopencl test suite?

I ran the pyopencl test suite as outlined [here](https://github.com/pocl/pocl/blob/master/examples/PyOpenCL/README) with the current master branch of pyopencl:

``` 
$ PYOPENCL_NO_CACHE=1 PYOPENCL_TEST=portable py.test -v --tb=native
========================================== test session starts ==========================================
platform linux -- Python 3.8.6, pytest-6.2.1, py-1.10.0, pluggy-0.13.1 -- /g/g91/diener3/Work/emirge/miniforge3/envs/ceesd/bin/python
cachedir: .pytest_cache
rootdir: /usr/WS1/diener3/Work/emirge/pyopencl, configfile: pytest.ini
plugins: pudb-0.7.0
collected 1147 items

[...]

================ 1011 passed, 42 skipped, 94 xfailed, 114 warnings in 1473.27s (0:24:33) ================
```

PR #919, opened 2021-02-05
support building with -fvisibility=hidden -fvisibility-inlines-hidden

anbe42:
Here I have a patch series to finally allow building with `-fvisibility=hidden -fvisibility-inlines-hidden` to reduce the amount of (mostly c++) symbols exposed by the shared libraries. It starts with some fixes I've long been carrying in the Debian packages and some further cleanup (e.g. pocl symbols should be prefixed with `pocl` and tests should depend on `pocl_version_check`).
Please merge or cherry-pick as much as you like and we can continue working on the remaining parts.

I've only tested it with the CPU devices (`basic` and `pthread`) and only with building an ICD and testing with pocl's internal testsuite. There may be adjustments needed for other build configurations, e.g. more declarations needing a `POCL_EXPORT` decoration. (I've tried to keep the `pocl_*` symbols exposed minimal.)
In order to quickly check for missing symbols I've added some `dlopen` tests.

I've intentionally not reformatted the commit adding the `POCL_EXPORT` decorations. Please advise how you'd like them.

Is there a reason for `dlopen()`ing the device library with `RTLD_LAZY` instead of `RTLD_NOW`? (it would indicate missing symbols immediately.)

What is currently unclear to me, is how `libllvmopencl.so` is to be used. It isn't used by pocl itself or any of the tests unless I missed something. If I simply `dlopen()` it, I get a missing symbol (or more), that is available in `libpocl` and if the symbols from`libpocl` are loaded, `dlopen()` fails with
```
Two passes with the same argument (-allocastoentry) attempted to be registered!
Segmentation fault
```
(That was with llvm-9, I haven't tried other versions.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,12 @@ option(POCL_ICD_ABSOLUTE_PATH "Use absolute path in pocl.icd" ON)
 
 option(ENABLE_POCL_BUILDING "When OFF, env var POCL_BUILDING has no effect. Defaults to ON" ON)
 
+option(VISIBILITY_HIDDEN "Build with -fvisibility=hidden -fvisibility-inlines-hidden" OFF)
+if(VISIBILITY_HIDDEN)
+  add_compile_options(-fvisibility=hidden)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fvisibility-inlines-hidden>)
+endif()
+
 #### these are mostly useful for pocl developers
 
 option(DEVELOPER_MODE "This will SIGNIFICANTLY slow down pocl (but speed up its compilation). Only turn on if you know what you're doing." OFF)
--- a/cmake/LLVM.cmake
+++ b/cmake/LLVM.cmake
@@ -243,8 +243,8 @@ string(STRIP "${LLVM_SYSLIBS}" LLVM_SYSLIBS)
 ####################################################################
 
 # llvm-config does not include clang libs
-if((9 LESS LLVM_MAJOR) AND (NOT STATIC_LLVM))
-  # For Clang 10+, link against a single shared library instead of multiple component shared
+if((8 LESS LLVM_MAJOR) AND (NOT STATIC_LLVM))
+  # For Clang 9+, link against a single shared library instead of multiple component shared
   # libraries.
   if("${LLVM_LIBNAMES}" MATCHES "LLVMTCE")
     set(CLANG_LIBNAMES clangTCE-cpp)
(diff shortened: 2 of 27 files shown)


CMakeLists.txt
```diff
@@ -179,6 +179,12 @@ option(POCL_ICD_ABSOLUTE_PATH "Use absolute path in pocl.icd" ON)
 
 option(ENABLE_POCL_BUILDING "When OFF, env var POCL_BUILDING has no effect. Defaults to ON" ON)
 
+option(VISIBILITY_HIDDEN "Build with -fvisibility=hidden -fvisibility-inlines-hidden" OFF)
```
pjaaskel, line 182: Is there a reason to not build with this mode? I prefer to avoid additional build options (especially default OFF ones) since they tend to bitrot easily.

pjaaskel (commented):
Thanks! Outside that question I think this (incl. the POCL_EXPORT indentation) looks good to me. I'll proceed to testing meanwhile.

pjaaskel:
Also please update CREDITS. I'm testing this, if it passes my local envs with the build mode OFF, I'll push with the current OFF by default and next I'll test ON by default.

CMakeLists.txt
anbe42: If it works with all configurations it would be nice to enable it by default and even drop the switch ;-)

pjaaskel: I pushed it but now that I tried to actually enable this, I'm getting 
```libpocl-devices-cuda.so:
  undefined symbol: pocl_get_string_option```
with my CUDA environment.

anbe42: As I wrote, so far I only tested it for the cpu devices and made the symbol set minimal for them.

Can you paste `nm -D lib/CL/devices/XXX/libpocl-devices-XXX.so` for the non-CPU devices from a "classic" build and I'll see what symbols are missing an `POCL_EXPORT`?

anbe42:
> Also please update CREDITS.

I'm already in ;-)

pjaaskel:
> > Also please update CREDITS.
> 
> I'm already in ;-)

Sorry meant CHANGES of course :)

CMakeLists.txt
pjaaskel: https://termbin.com/hju9

anbe42: A new commit is available in my `visibility-hidden` branch that should fix that for the `cuda` device, but we need to check the `accel`, `hsa` and `tce` devices, too.
Any hints on how `libllvmopencl.so` is to be used?

pjaaskel: I do not think libllvmopencl.so is used as a separate library any longer. We used to call it from opt command line but it was long time ago, and I believe this is not done anymore anywhere. Could you just try removing the build of the .so and see if any of the tests explode?

anbe42: Disabling `libllvmopencl.so` did not show any bad effects, the corresponding commit is on top of my `visibility-hidden` branch.
The last usage reference I could find was `scripts/pocl-standalone.in` which is gone since 0.14.

PR #921, opened 2021-02-11
Avoid division by 0 in profiling output if total_commands is 0

### LINKED ISSUE #909: Zero division in pocl_cq_profiling.c L107
https://github.com/pocl/pocl/blob/37e699e59cc43ed38597593bd9828e2987a4a267/lib/CL/pocl_cq_profiling.c#L107

Got a zero division when no kernel is actually executed, and total_commands = 0.

Oblomov:
This is achieved by adding !total_commands to the divisor, which is a
quick way to divide by 1 if total_commands is 0, and by total_commands
itself if total_command is non-zero.

This should fix #909

### DIFF
--- a/lib/CL/pocl_cq_profiling.c
+++ b/lib/CL/pocl_cq_profiling.c
@@ -103,8 +103,9 @@ pocl_atexit ()
   printf ("     %-30s %10s %15s %3s %10s\n", "",
           "==========", "==========", "====", "==========");
 
+  /* Add !total_commands to avoid a division by 0 if total_commands is 0 */
   printf ("     %-30s %10lu %15lu %4s %10lu\n", "", total_commands, total_time,
-          "100%", total_time / total_commands);
+          "100%", total_time / (total_commands + !total_commands) );
 
   /* TODO: Critical path information of the task graph. */
 }


pjaaskel:
I like the nifty C trickery, but this raises the question: why print at all if the total_commands is zero? Early exit from the function instead?

Oblomov:
I would argue that if the user asked for tracing output, they should get tracing output, even if it's “mostly empty”, but of course if we prefer an early exit I can implement that instead.

Kazhuu:
I think explicit trace output is also better if user asked for it than printing nothing. But of course the final decision is not mine to make :)

pjaaskel:
Then print "No commands executed." in that case? If I understood correctly that's the case. Printing nonsensical numbers (yet, but without crashing) seems the strange choice here.

Oblomov:
This is the output:
```
     kernel                           launches        total us          avg us
                                    ==========      ========== ==== ==========
                                             0               0 100%          0
```
I wouldn't call the number “nonsensical”: they are exactly what we would expect.

pjaaskel:
Fair enough :)

PR #939, opened 2021-04-27
Avoid inheritance from std::vector in ParallelRegion class.

rusyaev-roman:
Use composition instead of inheritance.

### DIFF
--- a/lib/llvmopencl/ParallelRegion.cc
+++ b/lib/llvmopencl/ParallelRegion.cc
@@ -55,8 +55,7 @@ using namespace pocl;
 
 int ParallelRegion::idGen = 0;
 
-ParallelRegion::ParallelRegion(int forcedRegionId) : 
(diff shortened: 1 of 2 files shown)


pjaaskel:
Did you try running tier1 tests with this? Pulling this in seems to cause a segfault in a PyOpenCL test.

rusyaev-roman:
Oh, sorry. I'll double check this

pjaaskel (changes requested):
Test issue (might be transitional).

pjaaskel:
Hi, can you update this PR? I'll then retest and pull it in if my local test envs pass. Thanks!

rusyaev-roman:
> Did you try running tier1 tests with this? Pulling this in seems to cause a segfault in a PyOpenCL test.

Could you check it again? I've run it manually and it passed successfully.

PR #948, opened 2021-06-10
cpuinfo: check for fopen() failure when opening sysfs nodes

niacat:
If sysfs is not mounted (or not available, e.g. on BSD), fopen() will return a NULL pointer and this will be passed to fscanf, potentially resulting in NULL dereference. This fixes a segmentation fault running poclcc -l on NetBSD.


pjaaskel (changes requested):
Can you add this bug fix to CHANGES and yourself to CREDITS (if not there already)? Thanks!

PR #949, opened 2021-06-10
pocl_timing: fix detection of Unix monotonic clocks

niacat:
CLOCK_UPTIME_FAST is an extension only available on FreeBSD and DragonFlyBSD, so don't try to use it on other BSDs - this previously resulted a compilation failure on NetBSD.

Instead, use CLOCK_MONOTONIC. This is available on NetBSD and OpenBSD and is defined in POSIX.2008. On older Unix-like operating systems, keep the fallback to CLOCK_REALTIME.

Replace OS-specific checks with more portable checks for definitions.


pjaaskel (changes requested):
Same here, please add to CHANGES as it fixes a previously little-supported platform.

PR #950, opened 2021-06-10
Fix detection of dlopen()/libdl

niacat:
- Many Unix systems do not need (macOS, Linux with musl?), or do  not have (most BSDs) libdl, and dlopen() is in libc instead.  Therefore, libdl being missing should not be fatal.   Replace checks for libdl.so with checks for dlfcn.h.

- Remove #error checks for libdl from indivial C files. The CMake build system will already refuse to run if dlopen()  is missing, so these checks are redundant.

Helps with building on NetBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -778,7 +778,7 @@ include_directories("include")
 
 ######################################################################################
 
-set(HAVE_LIBDL OFF CACHE BOOL "dlopen" FORCE)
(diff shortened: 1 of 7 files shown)


pjaaskel (changes requested):
...and here. Perhaps a single grouped entry in CHANGES for the NetBSD-related fixes should be the most informative addition.

PR #951, opened 2021-06-12
Add parallel loop metadata on LLVM 13 onward for load instructions inside conditions

Kazhuu:
Now that parallel loop metadata if-conversion safety is removed by [this](https://reviews.llvm.org/D103907) change for LLVM 13 branch. This change will enable adding metadata insertion on load instructions unconditionally. Prior to LLVM 13 will still use conditional placement of metadata to avoid miscompilations. This change is a follow-up to #850.

I tested this locally and with the fix using latest LLVM from trunk is not producing illegal code anymore. Also earlier vectorization failures from PolyBench kernels [this](https://github.com/cavazos-lab/PolyBench-ACC/blob/master/OpenCL/stencils/convolution-3d/3DConvolution.cl) and [this](https://github.com/cavazos-lab/PolyBench-ACC/blob/master/OpenCL/stencils/convolution-2d/2DConvolution.cl) are now vectorizing successfully and executing faster.

### DIFF
--- a/lib/llvmopencl/ParallelRegion.cc
+++ b/lib/llvmopencl/ParallelRegion.cc
@@ -468,12 +468,23 @@ ParallelRegion::Verify()
  *     !1 distinct !{}
  *     !2 distinct !{}
  *
- * Parallel loop metadata on memory reads also implies that
+ * Parallel loop metadata prior to LLVM 13 on memory reads also implies that
  * if-conversion (i.e., speculative execution within a loop iteration)
  * is safe. Given an instruction reading from memory,
  * IsLoadUnconditionallySafe should return whether it is safe under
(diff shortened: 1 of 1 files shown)


fodinabor:
I scheduled the patch to be backported to 12.0.1 as well, so the check could possibly be relaxed to support that as well in a week or so when the patch is in...

Kazhuu:
Okay that sounds great! Maybe add the reference here once you're ready with the patch?

fodinabor:
The LLVM patch is in the release/12.x branch now and should thus land in >=12.0.1-rc3
https://github.com/llvm/llvm-project/commit/b690ec54817d9f4ad0e6e16f88a12044660b794f

PR #954, opened 2021-06-16
Improve generic vload vstore functions

hominhquan:
- Current implementation splits a vload4 into two vload2 for example,
preventing compiler from performing load/store packing.
- This commit improve it by doing load/store packing explicitly
(vload3/vstore3 are not concerned).

### DIFF
--- a/lib/kernel/vload.cl
+++ b/lib/kernel/vload.cl
@@ -31,7 +31,8 @@
   TYPE##2 _CL_OVERLOADABLE                                              \
   vload2(size_t offset, const MOD TYPE *p)                              \
   {                                                                     \
-    return (TYPE##2)(p[offset*2], p[offset*2+1]);                       \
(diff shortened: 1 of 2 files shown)


pjaaskel (changes requested):
Is this legal? Given that the vector datatypes are vector-aligned and vloadN/vstoreN are for element-aligned accessess, you are basically performing an illegal pointer cast? Thus it might work by luck depending if e.g. an optimization uses the full alignment property or not. And if your target's SIMD loads/stores support element-aligned or not?

Oblomov:
Yeah, I think this is illegal. The whole point of these functions is to support unaligned packed loading. At the very least the optimization should be based on hardware properties (for example, intel has movups etc instructions that can be used for unaligned loads, and vload should map to those when possible, whereas direct accesses should map to movaps etc)

PR #961, opened 2021-06-22
[WIP] OpenCL 3.0 API conformance: Device-Side Enqueue

Vinsteri:
WIP 
- calling clSetDefaultCommandQueue() causes segfault. 
- Missing a query in clGetEventProfilingInfo.

This pull request implements missing queries and API functions related to device-side enqueue. Device-side enqueue is specified as optional in OpenCL 3.0. PoCL does not support device-side enqueue. This pull request implements the OpenCL API such that it conforms to the specification when device-side enqueue is not supported.
See https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_device_side_enqueue


lib/CL/clGetPlatformIDs.c
```diff
@@ -37,115 +37,115 @@
 */
 #ifdef BUILD_ICD
 struct _cl_icd_dispatch pocl_dispatch = {
-  &POname(clGetPlatformIDs),
```
pjaaskel, line 40: Can you revert this large style change for now in this commit? Is it according to the C formatter?
Let's avoid large white-space/style-only commits but fix the style gradually with other changes. So, just submit your additional functionality, not the large style fix, thanks.

tests/runtime/test_deviceside_enqueue.c
```diff
@@ -0,0 +1,139 @@
...
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+   IN THE SOFTWARE.
+*/
+
+#include "poclu.h"
+#include <CL/cl.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+/*
```
pjaaskel, line 29: Check the formatting of this comment.

tests/runtime/test_deviceside_enqueue.c
```diff
@@ -0,0 +1,139 @@
...
+
+      for (j = 0; j < ndevices; j++)
+        {
+          cl_uint device_queue_support;
+          err = clGetDeviceInfo (devices[j], CL_DEVICE_DEVICE_ENQUEUE_SUPPORT,
+                                 sizeof (cl_uint), &device_queue_support,
+                                 NULL);
+          CHECK_OPENCL_ERROR_IN ("clGetDeviceInfo");
+
+          TEST_ASSERT (device_queue_support == 0);
```
pjaaskel, line 71: It's wrong to assume none of the devices have the support here. This tests breaks if we add support for some of the upstream devices in the future, and breaks for out-of-tree devices which might have the support.

pjaaskel (changes requested):

lib/CL/clGetPlatformIDs.c
Vinsteri: The white space was added by the format-last-commit.sh script. I reverted the style changes.

tests/runtime/test_deviceside_enqueue.c
Vinsteri: Changed the tests to be skipped if clGetDeviceInfo() query CL_DEVICE_DEVICE_ENQUEUE_SUPPORT returns something else than 0.
I had not considered the possibility of out-of-tree devices existing that might support this feature.

pjaaskel (changes requested):
LGTM. Can you still update CREDITS (if you are not there already) and CHANGES?

PR #965, opened 2021-07-08
Add parallel loop metadata on LLVM 12 onward and not before on load instructions inside conditions

Kazhuu:
Follow-up PR for this #951. This time lower the LLVM version check to 12 because the fix got backported to LLVM 12 release branch llvm/llvm-project@b690ec5.

### DIFF
--- a/lib/llvmopencl/ParallelRegion.cc
+++ b/lib/llvmopencl/ParallelRegion.cc
@@ -468,23 +468,25 @@ ParallelRegion::Verify()
  *     !1 distinct !{}
  *     !2 distinct !{}
  *
- * Parallel loop metadata prior to LLVM 13 on memory reads also implies that
+ * Parallel loop metadata prior to LLVM 12 on memory reads also implies that
  * if-conversion (i.e., speculative execution within a loop iteration)
  * is safe. Given an instruction reading from memory,
  * IsLoadUnconditionallySafe should return whether it is safe under
  * (unconditional, unpredicated) speculative execution.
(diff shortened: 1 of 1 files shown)


fodinabor:
Mh.. I don't know how wide-spread 12.0.0 will be, but the issue is only fixed with 12.0.1, thus I used `#if LLVM_VERSION_MAJOR < 13 && !(LLVM_VERSION_MAJOR == 12 && LLVM_VERSION_MINOR >= 0 && LLVM_VERSION_PATCH >= 1)` for this check.

Kazhuu:
Thank you for this. Yeah this makes more sense. I'll add then change.

lib/llvmopencl/ParallelRegion.cc
```diff
@@ -498,9 +499,9 @@ ParallelRegion::AddParallelLoopMetadata(
         continue;
       }
 
-#ifdef LLVM_OLDER_THAN_13_0
+#if LLVM_VERSION_MAJOR < 13 && !(LLVM_VERSION_MAJOR == 12 && LLVM_VERSION_MINOR >= 0 && LLVM_VERSION_PATCH >= 1)
```
pjaaskel, line 502: Please mind the row length.

pjaaskel (changes requested):

lib/llvmopencl/ParallelRegion.cc
Kazhuu: Good point. Used provided formatter to format this line.

pjaaskel (approved):

PR #973, opened 2021-08-11
topology: setenv is linux* specific

xantares:


pjaaskel (approved):
Can you add notes about these MingGW fixes to CHANGES and yourself to CREDITS? Do these make all basic tests pass in that platform?

xantares:
the tests do not pass yet, something is up with the loader

PR #975, opened 2021-08-11
cmake: Simplify dlfcn detection for mingw

xantares:


CMakeLists.txt
```diff
@@ -844,39 +844,13 @@ include_directories("include")
...
-    else()
-      message(STATUS "libdl not found, assuming dlopen() is in libc")
-      set(DL_LIB "")
-    endif()
-
-    if(DL_H)
-      get_filename_component(DL_H_INCLUDE_DIR "${DL_H}" DIRECTORY)
-      string(FIND "${CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES}" "${DL_H_INCLUDE_DIR}" LTPOSITION)
-      # include the directory of dlfcn.h, if its not in the default system include dirs
-      # also when cross-compiling this includes <cross-compile-root>/usr/include, which screws things up
```
pjaaskel, line 868: Again worried about cross compiled targets if you remove this code path.

pjaaskel (changes requested):

CMakeLists.txt
xantares: when cross-compiling cmake should look in the target prefix and not the native prefix, os I dont understand this logic, and if we want to change that the variable CMAKE_FIND_ROOT_PATH_MODE_INCLUDE can be tweaked to get the desired bahavior so I say this logic is useless

pjaaskel: OK. Unfortunately xcompile is not regularly tested nor tested automatically, so I trust you on this.

PR #980, opened 2021-08-19
Add barrier to latch if all preds are barriers.

### LINKED ISSUE #553: More barrier fun (conditionals and loops)
I have another misbehaving kernel:

```
// Expected output:
//  outer=A inner=B
//  + outer=A inner=B
// for each value of A and B.
// However I see three copies of the second line (starting with +).
// Commenting out any one line marked with YYYY bring it down to two copies,
// and commenting out any one line marked with XXXX gives the expected output.


__kernel void pocltest(int xarg1, int xarg2) {
    int outerend = 1;
    int innerend = 1;
    outerend = 2; // YYYY
    innerend = 2; // YYYY
    int outer = 0;
    int inner = 0;
    int arg1 = 1;
    int arg2 = 1;
    arg1 = xarg1; // XXXX
    arg2 = xarg2; // XXXX
    for (outer = 0; outer < outerend; outer++) // XXXX
    {
	for(inner = 0; inner < innerend; inner++) // XXXX
	{
	    printf("outer=%d inner=%d\n", outer, inner);
	    if (arg2 > arg1) // XXXX
	    {
		barrier(CLK_LOCAL_MEM_FENCE); // XXXX
	    }
	    if (arg1 > 0) // XXXX
	    {
		barrier(CLK_LOCAL_MEM_FENCE); // XXXX
	    }
	    printf("+ outer=%d inner=%d\n", outer, inner);
	}
    }

    if (arg1 > 0) // XXXX
    {
	barrier(CLK_LOCAL_MEM_FENCE); // XXXX
    }
}
```

This is cut down from a version that does actual computation, of course.  The code is two nested loops enclosing two conditionals, and followed by another conditional.  All the conditionals need to be evaluated at runtime because they use parameters sent into the kernel.  The output, using pocl trunk with either LLVM 5.0.0 or LLVM trunk, with a workgroup size of 1, and arguments arg1 and arg2 set to some value greater than zero, is:

```
outer=0 inner=0
+ outer=0 inner=0
+ outer=0 inner=0
+ outer=0 inner=0
outer=0 inner=1
+ outer=0 inner=1
+ outer=0 inner=1
+ outer=0 inner=1
outer=1 inner=0
+ outer=1 inner=0
+ outer=1 inner=0
+ outer=1 inner=0
outer=1 inner=1
+ outer=1 inner=1
+ outer=1 inner=1
+ outer=1 inner=1
```

The printf before the barriers in the inner loop is printed once as expected, but the printf after the barriers is printed three times.  Commenting out any line with XXXX, which means:
- commenting out any one of the barriers, or
- removing any one of the surrounding conditionals/ifs, or
- making any conditional resolvable at compile-time (i.e. replacing arg1 or arg2 with a constant)

fixes the issue.  Also commenting out any line with YYYY, which replaces the stop value of either the inner or outer loop from 2 to 1 (i.e. making it a singleton loop) makes it print only twice (but still incorrectly).

Also a LLVM bug?  And please let me know if these bugs should be directed somewhere else. Thanks again!

fodinabor:
After BTR we need to re-add barriers in the single latch for the parallel region formation to work correctly. This is done in CanonicalizeBarriers.

This fixes #553.


pjaaskel:
Nice! I'm extremely happy to see someone else is brave enough to dig into the CFG SPMD passes :D Can you add a test case for this? There are unit tests for WG function generation issues you can use as a starting point.

pjaaskel:
regression/case_with_multiple_variable_length_loops_and_a_barrier_in_one fails with this patch.

fodinabor:
Heh, I sort of had to :)
I'm currently taking a look at efficiently implementing nd_range kernels on CPU in hipSYCL. For this, I pulled in (most of) the POCL pipeline, which is great work btw! This triggered a test on my end, so I fixed it for hipSYCL a while ago and wanted to contribute it back before I completely forget the circumstances.
As it turned out that was a rather big hammer fix that worked fine in hipSYCL, I investigated the POCL side a bit more and came up with another slightly smaller fix, also including a testcase now.

pjaaskel:
I see. Thanks - testing now. It would be nice to get the de-SPMD CFG transformations upstreamed to LLVM where they belong.

PR #1003, opened 2021-11-18
OpenCL 3.0 API conformance: Subgroups

Vinsteri:
Changes are based on the OpenCL 3.0 API specification. See
https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_subgroups

CHANGES
- Two new queries to clGetDeviceInfo ( CL_DEVICE_MAX_NUM_SUB_GROUPS and CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS ) and the corresponding fields to the cl_device_id struct. These fields get initialized in the devices/common.c pocl_init_default_device_infos() function to the values specified when subgroups are not supported.
- Minimal implementation of clGetKernelSubGroupInfo with only the necessary error handling required for OpenCL 3.0 conformance.


lib/CL/clGetKernelSubGroupInfo.c
```diff
@@ -0,0 +1,68 @@
...
+                            "could not find the "
+                            "device supplied in argument\n");
+    }
+  else
+    {
+      POCL_RETURN_ERROR_ON ((kernel->context->num_devices > 1),
+                            CL_INVALID_DEVICE,
+                            "No device given and context has > 1 device\n");
+      device = kernel->context->devices[0];
+    }
```
Vinsteri, line 55: Possible DRY violation. This logic is shared with ( and directly copied from ) clGetKernelWorkGroupInfo().

pjaaskel: Can you refactor it nicely to a helper function?

Issue #1009, opened 2021-11-25
pocl_pthread_driver_thread: Assertion `td->printf_buffer != NULL' failed.


anbe42:
On certain memory constrained systems pocl will fail to initialize the pthread device due to out-of-memory conditions. This was reported on 32-bit arm systems running automated tests using pocl as OpenCL backend for other applications. No detailed information about available cores or memory is known. But I've been able to reproduce this on an x86_64 system with 16 cores (incl. hyperthreading) by limiting the virtual memory with `ulimit -v 1100000` and running `clinfo`. This frequently fails, while with a memory limit of `ulimit -v 1250000` `clinfo` always passes. (You may need different parameters to reproduce this on your local machine.)

```
$ ulimit -v 1100000
$ clinfo
Number of platforms                               1
  Platform Name                                   Portable Computing Language
  Platform Vendor                                 The pocl project
  Platform Version                                OpenCL 2.0 pocl 1.8  Linux, None+Asserts, RELOC, LLVM 9.0.1, SLEEF, DISTRO, POCL_DEBUG
  Platform Profile                                FULL_PROFILE
  Platform Extensions                             cl_khr_icd cl_pocl_content_size
  Platform Extensions function suffix             POCL

  Platform Name                                   Portable Computing Language
Number of devices                                 1
  Device Name                                     pthread-Intel(R) Core(TM) i7-9800X CPU @ 3.80GHz
  Device Vendor                                   GenuineIntel
  Device Vendor ID                                0x8086
  Device Version                                  OpenCL 1.2 pocl HSTR: pthread-x86_64-pc-linux-gnu-skylake-avx512
  Driver Version                                  1.8
  Device OpenCL C Version                         OpenCL C 1.2 pocl
  Device Type                                     CPU
  Device Profile                                  FULL_PROFILE
  Device Available                                Yes
  Compiler Available                              Yes
  Linker Available                                Yes
  Max compute units                               16
  Max clock frequency                             4500MHz
clinfo: ./lib/CL/devices/pthread/pthread_scheduler.c:512: pocl_pthread_driver_thread: Assertion `td->printf_buffer != NULL' failed.
clinfo: ./lib/CL/devices/pthread/pthread_scheduler.c:512: pocl_pthread_driver_thread: Assertion `td->printf_buffer != NULL' failed.
Aborted
```

This happens in `lib/CL/devices/pthread/pthread_scheduler.c`
```
[...]

PR #1010, opened 2021-11-26
Pthread fixes

anbe42:
some fixes for the pthread device, mainly:
* `pthread_scheduler_init()`: `abort()` if `pthread_create()` fails (also if it fails with `EAGAIN`)
  * recovery would be hard: terminate the partially created thread pool and return an error code
* if any worker thread experienced an allocation failure during initialization, don't assert, but report this back to the scheduler: uninitialize the scheduler (shut down the worker thread pool) in this case and return an error code from  `pthread_scheduler_init()`

this may not totally solve the original issue, but at least in case of only the allocations failing the application would have the possibility to handle the unavailable device

there are more `pthread_create()` and other `pthread_*()` calls missing error handling in the pocl codebase, these are not covered here

fixes: #1009

### DIFF
--- a/lib/CL/devices/pthread/pthread_scheduler.c
+++ b/lib/CL/devices/pthread/pthread_scheduler.c
@@ -77,6 +77,9 @@ typedef struct scheduler_data_
   POCL_FAST_LOCK_T wq_lock_fast __attribute__ ((aligned (HOST_CPU_CACHELINE_SIZE)));
 
   int thread_pool_shutdown_requested;
+
+  pthread_barrier_t init_barrier
+      __attribute__ ((aligned (HOST_CPU_CACHELINE_SIZE)));
 } scheduler_data __attribute__ ((aligned (HOST_CPU_CACHELINE_SIZE)));
 
 static scheduler_data scheduler;
@@ -106,6 +109,9 @@ pthread_scheduler_init (cl_device_id device)
    * TODO fix this */
   scheduler.local_mem_size = device->local_mem_size + device->max_parameter_size * MAX_EXTENDED_ALIGNMENT;
 
+  PTHREAD_CHECK (pthread_barrier_init (&scheduler.init_barrier, NULL,
+                                       num_worker_threads + 1));
+
   for (i = 0; i < num_worker_threads; ++i)
     {
       scheduler.thread_pool[i].index = i;
@@ -113,6 +119,9 @@ pthread_scheduler_init (cl_device_id device)
                                      pocl_pthread_driver_thread,
                                      (void *)&scheduler.thread_pool[i]));
     }
+
+  PTHREAD_CHECK2 (PTHREAD_BARRIER_SERIAL_THREAD,
+                  pthread_barrier_wait (&scheduler.init_barrier));
 
   return CL_SUCCESS;
 }
@@ -135,6 +144,7 @@ pthread_scheduler_uninit ()
   pocl_aligned_free (scheduler.thread_pool);
   POCL_FAST_DESTROY (scheduler.wq_lock_fast);
   PTHREAD_CHECK (pthread_cond_destroy (&scheduler.wake_pool));
+  PTHREAD_CHECK (pthread_barrier_destroy (&scheduler.init_barrier));
 
   scheduler.thread_pool_shutdown_requested = 0;
 }
@@ -524,6 +534,9 @@ pocl_pthread_driver_thread (void *p)
           pthread_setaffinity_np (td->thread, sizeof (cpu_set_t), &set));
     }
 #endif
+
+  PTHREAD_CHECK2 (PTHREAD_BARRIER_SERIAL_THREAD,
+                  pthread_barrier_wait (&scheduler.init_barrier));
 
   while (1)
     {
--- a/lib/CL/devices/pthread/pthread_scheduler.c
+++ b/lib/CL/devices/pthread/pthread_scheduler.c
@@ -80,6 +80,8 @@ typedef struct scheduler_data_
 
   pthread_barrier_t init_barrier
       __attribute__ ((aligned (HOST_CPU_CACHELINE_SIZE)));
+
+  int worker_out_of_memory;
 } scheduler_data __attribute__ ((aligned (HOST_CPU_CACHELINE_SIZE)));
 
 static scheduler_data scheduler;
@@ -111,6 +113,7 @@ pthread_scheduler_init (cl_device_id device)
 
   PTHREAD_CHECK (pthread_barrier_init (&scheduler.init_barrier, NULL,
                                        num_worker_threads + 1));
+  scheduler.worker_out_of_memory = 0;
 
   for (i = 0; i < num_worker_threads; ++i)
     {
@@ -122,6 +125,12 @@ pthread_scheduler_init (cl_device_id device)
 
   PTHREAD_CHECK2 (PTHREAD_BARRIER_SERIAL_THREAD,
                   pthread_barrier_wait (&scheduler.init_barrier));
+
+  if (scheduler.worker_out_of_memory)
+    {
+      pthread_scheduler_uninit ();
+      return CL_OUT_OF_HOST_MEMORY;
+    }
 
   return CL_SUCCESS;
 }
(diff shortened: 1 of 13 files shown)


lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -28,6 +28,7 @@
 #endif
 
 #include <pthread.h>
+#include <stdatomic.h>
```
franz, line 31: We already have macros (`POCL_ATOMIC_INC` and so on) for atomics in `pocl_cl.h`, it would be great not to make new code less portable..

franz (commented):
LGTM in general.

BTW, are you aware that PoCL already has a CMake variable `PRINTF_BUFFER_SIZE` that can be used to set the printf buffer size to a smaller value (default is 16MB per CPU thread).  Perhaps it would make sense to change the 16MB to a smaller default on 32bit OS.

lib/CL/devices/pthread/pthread_scheduler.c
anbe42: I didn't know about them. Switched now to POCL_ATOMIC_INC.

lib/CL/devices/pthread/CMakeLists.txt
```diff
@@ -28,5 +28,5 @@ if(MSVC)
 endif(MSVC)
 add_pocl_device_library(pocl-devices-pthread pocl-pthread.h pthread.c pthread_scheduler.c pthread_utils.c)
 if(ENABLE_LOADABLE_DRIVERS)
-target_link_libraries(pocl-devices-pthread PRIVATE pocl-devices-basic)
+target_link_libraries(pocl-devices-pthread PRIVATE pocl-devices-basic ${PTHREAD_LIBRARY})
```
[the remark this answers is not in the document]
anbe42: Relying on transitive dependencies is not good practice ...
Debian tooling warns about such unresolved symbols, and having unneccessary warnings in there makes it harder to find the actually missing symbols when building with VISIBILITY_HIDDEN.

anbe42:
> BTW, are you aware that PoCL already has a CMake variable `PRINTF_BUFFER_SIZE` that can be used to set the printf buffer size to a smaller value (default is 16MB per CPU thread). Perhaps it would make sense to change the 16MB to a smaller default on 32bit OS.
I'll try that in addition to the other fixes on 32-bit arm. 4 MB should hopefully still be ok unless there are extremely chatty applications ...

anbe42:
I just noticed, there is a PTHREAD_CHECK macro. Is there a reason for using that only by the hsa device?
Otherwise I could convert all remaining pthread_*() calls to use it as well.

pjaaskel:
> LGTM in general.
> 
> BTW, are you aware that PoCL already has a CMake variable `PRINTF_BUFFER_SIZE` that can be used to set the printf buffer size to a smaller value (default is 16MB per CPU thread). Perhaps it would make sense to change the 16MB to a smaller default on 32bit OS.

Shouldn't that variable be defined per device driver for devices where it could be configured?

franz:
@pjaaskel yes good point

@anbe42 yes `PTHREAD_CHECK` should be used everywhere. There are also a few pthread-related `POCL_XYZ` macros as abstractions in `pocl_cl.h`,  these should also use `PTHREAD_CHECK` instead of asserts. It's quite inconsistent now.

franz:
Thanks. Quick check (only LLVM 12) on buildbots: 1) TCE fails to build on syntax errors; 2) ASAN-enabled PoCL (with CPU driver) builds but fails some tests. Other build configs pass.

TCE:
```
In file included from /worker/llvm_120_tcemc/source/lib/CL/devices/common.h:28,
                 from /worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:26:
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc: In function 'int pocl_tce_init_queue(cl_device_id, cl_command_queue)':
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:123:3: error: expected primary-expression before 'do'
   do                                                                          \
   ^~
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:186:27: note: in expansion of macro 'PTHREAD_CHECK'
 #define POCL_INIT_COND(c) PTHREAD_CHECK (pthread_cond_init (&c, NULL))
                           ^~~~~~~~~~~~~
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:1025:7: note: in expansion of macro 'POCL_INIT_COND'
   if (POCL_INIT_COND(dd->cq_cond))
       ^~~~~~~~~~~~~~
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:1025:7: error: expected ')' before 'do'
   if (POCL_INIT_COND(dd->cq_cond))
      ~^
       )
In file included from /worker/llvm_120_tcemc/source/lib/CL/devices/common.h:28,
                 from /worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:26:
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc: In function 'void pocl_tce_notify_cmdq_finished(cl_command_queue)':
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:123:3: error: expected primary-expression before 'do'
   do                                                                          \
   ^~
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:189:32: note: in expansion of macro 'PTHREAD_CHECK'
 #define POCL_BROADCAST_COND(c) PTHREAD_CHECK (pthread_cond_broadcast (&c))
                                ^~~~~~~~~~~~~
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:1087:11: note: in expansion of macro 'POCL_BROADCAST_COND'
   int r = POCL_BROADCAST_COND(dd->cq_cond);
           ^~~~~~~~~~~~~~~~~~~
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc: In function 'void pocl_tce_join(cl_device_id, cl_command_queue)':
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:123:3: error: expected primary-expression before 'do'
   do                                                                          \
   ^~
/worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:190:30: note: in expansion of macro 'PTHREAD_CHECK'
 #define POCL_WAIT_COND(c, m) PTHREAD_CHECK (pthread_cond_wait (&c, &m))
                              ^~~~~~~~~~~~~
/worker/llvm_120_tcemc/source/lib/CL/devices/tce/tce_common.cc:1100:15: note: in expansion of macro 'POCL_WAIT_COND'
       int r = POCL_WAIT_COND(dd->cq_cond, cq->pocl_lock);
               ^~~~~~~~~~~~~~
```

ASAN build (`cmake -DENABLE_ASAN=1  -DENABLE_ICD=0`):
```
	 92 - runtime/clGetDeviceInfo (Failed)
	 97 - runtime/test_kernel_cache_includes (Failed)
	100 - runtime/test_link_error (Failed)
	102 - runtime/test_fill-buffer (Failed)
	104 - runtime/clCreateKernel (Failed)
	107 - runtime/clGetSupportedImageFormats (Failed)
	116 - runtime/clSetMemObjectDestructorCallback (Failed)
```
tests fail with "  STDERR:   Unknown error -1".

pjaaskel (changes requested):
Pls check the issues Michal found.

anbe42:
> Thanks. Quick check (only LLVM 12) on buildbots: 1) TCE fails to build on syntax errors; 2) ASAN-enabled PoCL (with CPU driver) builds but fails some tests. Other build configs pass.

> /worker/llvm_120_tcemc/source/lib/CL/./pocl_cl.h:123:3: error: expected primary-expression before 'do'

TCE actually checked some return values ;-) that didn't work any longer after a block was added around the function call in the macro ...

> ASAN build (`cmake -DENABLE_ASAN=1 -DENABLE_ICD=0`):
...
> tests fail with " STDERR: Unknown error -1".

That was another occurrence of the barrier returning `PTHREAD_BARRIER_SERIAL_THREAD` ...

Both issues should be fixed, and vulkan as well ;-)

franz:
TCE still fails to build:

```
/worker/llvm_130_tcemc/source/lib/CL/devices/tce/tce_common.cc:1026:7: note: in expansion of macro ‘POCL_INIT_COND’
 1026 |   if (POCL_INIT_COND(dd->cq_cond))
      |       ^~~~~~~~~~~~~~
```

i'm guessing  it doesn't like the macro expanding to `if (do {} while ())`. Removing the if makes TCE build pass.

PR #1012, opened 2021-11-30
enabling VISIBILITY_HIDDEN by default

anbe42:
Here are a few more fixes needed when enabling VISIBILITY_HIDDEN. I also tested building the cuda and hsa devices in this configuration, but I can't build the tce device. If there are still missing symbols in tce, please send me `nm -D libpocl-devices-tce.so` and I should be able to figure out what needs to be exported as well.

With these fixes in place, I think we could enable VISIBILITY_HIDDEN by default. If you don't want to do this, just leave out that commit.

Since this is going to export more symbols from the basic device, are these routines that should be moved to the common code?


lib/llvmopencl/LLVMUtils.h
```diff
@@ -49,6 +49,7 @@ void
 regenerate_kernel_metadata(llvm::Module &M, FunctionMapping &kernels);
 
 // Remove a function from a module, along with all callsites.
+POCL_EXPORT
 void eraseFunctionAndCallers(llvm::Function *Function);
```
pjaaskel, line 53: I'm wondering should we add pocl_ prefix to this libpocl.so exported function for unity?

anbe42: This is a C++ symbol in pocl namespace, so I'm fine with it.
`_ZN4pocl23eraseFunctionAndCallersEPN4llvm8FunctionE@Base`

But I found some more symbols needed by the accel device, just added a commit to export them.
These don't fit the naming scheme and should probably be renamed: `alloc_buffer_from_region`, `free_chunk`, `init_mem_region`. Especially the second one sounds generic enough to allow clashes with other libraries.

That leaves only one already exported symbol not matching the naming scheme: `get_llvm_cpu_name`

franz:
indeed there are still missing symbols in TCE:

```
  /worker/llvm_120_tcemc/build/tests/regression/test_early_return: symbol
  lookup error:
  /worker/llvm_120_tcemc/build/lib/CL/devices/tce//libpocl-devices-ttasim.so:
  undefined symbol: pocl_SHA1_Init
  ```

I've put the `nm -D` output [here.](https://gist.github.com/franz/fa4c62c1675ac5c654d66134989d3eb0)

**Comment by anbe42** [comment]
@franz: I've tried to export all symbols needed by tce and added the pocl_ prefix to exported symbols where was missing.
Do you need more symbols exported for the vulkan and proxy devices?

**Comment by franz** [comment]
TCE still missing one:

```
  /worker/llvm_120_tcemc/build/tests/tce/ttasim/ttasim_host: symbol lookup
  error:
  /worker/llvm_120_tcemc/build/lib/CL/devices/tce//libpocl-devices-ttasim.so:
  undefined symbol: byteswap_uint32_t
```

Proxy device needs the visibility option force-disabled. This is necessary because it relies on `POcl*` symbols being visible in libpocl. Either something like
```
if(ENABLE_PROXY_DEVICE)
set(VISIBILITY_HIDDEN OFF CACHE BOOL "Proxy driver can't work with -fvisibility=hidden" FORCE)
else()...
```
... or return an error if both options are enabled.

WRT Vulkan: if you rebase this PR on current master, i can check. It does use `alloc_buffer_from_region` and `init_mem_region`  from `bufalloc.h`.

**Comment by anbe42** [comment]
@franz: rebased, renamed/exported one more symbol, fixed the symbol names used by the vulkan driver
still thinking how to handle the proxy driver ...

**Comment by anbe42** [comment]
rebased, handled the proxy device (but that fails to build due to unrelated reasons) and renamed a few exported `pocl_basic_*()` symbols to `pocl_driver_*()` and relocated them to `common_driver.c`

**Comment by pjaaskel** [comment]
Vulkan driver fails to build:
```
/home/visit0r/src/pocl/lib/CL/devices/vulkan/pocl-vulkan.c:501:3: error: implicit declaration of function ‘init_init_mem_region’; did you mean ‘pocl_init_mem_region’? [-Werror=implicit-function-declaration]
   init_init_mem_region (&d->kernarg_region, 0, d->kernarg_size);
   ^~~~~~~~~~~~~~~~~~~~
   pocl_init_mem_region
/home/visit0r/src/pocl/lib/CL/devices/vulkan/pocl-vulkan.c: In function ‘pocl_vulkan_setup_kernel_arguments’:
/home/visit0r/src/pocl/lib/CL/devices/vulkan/pocl-vulkan.c:2239:25: warning: assignment from incompatible pointer type [-Wincompatible-pointer-types]
       shader_info.pCode = program->binaries[dev_i];
                         ^
cc1: some warnings being treated as errors
```

anbe42:
Oops, misrenamed that one symbol. And vulkan was not yet enabled while build-testing the branch. Fixed.

PR #1013, opened 2021-12-01
Cleanup tests

anbe42:
Some cleanup for the tests, mostly increasing consistency.

Adds a new helper `poclu_show_program_build_log(cl_program)`.

Adds a new regression test for a llvm segfault triggered by pocl in llvm-10/llvm-11/llvm-12, seems to be fixed in llvm-13.

And while doing cleanup, I came across ...

There are two files left with mixed CRLF and CR line endings:
```
examples/IntelSVM/intelsvm.patch
examples/opencl-book-samples/ImageFilter2D.cl
```

With the minimal supported llvm version being 6.0, are the patches for older versions under tools/patches still needed?


lib/poclu/misc.c
```diff
@@ -198,6 +198,68 @@ poclu_write_file (const char *filemane, char *content, size_t size)
...
 }
 
+cl_int
+poclu_show_program_build_log (cl_program program)
+{
+  cl_int err;
+  cl_uint num_devices;
+  cl_device_id *devices;
+
+  err = clGetProgramInfo(program, CL_PROGRAM_NUM_DEVICES, sizeof(num_devices), &num_devices, NULL);
```
pjaaskel, line 208: These rows look long, can you run the clang formatter script on this commit?

pjaaskel (changes requested):
Other cleanups are fine, but I'd rather not pull huge white space cleanups as they make 'git blame' less handy. So perhaps revert the white space cleanups to the tests, if possible?

lib/poclu/misc.c
anbe42: done

anbe42:
> Other cleanups are fine, but I'd rather not pull huge white space cleanups as they make 'git blame' less handy. So perhaps revert the white space cleanups to the tests, if possible?

rebased, omitting the CRLF elimination

pjaaskel (approved):

anbe42:
just ran the whole series through the formatter

franz:
OK, minor issues found:

1) `regression/test_llvm_segfault_issue_889` segfaults on "distro" build (`cmake -DKERNELLIB_HOST_CPU_VARIANTS=distro -DPOCL_ICD_ABSOLUTE_PATH=OFF -DENABLE_POCL_BUILDING=OFF -DENABLE_LLVM=1  -DENABLE_ICD=1 -DENABLE_HOST_CPU_DEVICES=1 `)

2) putting internal label on all tests causes 3  tests to fail with the ASAN build (`runtime/test_buffer_migration`,  `runtime/test_buffer_ping_pong` and `runtime/test_deviceside_enqueue`) because they leak some memory.

Additionally the dlopen tests should only be enabled when `ENABLE_LOADABLE_DRIVERS` is ON (but i already have a commit in the vulkan PR that fixes this).

pjaaskel:
> With the minimal supported llvm version being 6.0, are the patches for older versions under tools/patches still needed?

Not needed.

anbe42:
@franz:
>     1. `regression/test_llvm_segfault_issue_889` segfaults on "distro" build

I assume that happened in a build targeting your AMD CPU? With which llvm versions? Does it pass with llvm-13?

Unfortunately there isn't a test property `MAY_FAIL` ... or `FLAKY` ...

>     2. putting internal label on all tests causes 3  tests to fail with the ASAN build (`runtime/test_buffer_migration`,  `runtime/test_buffer_ping_pong` and `runtime/test_deviceside_enqueue`) because they leak some memory.

I don't think these are regressions ... especially since I haven't touched the last one.

> Additionally the dlopen tests should only be enabled when `ENABLE_LOADABLE_DRIVERS` is ON (but i already have a commit in the vulkan PR that fixes this).

Should the first dlopen test be enabled on `ENABLE_ICD`, regardless of the  `ENABLE_LOADABLE_DRIVERS` setting?
And the outer condition should probably be `if (UNIX AND HAVE_DLFCN_H)`

franz:
> Unfortunately there isn't a test property MAY_FAIL ...

I'd rather avoid flaky tests, they waste time...

That said, this seems to fail (consistently) on AMD x86-64 with LLVM < 13, works on AMD with LLVM 13 and works on Intel (tested only LLVM 11). The AMD LLVM backend for whatever reason is very different from Intel LLVM backend, and i've also seen issues in opposite direction (crash on Intel, work on AMD).

Since the problem has been fixed in LLVM 13 ... how about we enable the test only for LLVM 13 and newer.

> I don't think these are regressions

You're right, my bad.

> Should the first dlopen test be enabled on ENABLE_ICD, regardless of the ENABLE_LOADABLE_DRIVERS setting?
> And the outer condition should probably be if (UNIX AND HAVE_DLFCN_H)

yes, good points.

anbe42:
> Since the problem has been fixed in LLVM 13 ... how about we enable the test only for LLVM 13 and newer.

Can we use cmake test property DISABLED (requires cmake 3.9)?

pjaaskel:
Yep or just do not add the test if older than LLVM. Check examples/conformance/CMakeLists.txt where there's (now obsolete, since we do not support so old LLVMs) check for older than 5.0.

anbe42:
rebased and added some dlopen test fixes for builds without `ENABLE_ICD`

marked llvm segfault test as `WILL_FAIL` where reproducible, as `DISABLED` where flaky
I'll happily extend the blacklist if more bad combinations show up

pjaaskel:
The LLVM link test fails here after pulling this in. My guess is that it doesn't propagate the correct C++ header switch to Clang (I use system's headers, not libstdc++), ending up with:
```
--   Using -march=
-- LLVM link test output: Change Dir: /home/visit0r/src/pocl/test-pr-cmake-llvm-13-build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_7b9d8/fast"
/usr/bin/make -f CMakeFiles/cmTC_7b9d8.dir/build.make CMakeFiles/cmTC_7b9d8.dir/build
make[1]: Entering directory '/mnt/olddisk2/visit0r/src/pocl/test-pr-cmake-llvm-13-build/CMakeFiles/CMakeTmp'
Building CXX object CMakeFiles/cmTC_7b9d8.dir/llvm_link_test_bpAzp.cc.o
/usr/bin/g++  -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  -g -O3 -Wall -Wextra -Wno-long-long  -fPIE   -g -O3 -Wall -Wextra -Wno-long-long -I/usr/lib/llvm-13/include -std=c++14 -fno-exceptions -std=gnu++11 -o CMakeFiles/cmTC_7b9d8.dir/llvm_link_test_bpAzp.cc.o -c /home/visit0r/src/pocl/test-pr-cmake-llvm-13-build/llvm_link_test_bpAzp.cc
In file included from /usr/lib/llvm-13/include/llvm/ADT/Hashing.h:49:0,
                 from /usr/lib/llvm-13/include/llvm/ADT/Optional.h:18,
                 from /usr/lib/llvm-13/include/llvm/ADT/STLExtras.h:19,
                 from /usr/lib/llvm-13/include/llvm/ADT/StringRef.h:12,
                 from /usr/lib/llvm-13/include/llvm/IR/DiagnosticHandler.h:15,
                 from /usr/lib/llvm-13/include/llvm/IR/LLVMContext.h:18,
                 from /home/visit0r/src/pocl/test-pr-cmake-llvm-13-build/llvm_link_test_bpAzp.cc:3:
/usr/lib/llvm-13/include/llvm/Support/SwapByteOrder.h:152:13: error: ‘enable_if_t’ in namespace ‘std’ does not name a template type
 inline std::enable_if_t<std::is_enum<T>::value, T> getSwappedBytes(T C) {
...
```

anbe42:
> The LLVM link test fails here after pulling this in.

cmake policy issue. Reproducible by simply bumping cmake_minimum_required to 3.9 in master. Weird cmake error: disappears if rerunning the cmake command with the partial CMakeCache.txt from previous failed run available.
For now I'm downgrading the policy, but at some point the custom_try_compile macros should be "modernized".

Added another WILL_FAIL test which should help diagnosing the printf issues.

pjaaskel:
Vulkan driver fails:
```
3/163 Test: pocl_test_dlopen_device_vulkan                                                                                                
Command: "/home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/tests/runtime/test_dlopen" "vulkan"                                     
Directory: /home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/tests                                                                  
"pocl_test_dlopen_device_vulkan" start time: Dec 16 10:25 EET                                                                             
Output:
----------------------------------------------------------
dlopen($ORIGIN/../../lib/CL/devices/vulkan/libpocl-devices-vulkan.so, RTLD_NOW) failed: /mnt/olddisk2/visit0r/src/pocl/test-pr-cmake-vulk\
an-tests-build/tests/runtime/../../lib/CL/devices/vulkan/libpocl-devices-vulkan.so: undefined symbol: pocl_alloc_buffer_from_region
<end of output>
Test time =   0.03 sec    
```

pjaaskel:
Also svm symbols missing (perhaps due to the 3.x commits pulled in?):
```
  1/332 Test   #3: pocl_test_dlopen_device_basic .......................................................***Failed    0.15 sec
dlopen($ORIGIN/../../lib/CL/devices/basic/libpocl-devices-basic.so, RTLD_NOW) failed: /mnt/olddisk2/visit0r/src/pocl/test-pr-cmake-llvm-13-build/tests/runtime/../../lib/CL/devices/basic/libpocl-devices-basic.so: undefined symbol: pocl_driver_svm_fill

  2/332 Test   #2: pocl_test_dlopen_libpocl ............................................................   Passed    0.15 sec
  3/332 Test   #4: pocl_test_dlopen_device_pthread .....................................................***Failed    1.16 sec
dlopen($ORIGIN/../../lib/CL/devices/pthread/libpocl-devices-pthread.so, RTLD_NOW) failed: /home/visit0r/src/pocl/test-pr-cmake-llvm-13-build/lib/CL/devices/basic/libpocl-devices-basic.so: undefined symbol: pocl_driver_svm_fill
```

anbe42:
> Vulkan driver fails:
> an-tests-build/tests/runtime/../../lib/CL/devices/vulkan/libpocl-devices-vulkan.so: undefined symbol: pocl_alloc_buffer_from_region

Did you use the latest version of that branch? That symbol (and the svn ones) were already exported ... but there were actually 3 symbols missing that are needed by the vulkan device. I've prepended two fixing commits to the cleanup-tests branch ...

pjaaskel:
Still fails:
```
  symbol lookup error:                                                                                                                    
  /home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/lib/CL/devices/vulkan//libpocl-devices-vulkan.so:                               
  undefined symbol: pocl_cache_create_program_cachedir    
```

anbe42:
>   `undefined symbol: pocl_cache_create_program_cachedir`

That usage was hidden behind `#ifdef HAVE_CLSPV` and thus not in a code path tested on my builds ...
More symbols exported and renamed.

pjaaskel:
Now complaining about undefined symbol: pocl_bitcode_is_spirv_execmodel_sha

anbe42:
> Now complaining about undefined symbol: pocl_bitcode_is_spirv_execmodel_sha

I cannot reproduce that. Sounds like an incomplete rebuild.

pjaaskel:
No, it's persistent. My guess is that you build the CPU devices in and I don't, in my test env, but use the following cmake flags:
```
-DENABLE_EXPERIMENTAL_DRIVERS=1 -DENABLE_TCE=OFF -DENABLE_VULKAN=ON -DENABLE_HOST_CPU_DEVICES=OFF -DCLSPV=/home/visit0r/local/bin/clspv
```

pjaaskel (changes requested):
A Vulkan linkage issue when host devices switched off.

anbe42:
I still cannot reproduce that issue ...
I've configured cleanup-tests at commit a9e0fae09baac with
```
cmake \ 
        -DWITH_LLVM_CONFIG=/usr/bin/llvm-config-13 \
        -DCMAKE_C_COMPILER=gcc-11 -DCMAKE_CXX_COMPILER=g++-11 \
        -DENABLE_TCE=OFF -DENABLE_VULKAN=ON -DENABLE_HOST_CPU_DEVICES=OFF -DCLSPV_DIR=/tmp/usr/local/bin \
        ..
```
The tests fail due to lack of a supported device, but I don't see missing symbols.
```
$ nm -D ./lib/CL/libpocl.so.2.9.0 | grep pocl_bitcode_is_spirv_execmodel_shader
0000000000062884 T pocl_bitcode_is_spirv_execmodel_shader
$ nm -D ./lib/CL/devices/vulkan/libpocl-devices-vulkan.so | grep pocl_bitcode_is_spirv_execmodel_shader
                 U pocl_bitcode_is_spirv_execmodel_shader
```

You can find `grep -vE '^(#|//|$)' CMakeCache.txt` at https://gist.github.com/anbe42/f38f786b90e34710b4342b7f07e12018

If I use `-DCLSPV=/tmp/usr/local/bin/clspv` instead of setting `CLSPV_DIR`, the code path calling `pocl_bitcode_is_spirv_execmodel_shader` gets deactivated and `libpocl-devices-vulkan.so` no longer references the symbol.

tests/runtime/test_dlopen.c
```diff
@@ -43,13 +43,15 @@ main (int argc, char **argv)
               SHLIB_ORIGIN "/../../lib/CL/devices/%s/libpocl-devices-%s.so", argv[1],
               argv[1]);
 
+#ifdef ENABLE_ICD
```
pjaaskel, line 46: This causes the breakage. The file should include config.h and use BUILD_ICD macro instead. Now it's never loading the libpocl.so, thus its utility functions are not imported unless linking statically. If I fix that, vulkan works, but then I see some other failures with LLVM 10 and conformance test micro.

pjaaskel (changes requested):
Getting there...

tests/runtime/test_dlopen.c
pjaaskel: (Managed to successfully run my test envs, pushing after another clean run)

anbe42:
Thanks for fixing the last bit.

BTW, there is another occurrence of `#ifdef ENABLE_ICD` in `lib/CL/devices/proxy/pocl_proxy.c`

PR #1014, opened 2021-12-01
Miscellaneous fixes

anbe42:
These fixes I have accumulated in the Debian package over time or they were born while I was prepareing the other merge requests.

The underlinked device libraries (relying transitively on libpocl.so.2 being linked against the missing libraries) made finding the actually missing symbol exports for VISIBILITY_HIDDEN more difficult.

As with my the other merge requests, feel free to cherry-pick only some bits if you don't want the others.


anbe42:
Added some fixes for the formatting scripts.

While doing that, I noticed `tools/scripts/clang-format-diff.py` is not accompanied by its license file. The same is true for `lib/llvmopencl/BreakConstantGEPs.*`
And you could also add a copy of the Boost license for `lib/kernel/sleef/*` ...

(I haven't done a complete license audit.)

franz:
LGTM in general, and passes the buildbots, but i have not checked the HSA commits. @pjaaskel  can you check

Also, the `libllvmopencl` might be still useful for testing LLVM PoCL passes - so maybe it should not be removed, just disabled by default ? @pjaaskel what do you think ?

anbe42:
> Also, the `libllvmopencl` might be still useful for testing LLVM PoCL passes - so maybe it should not be removed, just disabled by default ?

building `libllvmopencl.so` is now disabled by default

pjaaskel:
(At least) Vulkan fails:
```
142/161 Testing: examples/example2                                                                                                        
142/161 Test: examples/example2                                                                                                           
Command: "/home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/examples/example2/example2"                                             
Directory: /home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/examples/example2                                                      
"examples/example2" start time: Dec 16 10:44 EET                                                                                          
Output:                                                                                                                                   
----------------------------------------------------------
/home/visit0r/src/pocl/test-pr-cmake-vulkan-tests-build/examples/example2/example2: symbol lookup error: /home/visit0r/src/pocl/test-pr-c\
make-vulkan-tests-build/lib/CL/devices/vulkan//libpocl-devices-vulkan.so: undefined symbol: pocl_cache_create_program_cachedir
<end of output>                                                                                                                           
Test time =   0.13 sec   
```

PR #1015, opened 2021-12-05
Avoid loss of precision in CUDA event time calculations

nchristensen:
The current code implicitly converts `epoch` to a float before calculating `time_end` and `time_start` and converting back to a `cl_ulong`. A `float` only has seven or so digits of accuracy meaning this conversion throws away the least significant digits of `epoch` if `epoch` is larger than about 1e8. Conversely, the precision of `diff` is around 1e-3 so multiplying by 1e6 makes all of these digits > 1 and converting this number to a `cl_ulong` will not lose any  digits. The gist below illustrates this in Python.

https://gist.github.com/nchristensen/15ab1fd53ba099accaab780ab6dbad90

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,7 @@ Notable Bug Fixes
 - Fixed the post-event-finish-cleanup process in the drivers,
   this should improve reliability
 - Few more fixes for Android
+- Fixed loss of precision in CUDA event times due to type conversion
 
 Other
 -----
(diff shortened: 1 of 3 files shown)


pjaaskel:
I think this is enough user visible change (although small) that could be added to CHANGES. Also if you are not yet in CREDITS, can you add yourself there?

nchristensen:
Done!

PR #1016, opened 2021-12-08
OpenCL 3.0 CTS support + fixes

franz:
Main changes:

* updates the platform version to 3.0
* updates the CPU driver version to 3.0 if LLVM is sufficiently recent (12+)
* updates the included OpenCL headers, 
* updates `examples/conformance` to use the 3.0 version of CTS when possible.

Also changes the way PoCL codebase is built (this is to make it easier to use OpenCL 3.0 internally and at the same time build against older ICD / header versions).

* the library (`lib/*`) is always built against internal OpenCL 3.0 headers. 
* the poclu, tests & examples are now built against either 1) internal OpenCL 3.0 headers if ENABLE_ICD is OFF, or 2) against system headers & ICD (whatever the version) if ENABLE_ICD is ON.

At some point we can switch everything to 3.0-only but right now most systems have older ICDs / headers.

There are still a few failures with CTS 3.0 b/c of missing features.


pjaaskel:
Great work again Michal, thanks. Can you add documentation on how to enable the OpenCL 3.0 build (how to install new enough ICD loader, it boils down to, I suppose) and what is the overall status of the CTS? Can we claim minimal OpenCL 3.0 compliance with this PR or is there something to do still? CHANGES needs to state that too.

Vinsteri:
> Can we claim minimal OpenCL 3.0 compliance with this PR or is there something to do still? 

I would say that we can't claim OpenCL 3.0 compliance yet, there's still some mandatory OpenCL 3.0 features missing implementations. For example,
- [clSetContextDestructorCallback](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetContextDestructorCallback)
- All new versioning queries like [CL_PLATFORM_EXTENSIONS_WITH_VERSION](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_PLATFORM_EXTENSIONS_WITH_VERSION)
- [clCreateBufferWithProperties](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBufferWithProperties)
- [clCreateImageWithProperties](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateImageWithProperties)
and more..

franz:
OK, i have apparently missed some tests in the "api" test group. After adding them, still only 2 tests fail (one of them, get_kernel_arg_info, is expected). However it skips many tests:

```
context_destructor_callback skipped (requires at least OpenCL version 3.0, but the device reports OpenCL version 1.2)
```

I'm not sure why it skips context destructor tests if the platform reports 3.0 (IOW why is that test device dependent not platform dependent...)

But 3.0 can't be enabled on device yet, because it doesn't build, because of missing feature macros:

```
error: options cl_khr_fp64 and __opencl_c_fp64 are set to different values
```

so i think feature macros need to be solved first.

Vinsteri:
make install fails with 
```
CMake Error at include/CL/cmake_install.cmake:49 (file):
file INSTALL cannot find "~/pocl/include/CL/cl2.hpp": No such
file or directory.
Call Stack (most recent call first):
include/cmake_install.cmake:67 (include)
cmake_install.cmake:81 (include)
```
when -DINSTALL_OPENCL_HEADERS=ON

I suspect this is result of cl2.hpp being moved in commit [Move include/CL/cl2.hpp -> include/hpp/CL/cl2.hpp ](https://github.com/pocl/pocl/pull/1016/commits/96cd4e3d8c5354a6611ef5aef43a3421b30cd7b1), but cl2.hpp wasn't removed from the [include/CL/CMakeLists](https://github.com/franz/pocl/blob/opencl_30/include/CL/CMakeLists.txt#L30).

franz:
Rebased the PR on master, and removed two commits that enable the 3.0 CTS and 3.0 platform. I can commit them later when all 3.0 features are in.

pjaaskel:
Thanks! The diffs look good overall, but I get a build error when testing:
```
[ 71%] Building C object poclu/CMakeFiles/poclu.dir/misc.c.o
/home/visit0r/src/pocl/poclu/misc.c: In function ‘poclu_load_program_multidev’:
/home/visit0r/src/pocl/poclu/misc.c:469:17: error: implicit declaration of function ‘clCreateProgramWithIL’; did you mean ‘clCreateProgramWithBinary’? [-Werror=implicit-function-declaration]
       program = clCreateProgramWithIL (context, (const void *)binary,
                 ^~~~~~~~~~~~~~~~~~~~~
                 clCreateProgramWithBinary
/home/visit0r/src/pocl/poclu/misc.c:469:15: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
       program = clCreateProgramWithIL (context, (const void *)binary,

```

franz:
@pjaaskel  Should be fixed. Which OS are you testing with ?

Issue #962, opened 2021-06-24
Quotes in the compiler options are being ignored


jsteube:
I've tested this with latest POCL 1.7, but it occurs with older POCL versions, too.

Some users have space characters in their user name or installation folder of my application. My application puts the path to the applications' OpenCL include folder in quotes. This work well with all OpenCL runtimes (ROCm, AMD legacy, Intel OneAPI, NVIDIA OpenCL, Apple OpenCL) except POCL.

Example compiler options string:

```
-D KERNEL_STATIC -I OpenCL -I "/root/space test/hashcat/OpenCL" -D LOCAL_MEM_TYPE=2 -D VENDOR_ID=64
```

Received error string:

```
clCompileProgram(): CL_INVALID_COMPILER_OPTIONS

Invalid build option: test/hashcat/OpenCL"
```

Wild guess: There's a function somewhere in POCL which splits the string naively only with space, but ignores if strings put in quotes.

LionNatsu:
FYI: https://www.gnu.org/software/libc/manual/html_node/Argp.html

`error_t argp_parse (const struct argp *argp, int argc, char **argv, unsigned flags, int *arg_index, void *input)`

matrix:
Hi @pjaaskel, any update on this issue? do you need help?

Thanks!

pjaaskel:
Hi @matrix, yes helping hands are appreciated. Please fix and send a pull request. Since our group's work is not affected by this issue, we work on more urgent issues within the team.

matrix:
Hi @pjaaskel, I finished the patch :) Thanks

pjaaskel:
@matrix merged in, thanks. @jsteube can you test with the HEAD and close if it works?

matrix:
> @matrix merged in, thanks. @jsteube can you test with the HEAD and close if it works?

you can close, I did the test for him

pjaaskel:
Thanks!

PR #1028, opened 2022-01-19
handling whitespace chars in build options

matrix:
details in issue #962 

at the moment I only tried it outside POCL

### DIFF
--- a/lib/CL/pocl_llvm_build.cc
+++ b/lib/CL/pocl_llvm_build.cc
@@ -80,6 +80,7 @@ IGNORE_COMPILER_WARNING("-Wstrict-aliasing")
 #include "pocl_file_util.h"
 #include "pocl_cache.h"
 #include "LLVMUtils.h"
+#include "pocl_util.h"
 
 using namespace clang;
 using namespace llvm;
@@ -176,24 +177,6 @@ static std::string getPoclPrivateDataDir() {
     }
 #endif
     return POCL_INSTALL_PRIVATE_DATADIR;
-}
-
-/*
- * search for an unused ASCII character in temp_options,
- * to be used to replace whitespaces within double quoted substrings
- */
-static int find_unused_char (const char *options, char *replace_me)
-{
-  for (int y = 35; y < 128; y++)
-  {
-    if (strchr (options, (char) y) == NULL)
-    {
-      *replace_me = (char) y;
-      return 0;
-    }
-  }
-
-  return -1;
 }
 
 int pocl_llvm_build_program(cl_program program,
@@ -408,7 +391,7 @@ int pocl_llvm_build_program(cl_program program,
           /* at first need, get an unused char */
           if (replace_cnt == 0)
           {
-            if (find_unused_char (temp_options, &replace_me) == -1)
+            if (pocl_find_unused_char (temp_options, &replace_me) == -1)
             {
               /* no replace, no party */
               free (temp_options);
--- a/lib/CL/pocl_llvm_build.cc
+++ b/lib/CL/pocl_llvm_build.cc
@@ -357,18 +357,18 @@ int pocl_llvm_build_program(cl_program program,
 
   char WSReplacementChar = 0;
 
-  char *WSTempOptions = (char *) malloc (ss.str().length() + 1);
+  char *TempOptions = (char *) malloc (ss.str().length() + 1);
 
-  memset (WSTempOptions, 0, ss.str().length() + 1);
-  strncpy (WSTempOptions, ss.str().c_str(), ss.str().length());
+  memset (TempOptions, 0, ss.str().length() + 1);
+  strncpy (TempOptions, ss.str().c_str(), ss.str().length());
 
-  if (pocl_escape_quoted_whitespace (WSTempOptions, &WSReplacementChar) == -1)
+  if (pocl_escape_quoted_whitespace (TempOptions, &WSReplacementChar) == -1)
   {
-    POCL_MEM_FREE (WSTempOptions);
+    POCL_MEM_FREE (TempOptions);
     return CL_INVALID_BUILD_OPTIONS;
   }
 
-  std::istringstream iss(WSTempOptions);
+  std::istringstream iss(TempOptions);
   std::vector<const char *> itemcstrs;
   std::vector<std::string> itemstrs;
 
@@ -388,9 +388,9 @@ int pocl_llvm_build_program(cl_program program,
     // if quoted, remove it to make compiler happy
     if (s.find("\"") != std::string::npos)
     {
-      std::regex target("\"");
-      std::string replacement = " ";
-      s = std::regex_replace(s, target, replacement);
+      std::regex Target("\"");
+      std::string Replacement = " ";
+      s = std::regex_replace(s, Target, Replacement);
     }
 
     itemstrs.push_back(s);
@@ -402,7 +402,7 @@ int pocl_llvm_build_program(cl_program program,
     itemcstrs.push_back(itemstrs[idx].c_str());
   }
 
-  POCL_MEM_FREE (WSTempOptions);
+  POCL_MEM_FREE (TempOptions);
 
 #ifdef DEBUG_POCL_LLVM_API
   // TODO: for some reason the user_options are replicated,
--- a/lib/CL/pocl_llvm_build.cc
+++ b/lib/CL/pocl_llvm_build.cc
@@ -355,56 +355,17 @@ int pocl_llvm_build_program(cl_program program,
 
   POCL_MSG_PRINT_LLVM("all build options: %s\n", ss.str().c_str());
 
-  size_t replace_cnt = 0;
   char replace_me = 0;
 
   char *temp_options = (char *) malloc (ss.str().length() + 1);
+
+  memset (temp_options, 0, ss.str().length() + 1);
   strncpy (temp_options, ss.str().c_str(), ss.str().length());
 
-  /* searching for double quote in temp_options */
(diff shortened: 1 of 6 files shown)


matrix:
Tested on linux, with and without whitespaces in path

pjaaskel:
Thanks for your contribution. Looking at it though, it seems a bit complex. Could you simplify this by implementing a simple tokenizer instead that handles the quotes as a special case? Now you go through hoops to hide the space char just to pass it to the standard tokenizer whereas we could have our simpler one that scans the string and returns new tokens at space char unless a quote is being processed?

matrix:
> Thanks for your contribution. Looking at it though, it seems a bit complex. Could you simplify this by implementing a simple tokenizer instead that handles the quotes as a special case? Now you go through hoops to hide the space char just to pass it to the standard tokenizer whereas we could have our simpler one that scans the string and returns new tokens at space char unless a quote is being processed?

Hi, I think this solution it's the faster and less complex possible. As you see, I'm not changing the tokenizer, but only the input and the output. The best thing to do will be using array instead of strings, from the beginning.

I'm very busy with work and also in porting Metal on hashcat, so I can't help you at the moment. However the workaround is functional, you can always optimize it in the future.

Thanks for the comments, they are always appreciated :)

pjaaskel:
The workaround you add seems much more code than a simple tokenizer, but I suppose that's fine as it's not a beauty contest. However, I see duplicated code in injecting those chars on top of the quoted white spaces, thus at least should be refactored to a utility function to get this to an acceptable level. I still don't have extra time to "optimize" it by myself due to my own itches to scratch - sorry.

matrix:
yes, I can move the duplicate static function somewhere if you want. Thanks

matrix:
any suggestions on where you keep the shared functions?

pjaaskel:
pocl_util.c could fit this one, thanks.

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,84 @@ int pocl_llvm_build_program(cl_program program,
...
-  std::istream_iterator<std::string> end;
-  std::istream_iterator<std::string> i = begin;
-  std::vector<const char*> itemcstrs;
+  size_t replace_cnt = 0;
+  char replace_me = 0;
+
+  char *temp_options = (char *) malloc (ss.str().length() + 1);
+  strncpy (temp_options, ss.str().c_str(), ss.str().length());
+
+  /* searching for double quote in temp_options */
```
pjaaskel, line 364: This part still seems largely duplicated. Perhaps a function escape_quoted_whitespace() or such could be extracted?

pjaaskel (changes requested):

lib/CL/pocl_llvm_build.cc
matrix: yes, I will do soon

matrix: done

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,45 @@ int pocl_llvm_build_program(cl_program program,
...
+    // if needed, put back whitespace
+    if (replace_me != 0)
+    {
+      if (s.find(replace_me) != std::string::npos)
+      {
+        std::replace(s.begin(), s.end(), replace_me, ' ');
+      }
+    }
+
+    // if quoted, remove it to make compiler happy
```
pjaaskel, line 388: I'm not sure about this snippet here: Why do you do this here? Doesn't it break the tokenization when you have white spaces in the args for switches?

pjaaskel (changes requested):

lib/CL/pocl_llvm_build.cc
matrix: Hi, I'm sure because I test it before push. If you don't strip out quotes, all includes declared in the kernel source are appended to the quoted string, consequently the compiler returns an error because it cannot find the file to include.

Example:

`
"/this/is your/path"/inc_vendor.h
`

Thanks

pjaaskel: Doesn't your code convert ```/foo/"bar dir"/inc_vendor.h````to ```/foo/ bar dir /inc_vendor.h```, that is, add extra whitespaces to the path?

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,45 @@ int pocl_llvm_build_program(cl_program program,
 
   POCL_MSG_PRINT_LLVM("all build options: %s\n", ss.str().c_str());
 
-  std::istream_iterator<std::string> begin(ss);
-  std::istream_iterator<std::string> end;
-  std::istream_iterator<std::string> i = begin;
-  std::vector<const char*> itemcstrs;
+  char replace_me = 0;
```
pjaaskel, line 358: Also please take note this is a .cc file so we follow the LLVM style (yeah - I'm still sorry for the two different styles), so please make sure the new codes convert these files towards the style guide. It here for example means WSReplacementChar as the variable name that follows the convention.

pjaaskel (changes requested):
Please check my new comments, thanks for your patience.

matrix:
Sorry but I don’t understand what you say about style and the last
sentence. This patch works, and it’s tested on Linux. If you want to add I
and all hashcat users will be happy, else we need to handle this by stop
hashcat execution in case of white space on path.

Thanks, have a nice day

pjaaskel:
> Sorry but I don’t understand what you say about style and the last sentence. This patch works, and it’s tested on Linux. If you want to add I and all hashcat users will be happy, else we need to handle this by stop hashcat execution in case of white space on path. Thanks, have a nice day

Sorry for being unclear, like too often. What I meant to say is that the .cc (since being a C++) doesn't adhere to the PoCL style guidelines, which we gradually fix with the new commits. The idea is to improve the style with each commit, thus we do not push commits that go away from the agreed style guidelines, but preferably improve it a bit from the surroundings also.

There's a script `tools/scripts/format_branch.sh` which should fix the style automatically, but doesn't fix the naming conventions. E.g. ThisIsAVariable instead of this_is_a_variable. Yes, I understand it would be beneficial to your use cases to get this committed quickly and with minimal contributor effort, but surely you understand I must also guard the PoCL's code base since being responsible of its longer term maintenance. You'd likely require the same for a code base you were responsible for, right?

Have a great week!

lib/CL/pocl_llvm_build.cc
```diff
@@ -355,14 +355,14 @@ int pocl_llvm_build_program(cl_program program,
 
   POCL_MSG_PRINT_LLVM("all build options: %s\n", ss.str().c_str());
 
-  char replace_me = 0;
+  char WSReplacementChar = 0;
 
   char *temp_options = (char *) malloc (ss.str().length() + 1);
```
pjaaskel, line 360: This also needs renaming. Please check: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly    Thanks!

pjaaskel (changes requested):

matrix:
I'm sorry but the time I could devote to this is over. If you think that
what has been done is not enough, I will close the PR asap. Thank you

Il giorno lun 31 gen 2022 alle ore 13:09 Pekka Jääskeläinen <
***@***.***> ha scritto:

pjaaskel:
Sure, I understand. We all have our time priorities. Thanks for your first steps, perhaps someone else can take it to the mergeable level.

matrix:
"first steps" seems a bit simplistic to me, thanks anyway

Il giorno lun 31 gen 2022 alle ore 14:37 Pekka Jääskeläinen
***@***.***> ha scritto:

pjaaskel:
Yep, you are right, it's not far.

matrix:
@pjaaskel now it's all done or there're others changes to do? If not, please do a complete review also considering that the present code does not follow everything you have told me to do.
Thanks!

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,45 @@ int pocl_llvm_build_program(cl_program program,
 
   POCL_MSG_PRINT_LLVM("all build options: %s\n", ss.str().c_str());
 
-  std::istream_iterator<std::string> begin(ss);
-  std::istream_iterator<std::string> end;
-  std::istream_iterator<std::string> i = begin;
-  std::vector<const char*> itemcstrs;
+  char WSReplacementChar = 0;
```
pjaaskel, line 358: The WS here was meant to denote "White Space", not a prefix for all variables. So WSTempOptions could be just TempOptions.

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,45 @@ int pocl_llvm_build_program(cl_program program,
...
+      if (s.find(WSReplacementChar) != std::string::npos)
+      {
+        std::replace(s.begin(), s.end(), WSReplacementChar, ' ');
+      }
+    }
+
+    // if quoted, remove it to make compiler happy
+    if (s.find("\"") != std::string::npos)
+    {
+      std::regex target("\"");
```
pjaaskel, line 391: -> Target

lib/CL/pocl_llvm_build.cc
```diff
@@ -353,14 +355,45 @@ int pocl_llvm_build_program(cl_program program,
...
+      std::string replacement = " ";
```
pjaaskel, line 392: -> Replacement  (capital first)

pjaaskel (changes requested):
Thanks, just these couple of style issues pop up anymore.

matrix:
done

PR #1042, opened 2022-03-30
Check cl_version.h for CL version

nchristensen:
The CL_VERSION_X_X symbols appear live in cl_version.h (see https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_version.h). With this change cl_version.h is checked for the CL versions if it is present.


franz:
Hello, and thanks for contribution. Out of curiosity, did you run into an issue with the current CMake code, on some platform ? AFAICT, `CL/opencl.h` should include `CL/cl.h` which in turn includes `CL/cl_version.h`.

franz:
Can you check in the conda environment if the `CL/opencl.h` includes `CL/cl.h`, and cl.h includes `CL/cl_version.h` ? if that is the case, probably `check_symbol_exists` can't find the header because the include path doesn't contain the conda include directory. In that case the solution is to add the full include path explicitly to the `check_symbol_exists` macro.

nchristensen:
Closing since this is most likely an issue with the include paths and I haven't needed to rebuild pocl in a while. If I encounter this problem again and find it isn't a configuration problem I will re-open this.

PR #1043, opened 2022-03-30
test fixes + opencl.hpp

franz:
small fixes in tests + an update by jansol, updates headers to latest from Khronos (`cl2.hpp` was renamed to `opencl.hpp`).


tests/regression/test_barrier_before_return.cpp
```diff
@@ -1,166 +1,166 @@
-/* Tests a kernel with a barrier as the last statement.
```
pjaaskel, line 1: I prefer not to push mass white space cleanup commits since they make 'git blame' more difficult to track etc.

pjaaskel (changes requested):
Please drop the mass WS commit and this is OK.

tests/regression/test_barrier_before_return.cpp
franz: OK, dropped

PR #1046, opened 2022-04-10
CUDA: Set max_parameter_size to correct value

matthiasdiener:
This limit is imposed by ptxas, and is the same value
returned by NVidia CL.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -428,6 +428,9 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
   dev->max_mem_alloc_size = max (memtotal / 4, 128 * 1024 * 1024);
   dev->global_mem_size = memtotal;
 
+  // All devices starting from Compute Capability 2.0 have this limit;
+  // See e.g.
+  // https://forums.developer.nvidia.com/t/max-size-of-cuda-arguments/50218
   dev->max_parameter_size = 4352;
 
   dev->data = data;
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -428,6 +428,8 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
   dev->max_mem_alloc_size = max (memtotal / 4, 128 * 1024 * 1024);
   dev->global_mem_size = memtotal;
 
+  dev->max_parameter_size = 4352;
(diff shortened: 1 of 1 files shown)


lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -428,6 +428,8 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
   dev->max_mem_alloc_size = max (memtotal / 4, 128 * 1024 * 1024);
   dev->global_mem_size = memtotal;
 
+  dev->max_parameter_size = 4352;
```
inducer, line 431: I don't think this is constant across different microarchitectures. On Pascal (Compute Capability 5.2), Nvidia CL gives 1024.

Maybe this can be found from a CUDA-side device attribute:

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ge12b8a782bebe21b1ac0091bf9f4e2a3

inducer: Whoops. Accidentally looked at what POCL returns. It's 4352 for Pascal as well, so I don't actually have evidence of this changing. I think it was different in ancient times, but I think it's safe to say that's no longer relevant. If possible, I think it'd still be good to find a primary source for this number, but practically I agree this is good enough.

matthiasdiener: The oldest CUDA Programming Guide I could find online (8.0) already has the 4k limit: https://docs.nvidia.com/cuda/archive/8.0/cuda-c-programming-guide/index.html#function-parameters

Nvidia Forum posts suggest that the change from 256 bytes to 4352 happened with Compute Capability 2: https://forums.developer.nvidia.com/t/max-size-of-cuda-arguments/50218 , https://forums.developer.nvidia.com/t/is-there-any-limit-on-of-arguments-in-cuda-kernel/14784

matthiasdiener: > Maybe this can be found from a CUDA-side device attribute:
> 
> https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ge12b8a782bebe21b1ac0091bf9f4e2a3

I could not find any attribute that provides this information. `nvidia-smi -a` also seems to be silent about it.

inducer (approved):

matthiasdiener:
Related question: What do we do with the 1024-byte limit on the pthread device?

PR #1050, opened 2022-04-24
Don't call pocl_<device>_init_ops twice

s-barannikov:
If a device driver allocates resources in `pocl_<device>_init_ops`,
calling it twice can result in a memory leak.
Also, don't append the path separator twice when building the path
to the device's DSO.


pjaaskel:
LGTM. Can you add yourself to CREDITS if you are not there yet?

pjaaskel:
LGTM. Can you add yourself to CREDITS if you are not there yet?

PR #1055, opened 2022-05-14
Proof-of-concept cl_ext_device_side_abort impl. for CPU devices

pjaaskel:
https://github.com/parmance/OpenCL-Docs/blob/abort/extensions/cl_ext_device_side_abort.asciidoc


pjaaskel:
Fails a conformance test conformance_computeinfo

```
...
Device extension mismatch                                                                                                       
        Extensions only in numeric:
        Extensions only in string: cl_ext_device_side_abort
...
```

PR #1063, opened 2022-06-10
Vulkan fixes

franz:
Bunch of fixes to the vulkan driver, plus some (last 3) commits are not Vulkan related


jansol:
I'm not a huge fan of keeping generated binary blobs in git. AIUI it is not at all possible to build the vulkan device without having clspv? If that is correct I'd much rather have the .spv blobs generated as a build step.

franz:
It actually should be possible to build Vulkan without clspv (and once the driver is upgraded to use SPIR-V metadata it could be very useful) but i haven't tried for a while. We could add a build step but i saw little point in it, as SPIR-V is both target-independent and LLVM version independent.

PR #1067, opened 2022-06-21
cuda: Coarse-grained SVM

matthiasdiener:


matthiasdiener:
This is ready for a first look @inducer . It passes the test in https://github.com/inducer/pyopencl/blob/main/examples/svm.py

inducer:
We'll definitely also need `clEnqueueSVMFree`.

matthiasdiener:
> We'll definitely also need `clEnqueueSVMFree` (which is [`cuMemFreeAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf)).

As far as I can see, the other drivers don't handle `clEnqueueSVMFree` in a special way. Is the handling in https://github.com/pocl/pocl/blob/master/lib/CL/clEnqueueSVMFree.c sufficient?

inducer:
Fair point, I think the generic handler ought to suffice.

matthiasdiener:
With https://github.com/pocl/pocl/pull/1067/commits/229b754910d6174952c097bff66582fc248821ff, the [example](https://github.com/inducer/pyopencl/blob/24042536aed7b33765606e9d30fdf91196c0e746/examples/demo_array_svm.py) in https://github.com/inducer/pyopencl/pull/452  also appears to work, after modifying it to use coarse-grained SVM.

matthiasdiener:
This is ready for review.

pjaaskel (changes requested):
It looks good, but can you describe the improvement in the CHANGES file?

inducer:
> Fair point, I think the generic handler [for `clclEnqueueSVMFree`] ought to suffice.

Upon further thought, I don't think that's correct. IIUC, the generic handler will free the memory when the enqueue thread gets to that point in the queue, which is too soon, as it does not guarantee that kernels executing against that memory are complete. IMO, we need to override to use [`cuMemFreeAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf).

matthiasdiener:
> It looks good, but can you describe the improvement in the CHANGES file?

I added this in [bec4b4f](https://github.com/pocl/pocl/pull/1067/commits/bec4b4ffbb4372103829b8095c4fb527735909b0).

matthiasdiener:
> > Fair point, I think the generic handler [for `clclEnqueueSVMFree`] ought to suffice.
> 
> Upon further thought, I don't think that's correct. IIUC, the generic handler will free the memory when the enqueue thread gets to that point in the queue, which is too soon, as it does not guarantee that kernels executing against that memory are complete. IMO, we need to override it to use [`cuMemFreeAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf).

I see your point, but I'm just wondering why the other drivers don't encounter this issue?

matthiasdiener:
> > Fair point, I think the generic handler [for `clclEnqueueSVMFree`] ought to suffice.
> 
> Upon further thought, I don't think that's correct. IIUC, the generic handler will free the memory when the enqueue thread gets to that point in the queue, which is too soon, as it does not guarantee that kernels executing against that memory are complete. IMO, we need to override it to use [`cuMemFreeAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf).

I looked at this, but since we can not change the interface to the `pocl_*_svm_free` functions, it seems it is only possible to call `cuMemFreeAsync` with the default stream. Would implementing it like this make sense?

inducer:
> I looked at this, but since we can not change the interface to the `pocl_*_svm_free` functions,

How come we can't? IIUC that's an internal interface, right?

> it seems it is only possible to call `cuMemFreeAsync` with the default stream. Would implementing it like this make sense?

That would synchronize with the wrong thing, so no, I don't think that helps in any way.

matthiasdiener:
> > I looked at this, but since we can not change the interface to the `pocl_*_svm_free` functions,
> 
> How come we can't? IIUC that's an internal interface, right?

No, since these functions are called from the generic CL functions, their interface is fixed. See e.g. https://github.com/pocl/pocl/blob/03226ffe1ab453422a13799d64e32c07d6a08a6d/lib/CL/devices/prototypes.inc#L174 for the prototype of `svm_free`.


> > it seems it is only possible to call `cuMemFreeAsync` with the default stream. Would implementing it like this make sense?
> 
> That would synchronize with the wrong thing, so no, I don't think that helps in any way.

Ok, that was my suspicion too.

inducer:
> No, since these functions are called from the generic CL functions, their interface is fixed.

I still don't understand: All call sites appear to be within pocl, so I don't see a big issue with changing the interface?

matthiasdiener:
> > No, since these functions are called from the generic CL functions, their interface is fixed.
> 
> I still don't understand: All call sites appear to be within pocl, so I don't see a big issue with changing the interface?

We could change the interface to include something like the queue, but would have to change all drivers to this new interface then. I can do it if that sounds reasonable.

matthiasdiener:
> > > Fair point, I think the generic handler [for `clclEnqueueSVMFree`] ought to suffice.
> > 
> > 
> > Upon further thought, I don't think that's correct. IIUC, the generic handler will free the memory when the enqueue thread gets to that point in the queue, which is too soon, as it does not guarantee that kernels executing against that memory are complete. IMO, we need to override it to use [`cuMemFreeAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf).

79a33a29 should implement what you suggested, and keep the CL semantics, ie, clSVMFree frees the memory immediately, while clEnqueueSVMFree enqueues the free in the stream.

pjaaskel:
Is this one ready?

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -1440,6 +1459,53 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq, int locked)
...
+                    break;
+                  }
+              }
+              POCL_UNLOCK_OBJ (event->context);
+              assert (item);
+              POCL_MEM_FREE (item);
+              // Leads to 'undefined symbol: POclReleaseContext'
+              // POname (clReleaseContext) (event->context);
+
+              dev->ops->svm_free (dev, ptr);
```
inducer, line 1505: This free should be stream-synchronized.

matthiasdiener: To add a bit more info, we tried to stream-synchronize the free via `cuMemFreeAsync` (https://github.com/pocl/pocl/pull/1067/commits/bf1642ea34db1236e7e8796d3ba0a09c3a11f94b), but it turns out `cuMemFreeAsync` does not work with managed allocations (it returns `CUDA_ERROR_NOT_SUPPORTED`). 

We could do 2 things:

- Synchronize in a different way (how?)
- Don't use managed memory

inducer: Since we're only looking to have coarse-grain buffer SVM on Nvidia, I think moving away from managed memory might make sense.

franz:
I've rebased this on master, did some changes (i've changed it to report fine-grained SVM, to support coarse-grained only it'd have to implement `svm_map` & `svm_unmap` device ops), and did some basic testing. If nobody has objections, i would somehow make the clEnqueueSVMFree either return an error or abort (with CUDA), and merge the PR, since it's quite useful even without  `clEnqueueSVMFree`.

WRT "I'm just wondering why the other drivers don't encounter this issue?" - the way other drivers use PoCL's event system, the svm_free event (or any other event really) will only enter the submitted state once all its dependencies have finished executing, so the event is safe to execute at that point. Unfortunately the CUDA driver has its own, more complicated handling of events, and i haven't really examined the code in detail, so i cannot say if it's safe WRT svm_free or not.

pjaaskel:
I don't see a reason not to pull this. Please improve in further PRs. We need this to get chipStar running to run CUDA/HIP on CUDA via PoCL-CUDA :bug:

PR #1069, opened 2022-07-01
`clEnqueueSVMMemFree`: copy pointers

inducer:
The spec for [`clEnqueueSVMMemFree`](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#clEnqueueSVMFree) says:

> The memory associated with `svm_pointers` can be reused or freed after the function returns.

AFAICT, the current code in pocl simply holds onto the memory passed in by the user, which I believe is not conformant behavior. This change makes a copy instead.

cc @matthiasdiener

### DIFF
--- a/lib/CL/clEnqueueSVMFree.c
+++ b/lib/CL/clEnqueueSVMFree.c
@@ -60,11 +60,11 @@ POname(clEnqueueSVMFree) (cl_command_queue command_queue,
   if (errcode != CL_SUCCESS)
     return errcode;
 
-  const size_t svm_pointers_nbytes = sizeof(void *) * num_svm_pointers;
-  void *svm_pointers_copy = malloc(svm_pointers_nbytes);
+  const size_t svm_pointers_nbytes = sizeof (void *) * num_svm_pointers;
+  void *svm_pointers_copy = malloc (svm_pointers_nbytes);
   if (svm_pointers_copy == NULL)
     return CL_OUT_OF_HOST_MEMORY;
-  memcpy(svm_pointers_copy, svm_pointers, svm_pointers_nbytes);
+  memcpy (svm_pointers_copy, svm_pointers, svm_pointers_nbytes);
 
   _cl_command_node *cmd = NULL;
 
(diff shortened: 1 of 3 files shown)


matthiasdiener:
Is this already done here?

https://github.com/pocl/pocl/blob/4627171d40543091e399989c277faa52fcee0ff8/lib/CL/devices/common.c#L630-L659

inducer:
> Is this already done here?

I'm not sure I understand the question. The location you point runs when the command is executed. At that point, the pointers that the application has passed to `clEnqueueSVMMemFree` are likely long gone.

matthiasdiener:
> > Is this already done here?
> 
> I'm not sure I understand the question. The location you point runs when the command is executed. At that point, the pointers that the application has passed to `clEnqueueSVMMemFree` are likely long gone.

Ah, I see, you're right.

inducer:
Based on @franz's comment [here](https://github.com/pocl/pocl/pull/1070#issuecomment-1173644569), would it be better to free the pointer memory in `pocl_free_event_node`?

matthiasdiener:
> Based on @franz's comment [here](https://github.com/pocl/pocl/pull/1070#issuecomment-1173644569), would it be better to free the pointer memory in `pocl_free_event_node`?

For `clEnqueueSVMMigrateMem`, which also needs to copy its `svm_pointers` argument, the memory of the copy is also freed in `pocl_free_event_node`.

pjaaskel:
Looked good and worked with my local test envs, so I pushed. Thanks! Although I wonder are there test cases that hit this code path in the repo now, are you aware of any?

matthiasdiener:
> Looked good and worked with my local test envs, so I pushed. Thanks! Although I wonder are there test cases that hit this code path in the repo now, are you aware of any?

Thanks for merging! I'm not aware of any test cases that show this issue in the pocl repo, but we have seen lots of failures in pyopencl without this fix (especially while working on https://github.com/inducer/pyopencl/pull/452).

PR #1071, opened 2022-07-04
atomic

anbe42:
Some platforms (e.g. mipsel) need to link with -latomic for 64-bit atomics.
Switch to the modern __atomic_*() builtins (available from gcc 4.7 onwards) that are backed by -latomic (from gcc) if needed.
Add a cmake test and link with -latomic if needed.
If there is any supported compiler that does not have __atomic_*(), I can add a fallback for __sync_*()


pjaaskel:
Ping.

pjaaskel (approved):

PR #1086, opened 2022-08-10
clang-14: handle deprecated warnings for opaque ptr

trixirt:
clang is moving to typeless, opaque pointers
clang-14 warns about some deprecations.  These are a subset of
what must be done for clang-15. So clean them up in 14.

see
https://llvm.org/docs/OpaquePointers.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1093,6 +1093,39 @@ else()
   set(HOST_DEVICE_CL_VERSION_MINOR 2)
 endif()
 
+#
+# LLVM_OPAQUE_POINTERS
+#
+# llvm has changed from typed pointers to opaque pointers, their guidance is
+#
+# From https://llvm.org/docs/OpaquePointers.html, version support
+# LLVM 14: Supports all necessary APIs for migrating to opaque pointers and deprecates/removes
+#          incompatible APIs. However, using opaque pointers in the optimization pipeline is not fully
+#          supported. This release can be used to make out-of-tree code compatible with opaque
+#          pointers, but opaque pointers should not be enabled in production.
+#
+# LLVM 15: Opaque pointers are enabled by default. Typed pointers are still available, but only
+#          supported on a best-effort basis and may be untested.
+#
+# LLVM 16: Only opaque pointers will be supported. Typed pointers will not be supported.
+#
+# Pre llvm 14, ENABLE_LLVM_OPAQUE_POINTERS will not be defined and the legacy typed pointers will be used.
+# Post llvm 15, ENABLE_LLVM_OPAQUE_POINTERS will not be defined and opaque pointers will be used.
+# For llvm 14 and 15 this macro will be controlled by the build parameter ENABLE_LLVM_OPAQUE_POINTERS
+# For llvm 14 it will default to OFF
+# For llvm 15 it will default to  ON
+if(LLVM_VERSION VERSION_LESS_EQUAL 13.0)
+  set(ENABLE_LLVM_OPAQUE_POINTERS 0)
+elseif(LLVM_VERSION VERSION_EQUAL 14.0)
+  option(ENABLE_LLVM_OPAQUE_POINTERS "Handle the change to llvm opaque pointers." OFF)
+elseif(LLVM_VERSION VERSION_EQUAL 15.0)
+  option(ENABLE_LLVM_OPAQUE_POINTERS "Handle the change to llvm opaque pointers." ON)
+else()
+  set(ENABLE_LLVM_OPAQUE_POINTERS 1)
+endif()
+if (ENABLE_LLVM_OPAQUE_POINTERS)
+  set(LLVM_OPAQUE_POINTERS 1)
+endif()
 
 # These are mandatory for OpenCL 1.1 (except for 3D writes but those are CPU-independent)
 set(HOST_DEVICE_EXTENSIONS "cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics \
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -263,3 +263,5 @@
 
 
 #cmakedefine USE_POCL_MEMMANAGER
+
+#cmakedefine LLVM_OPAQUE_POINTERS
--- a/lib/llvmopencl/LLVMUtils.h
+++ b/lib/llvmopencl/LLVMUtils.h
@@ -135,4 +135,9 @@ void CloneFunctionIntoAbs(llvm::Function *NewFunc,
 #endif
 }
 
+#ifndef LLVM_OPAQUE_POINTERS
+// Globals
+#define getValueType getType()->getElementType
+#endif /* LLVM_OPAQUE_POINTERS */
+
 #endif
--- a/lib/llvmopencl/LLVMUtils.h
+++ b/lib/llvmopencl/LLVMUtils.h
@@ -140,4 +140,8 @@ void CloneFunctionIntoAbs(llvm::Function *NewFunc,
 #define getValueType getType()->getElementType
 #endif /* LLVM_OPAQUE_POINTERS */
 
+#ifdef LLVM_OLDER_THAN_14_0
+#define LLVMBuildGEP2(A, B, C, D, E, F) LLVMBuildGEP(A, C, D, E, F)
+#endif
+
 #endif
--- a/lib/llvmopencl/Workgroup.cc
+++ b/lib/llvmopencl/Workgroup.cc
@@ -1280,7 +1280,8 @@ LLVMValueRef Workgroup::createArgBufferLoad(LLVMBuilderRef Builder,
   LLVMValueRef Offs =
       LLVMConstInt(LLVMInt32TypeInContext(LLVMContext), ArgPos, 0);
   LLVMValueRef ArgByteOffset =
-      LLVMBuildGEP(Builder, ArgBufferPtr, &Offs, 1, "arg_byte_offset");
+      LLVMBuildGEP2(Builder, LLVMTypeOf(ArgBufferPtr), ArgBufferPtr, &Offs, 1,
+                    "arg_byte_offset");
 
   llvm::Argument &Arg = cast<Argument>(*unwrap(Param));
 
@@ -1416,7 +1417,8 @@ Workgroup::createArgBufferWorkgroupLauncher(Function *Func,
         uint64_t ArgPos = ArgBufferOffsets[i];
         LLVMValueRef Offs = LLVMConstInt(Int32Type, ArgPos, 0);
         LLVMValueRef SizeByteOffset =
-            LLVMBuildGEP(Builder, ArgBuffer, &Offs, 1, "size_byte_offset");
+            LLVMBuildGEP2(Builder, LLVMTypeOf(ArgBuffer), ArgBuffer, &Offs, 1,
+                          "size_byte_offset");
         LLVMTypeRef DestTy = LLVMPointerType(ParamIntType, 0);
         LLVMValueRef SizeOffsetBitcast =
             LLVMBuildPointerCast(Builder, SizeByteOffset, DestTy, "size_ptr");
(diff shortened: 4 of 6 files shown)


pjaaskel (changes requested):
Thanks! Unfortunately this breaks the TCE/OpenASIP driver since OA is not ported yet to cope with the opaque pointers change. I think minimally we need to use the opaque pointers disabling flags when we call opt in tcecc. 

If you want to get this pulled before Someone finds the time to fix TCE, can you add another cmake macro that keeps the current typed behavior for LLVM 14 so I can use it to avoid regressing my test envs for now?

trixirt:
I will add a macro
I am really looking at llvm 15 which causes many more similar problems.   In 15 some of the functions are still available but assert at runtime because they have been hallowed out. so 15 is a mess. 14 is the crossover when the functionality was still available in both.  so my next steps were to find the problems in 15 and fix them in 14.

CMakeLists.txt
```diff
@@ -1093,6 +1093,39 @@ else()
...
+#
+# LLVM 16: Only opaque pointers will be supported. Typed pointers will not be supported.
+#
+# Pre llvm 14, ENABLE_LLVM_OPAQUE_POINTERS will not be defined and the legacy typed pointers will be used.
+# Post llvm 15, ENABLE_LLVM_OPAQUE_POINTERS will not be defined and opaque pointers will be used.
+# For llvm 14 and 15 this macro will be controlled by the build parameter ENABLE_LLVM_OPAQUE_POINTERS
+# For llvm 14 it will default to OFF
+# For llvm 15 it will default to  ON
+if(LLVM_VERSION VERSION_LESS_EQUAL 14.0)
+  set(LLVM_OPAQUE_POINTER 0)
```
pjaaskel, line 1118: Is this intentional? Should it be "ENABLE_OPAQUE_POINTERS 0"?

CMakeLists.txt
```diff
@@ -1093,6 +1093,39 @@ else()
...
+elseif(LLVM_VERSION VERSION_EQUAL 14.0)
+  option(ENABLE_LLVM_OPAQUE_POINTERS "Handle the change to llvm opaque pointers." OFF)
+elseif(LLVM_VERSION VERSION_EQUAL 15.0)
+  option(ENABLE_LLVM_OPAQUE_POINTERS "Handle the change to llvm opaque pointers." ON)
+else()
+  set(ENABLE_LLVM_OPAQUE_POINTERS 1)
+endif()
+if (ENABLE_LLVM_OPAQUE_POINTERS)
+  add_definitions(-DLLVM_OPAQUE_POINTERS)
```
pjaaskel, line 1127: Is there a need to pass this as a cpp flag instead of using the config.h?

pjaaskel (changes requested):

lib/llvmopencl/Workgroup.cc
```diff
@@ -1411,15 +1430,20 @@ Workgroup::createArgBufferWorkgroupLauncher(Function *Func,
...
+        LLVMValueRef SizeByteOffset =
+            LLVMBuildGEP2(Builder, ArgBufferTy, ArgBuffer, &Offs, 1, "size_byte_offset");
+#endif
         LLVMTypeRef DestTy = LLVMPointerType(ParamIntType, 0);
         LLVMValueRef SizeOffsetBitcast =
             LLVMBuildPointerCast(Builder, SizeByteOffset, DestTy, "size_ptr");
 
         // The buffer size passed from the runtime is a byte size, we
         // need to convert it to an element count for the alloca.
-#ifdef LLVM_OLDER_THAN_14_0
```
pjaaskel, line 1422: Same here.

lib/llvmopencl/Workgroup.cc
```diff
@@ -1298,7 +1315,7 @@ LLVMValueRef Workgroup::createArgBufferLoad(LLVMBuilderRef Builder,
     LLVMTypeRef DestTy = LLVMPointerType(ParamType, DeviceArgsASid);
     LLVMValueRef ArgOffsetBitcast =
         LLVMBuildPointerCast(Builder, ArgByteOffset, DestTy, "arg_ptr");
-#ifdef LLVM_OLDER_THAN_14_0
+#ifndef LLVM_OPAQUE_POINTERS
```
pjaaskel, line 1318: This breaks if LLVM_OPAQUE_POINTERS undefined and LLVM 14? Produces opaque pointers?

pjaaskel (changes requested):
Please check the macros logic. It still produces opaque pointers with LLVM 14 and fails TCE even if it undefs LLVM_OPAQUE_POINTERS.

pjaaskel:
Getting there. It now works with my test envs down to LLVM 9.0 where it fails due to:
```
lib/llvmopencl/linker.cpp:84:31: error: ‘class llvm::Function’ has no member named ‘getArg’
```

It doesn't yet have that API. I don't know how long we will care to keep up these older LLVM versions, but likely still for the 3.1 release.

trixirt:
I am thinking looping over args can be handled better, here's where i am going so the #idef  opaque ptr can be hidden in the macro.
https://github.com/trixirt/pocl/commit/e9a23b52d77449908a538b05545edd821844cee1

pjaaskel:
> I see this error so I will handle. Handling these old releases is becoming more difficult, it would be good if the oldest (6) could be knocked off when the newest (15) is enabled. Could we knock off 6 ?

In fact, it seems I only test down to LLVM 7 now ny pre-push testing script, so we have officially dropped LLVM 6 support already.  I was thinking to drop all but the latest 2-3 releases in the future unless someone steps up to test&maintain the older ones.

trixirt:
I am in favor of doing just the recent llvms 
Fedora and RHEL are always moving to the latest llvm, that's 14 now and why I am doing these changes.  For the older llvm's, the llvm project does not have great support for the older releases, so if  was a problem, would they take the fix ?

pjaaskel:
> For the older llvm's, the llvm project does not have great support for the older releases, so if was a problem, would they take the fix ?

No they won't. But it's more about potential private off-tree drivers that utilize PoCL code base which are stuck with older LLVM versions just because they haven't seen other reasons to upgrade their toolchain. Anyhow, I don't think we should be bothered by those too much, so let's do the version pruning after the next release and see if someone steps up.

pjaaskel:
Is this branch ready? TCE through LLVM 14 fails again due to opaque pointers getting through. LLVM 7 works.

trixirt:
AFAIK it is ready but I have no idea how to test the TCE side.  Where is the TCE toolchain etc. coming from ?

config.h.in.cmake
```diff
@@ -263,3 +263,5 @@
 
 
 #cmakedefine USE_POCL_MEMMANAGER
+
+#cmakedefine LLVM_OPAQUE_POINTERS
```
pjaaskel, line 267: Does it always define LLVM_OPAQUE_POINTERS (to 0 or 1)? Just a wild guess what could cause LLVM 14 getting opaque ptrs in to the TCE target.

lib/llvmopencl/LLVMUtils.h
```diff
@@ -135,4 +135,13 @@ void CloneFunctionIntoAbs(llvm::Function *NewFunc,
 #endif
 }
 
+#ifndef LLVM_OPAQUE_POINTERS
```
pjaaskel, line 138: ...then this would always get skipped.

pjaaskel (changes requested):
OpenASIP (TCE) can be installed via http://openasip.org/. Anyhow, I think the easiest way to check if LLVM 14 gets opaque ptrs generated by default is to dump the parallel.bc from the compiler temp dirs. The user manual lists envs for leaving the compiler temp bc for inspection then you can check with llvm-dis. parallel.bc is the last LLVM IR output before target code gen.

lib/llvmopencl/LLVMUtils.h
trixirt: These changes are only for the api deprecations,  I _think_ TCE issue is with code generation and would need a different change, likely an arg passed in the build command.  but this problem would exist in the current code, so I am wondering why it hasn't shown up ..

pjaaskel:
Some strange indeed with the arg loads. The BT:
```
#3  0x00007ffff7bd7fd6 in __assert_fail () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007fffeee5cc8c in llvm::GetElementPtrInst::Create(llvm::Type*, llvm::Value*, llvm::ArrayRef<llvm::Value*>, llvm::Twine const&, llvm::Instruction*) () from /home/visit0r/local/stow/llvmtce-14/lib/libLLVMTCE-14.so
#5  0x00007fffeee89f39 in LLVMBuildGEP2 () from /home/visit0r/local/stow/llvmtce-14/lib/libLLVMTCE-14.so
#6  0x00007ffff7b3650e in pocl::Workgroup::createArgBufferLoad (this=0x7fffe002d490, Builder=0x7fffe00e23f0, ArgBufferPtr=0x7fffe00a6f30, ArgBufferOffsets=0x7fffe55c9fd0, F=<optimized out>, ParamIndex=<optimized out>)
    at /home/visit0r/src/pocl/lib/llvmopencl/Workgroup.cc:1284
#7  0x00007ffff7b453ff in createArgBufferWorkgroupLauncher (KernName=..., Func=<optimized out>, this=<optimized out>) at /home/visit0r/src/pocl/lib/llvmopencl/Workgroup.cc:1462
#8  pocl::Workgroup::runOnModule(llvm::Module&) () at /home/visit0r/src/pocl/lib/llvmopencl/Workgroup.cc:269
```
This is the strange part (popped to stack frame 6):
```
(gdb) list
1279
1280      uint64_t ArgPos = ArgBufferOffsets[ParamIndex];
1281      LLVMValueRef Offs =
1282          LLVMConstInt(LLVMInt32TypeInContext(LLVMContext), ArgPos, 0);
1283      
1284      LLVMValueRef ArgByteOffset =
1285          LLVMBuildGEP2(Builder, LLVMTypeOf(ArgBufferPtr), ArgBufferPtr, &Offs, 1,
1286                        "arg_byte_offset");
1287
1288      llvm::Argument &Arg = cast<Argument>(*unwrap(Param));
(gdb) call LLVMDumpValue(ArgBufferPtr)
i8 addrspace(1)* %0$8 = 0
(gdb) call LLVMDumpType(ArgBufferPtr)
half$9 = 0

```

lib/llvmopencl/Workgroup.cc
```diff
@@ -1276,7 +1280,8 @@ LLVMValueRef Workgroup::createArgBufferLoad(LLVMBuilderRef Builder,
   LLVMValueRef Offs =
       LLVMConstInt(LLVMInt32TypeInContext(LLVMContext), ArgPos, 0);
   LLVMValueRef ArgByteOffset =
-      LLVMBuildGEP(Builder, ArgBufferPtr, &Offs, 1, "arg_byte_offset");
+      LLVMBuildGEP2(Builder, LLVMTypeOf(ArgBufferPtr), ArgBufferPtr, &Offs, 1,
```
TopiLeppanen, line 1283: Changing the Workgroup.cc:s LLVMBuildGEP2's second argument to LLVMGetElementType(LLVMTypeOf(ArgBufferPtr)), or alternatively to LLVMInt8TypeInContext(LLVMContext) seems to fix the type mismatch assertions @pjaaskel reported.

lib/llvmopencl/Workgroup.cc
```diff
@@ -1412,7 +1417,8 @@ Workgroup::createArgBufferWorkgroupLauncher(Function *Func,
         uint64_t ArgPos = ArgBufferOffsets[i];
         LLVMValueRef Offs = LLVMConstInt(Int32Type, ArgPos, 0);
         LLVMValueRef SizeByteOffset =
-            LLVMBuildGEP(Builder, ArgBuffer, &Offs, 1, "size_byte_offset");
+            LLVMBuildGEP2(Builder, LLVMTypeOf(ArgBuffer), ArgBuffer, &Offs, 1,
```
TopiLeppanen, line 1420: This one might not get tested by our tests, but perhaps should be changed as well similarly to the above one

PR #1094, opened 2022-09-07
Fix runtime dependant loss of precision with cuda event timers

MMittenbuehler:
I wanted to run a benchmark with PoCL (CUDA backend) and OpenCL with our kernels, however, the PoCL event timers seemed to be quite off. In a micro-benchmark, their resolution was quite grainy and sometimes even returned zero runtimes.

Upon closer inspection, it seems that the CUDA backend still suffers from a loss of precision depending on how long the program is running. The issue is that CUDA utilizes float to measure the time between events in milliseconds. 
Our device has a resolution of 1us. After just 9s of runtime, the precision drops to 2us due to the limited resolution of floats. Usually, this is not an issue, however, the way the timers in the CUDA backend work is by creating one epoch event at the start of the program and then measuring the time of other events (e.g. start and end time of a kernel) relative to this epoch event. Therefore, the timer resolution is determined by the total runtime of the program and not just of the kernel. This makes micro-benchmarks problematic. 

My suggestion would be to measure the start time of a kernel relative to the epoch event as it is currently done to keep absolute time. However, I would measure the end time of a kernel relative to its start time so that we do not lose precision there. As the numbers are converted into longs representing ns, this should increase the overall precision of the timers.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -1593,11 +1593,12 @@ pocl_cuda_update_event (cl_device_id device, cl_event event)
        * only the elapsed time between two events. We use the elapsed
        * time from the epoch event enqueued on device creation to get
        * the actual timestamps.
-       *
-       * Since the CUDA timer resolution is lower than the host timer,
-       * this can sometimes result in the start time being before the
-       * submit time, so we use max() to ensure the timestamps are
-       * sane. */
+       * More specifically, we measure the time between the epoch event
+       * and the start event. This results in the start time. Then we
+       * take the elapsed time between the start and end event to
+       * compute the end time. Measuring the end time relative to the
+       * epoch event may result in unprecise measurements due to the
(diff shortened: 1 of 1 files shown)


Oblomov:
If we're fixing this, I would recommend getting rid of the epoch time altogether, compute all times as deltas from the most recently created event.

pjaaskel:
Seems no one has contributed improvements to this PR, so I ask is this state an improvement that I should pull?

nchristensen:
I no longer see kernel execution times of 1e-9 with this pull request so I think it is an improvement.

MMittenbuehler:
Sorry for not responding for a while. It currently fixes the issue if we want to measure the time for one kernel.

@Oblomov is correct that this does not fix the problem completely. For example, if we have kernels A and B, and we measure the time between the start of A and the end of B, we currently may lose precision due to the epoch timer.
I tried to implement his suggestion but did not get it to work.

pjaaskel:
Can you update CHANGES with this fix? Thanks!

PR #1096, opened 2022-09-28
SPIR-V fixes

franz:
Improvements to SPIR-V support mainly for x86-64 CPU driver, and  also for CUDA (though that is untested).

The SPIR-V subtest from the OpenCL-CTS is now down to 7 failing tests out of 192 (previously 25). These fails are because of issues that need fixing in the linker.


cmake/bitcode_rules.cmake
```diff
@@ -138,10 +138,11 @@ function(compile_sleef_c_to_bc EXT FILENAME SUBDIR BCLIST)
 endfunction()
 
 
-function(compile_ll_to_bc FILENAME SUBDIR BC_FILE_LIST)
+function(compile_ll_to_bc FILENAME SUBDIR BCLIST)
     get_filename_component(FNAME "${FILENAME}" NAME)
     set(BC_FILE "${CMAKE_CURRENT_BINARY_DIR}/${SUBDIR}/${FNAME}.bc")
-    set(${BC_FILE_LIST} ${${BC_FILE_LIST}} ${BC_FILE} PARENT_SCOPE)
+    list(APPEND ${BCLIST} "${BC_FILE}")
+    set(${BCLIST} ${${BCLIST}} PARENT_SCOPE)
```
jansol, line 145: Can you add a short comment explaining what is happening here?

lib/CL/clCreateProgramWithIL.c
```diff
@@ -28,6 +28,73 @@
 #include "pocl_util.h"
 #include "pocl_shared.h"
 
+static int
+get_program_spec_constants (cl_program program, char *program_bc_spirv)
+{
+  char *args[] = { LLVM_SPIRV, "--spec-const-info", program_bc_spirv, NULL };
+  char captured_output[8192];
+  size_t captured_bytes = 8192;
```
jansol, line 36: Please use constant variables or defines for magic numbers.

lib/CL/clCreateProgramWithIL.c
```diff
@@ -28,6 +28,73 @@
...
+                                             args);
+  POCL_RETURN_ERROR_ON ((errcode != 0), CL_INVALID_BINARY,
+                        "External command (llvm-spirv translator) failed!\n");
+
+  captured_output[captured_bytes] = 0;
+  char *lines[4096];
+  unsigned num_lines = 0;
+  char delim[2] = { 0x0A, 0x0 };
+  char *token = strtok (captured_output, delim);
+  while (num_lines < 4096 && token != NULL)
```
jansol, line 49: Ditto re: magic numbers.

lib/kernel/host/CMakeLists.txt
```diff
@@ -413,7 +407,18 @@ function(compile_sleef VARIANT SLEEF_CONFIG SLEEF_CONFIG_NEW SLEEF_BC)
...
 
-  foreach(VECSIZE "128" "256" "512")
+  # for X86-64, SPIR wrapper is compiled based on largest supported vector size
+  if(HOST_DEVICE_ADDRESS_BITS EQUAL 32)
+    set(SPIR_WRAPPER_FILE "host/spir_wrapper32.ll")
+  else()
+    if(NOT X86)
+      set(SPIR_WRAPPER_FILE "host/spir_wrapper64.ll")
+    else()
+      set(SPIR_WRAPPER_FILE "")
```
jansol, line 417: This could have a comment mentioning the fallback that is applied later in the code?

lib/llvmopencl/linker.cpp
```diff
@@ -159,12 +140,13 @@ CloneFuncFixOpenCLImageT(llvm::Module *Mod, llvm::Function *F, unsigned AS)
 
     return DstFunc;
 }
+*/
```
jansol, line 143: Stray end of comment marker?

jansol (changes requested):
Not very familiar with these parts but nothing jumped out as particularly weird. Some small nits though.

lib/CL/devices/common_driver.c
```diff
@@ -449,13 +453,97 @@ pocl_driver_svm_fill (cl_device_id dev, void *__restrict__ svm_ptr,
...
+static int
+pocl_regen_spirv_binary (cl_program program, cl_uint device_i)
+{
+#ifdef LLVM_SPIRV
+  int errcode = CL_SUCCESS;
+
+  int spec_constants_changed = 0;
+
+  for (unsigned i = 0; i < program->num_spec_consts; ++i)
+    spec_constants_changed += program->spec_const_is_set[i];
```
pjaaskel, line 468: Does this regenerate when the constant is set but to a different value? Should we just add the specialization constants to the hash?

lib/kernel/host/CMakeLists.txt
```diff
@@ -447,6 +455,8 @@ function(compile_sleef VARIANT SLEEF_CONFIG SLEEF_CONFIG_NEW SLEEF_BC)
 
   endforeach()
 
+  compile_ll_to_bc("${SPIR_WRAPPER_FILE}" "${VARIANT}" BC_FILE_LIST)
+
   file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}")
   set(SLEEF_BC "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}/sleef.bc")
   set(SLEEF_BC "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}/sleef.bc" PARENT_SCOPE)
```
pjaaskel, line 462: Do we want to put the generated wrappers to the repository? I think not if we can always regenerate them using the script.

lib/llvmopencl/linker.cpp
```diff
@@ -445,18 +418,25 @@ int link(llvm::Module *Program, const llvm::Module *Lib, std::string &log,
...
       DB_PRINT("%s is not defined\n", fi->getName().data());
-      declared.push_back(fi->getName());
+      DeclaredFunctions.insert(fi->getName());
       continue;
     }
 
+    // anonymous functions have no name, which breaks the algorithm later
+    // when it searches for undefined functions in the kernel library.
+    // assign a randomized name here
+    std::string temp = std::to_string(UniDist(Mersenne));
```
pjaaskel, line 430: You might not need this, I think setName appends a running number if there's already one (not sure)?

franz: Ok, i'll take a look at setName.

lib/CL/devices/common_driver.c
franz: Yes, the current value (last set) is used at every clBuildProgram. If the user sets the constant and doesn't call clBuildProgram then it will use the old value, but that's the user's fault. Not sure about adding to the build hash. It should work without as is, but in a sense these are like build options, and IIRC we add the build options to the build hash of non-spirv programs, so it probably should be included.

lib/kernel/host/CMakeLists.txt
franz: The wrappers are independent of the build host so there's not much point regenerating, but it's true they're quite large. I can remove them and add a custom command to generate it. It adds build dependency on python3 but that should be fine.

lib/CL/devices/common_driver.c
pjaaskel: That way the specialized SPIR-Vs and associated generated codes would get cached correctly?

lib/kernel/host/CMakeLists.txt
pjaaskel: Right. The benefit is that we ensure the generator doesn't bitrot, stays functional.

lib/CL/clCreateProgramWithIL.c
franz: fixed

cmake/bitcode_rules.cmake
franz: Added

lib/kernel/host/CMakeLists.txt
franz: done

lib/llvmopencl/linker.cpp
franz: yep, removed.

lib/llvmopencl/linker.cpp
franz: Yep, it seems setName automatically makes the name unique. Fixed

pjaaskel:
With LLVM 8.0 (and probably older):
```
/home/visit0r/local/stow/llvm-8.0-tcepatched/bin/llvm-as: /home/visit0r/src/pocl/test-pr-cmake-llvm-8.0-build/lib/kernel/host/broadwell/spir_wrapper_256bit.ll:5053:60: error: expected ')' at end of argument list
declare <16 x float> @_Z8_cl_acosDv16_f(<16 x float>* byval(<16 x float>) align 64) local_unnamed_addr #0
```
So, let's postpone merging this in until after the 3.1 is branched and we will drop older than LLVM 10 support.

jansol:
Should we cherry-pick the CMake commit in beforehand?

PR #1104, opened 2022-10-18
fix(lib/CL/clReleaseCommandBufferKHR)

yingmanwumen:
A spelling mistake at the end of this file causes a compile error.


jansol (approved):
Weird, for some reason my compiler accepted this -- but only in this particular file.

PR #1110, opened 2022-10-28
TCE driver fixes

TopiLeppanen:
Two commits:
1. Fix the tce-driver to not call "bare" tcecc, but instead propagate the binary name from CMake.
2. Add old timestamping code to update_event callback. This is an old feature that seems to have not been propagated to here yet.

### DIFF
--- a/lib/CL/devices/tce/tce_common.cc
+++ b/lib/CL/devices/tce/tce_common.cc
@@ -267,12 +267,12 @@ TCEString TCEDevice::tceccCommandLine(_cl_command_run *run_cmd,
   /* Compile in steps to save the program.bc for automated exploration 
      use case when producing the kernel capture scripts. */
   TCEString cmdLine;
-  cmdLine << "tcecc -llwpr " + poclIncludePathSwitch + " " + deviceMainSrc +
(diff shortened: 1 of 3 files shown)


lib/CL/devices/tce/tce_common.cc
```diff
@@ -64,6 +64,10 @@ using namespace TTAMachine;
 
 //#define DEBUG_TTA_DRIVER
 
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+#define TCECC_PATH TOSTRING(TCECC)
```
jansol, line 69: I think I'd prefer having this handled by a `#cmakedefine` in one of the config headers. And maybe call it TCECC_EXECUTABLE or something, since from the usage it seems to refer to the actual binary, not just a toolchain directory.

pjaaskel: And maybe convert the naming (at least gradually) to openasip / oacc.

TopiLeppanen: Thanks! Moved this to config.h.in
Also changed the name to OACC_EXECUTABLE

nrother:
The event timestamps look plausible from my testing. There is still some variation of +/-10k cycles, but that is expected because the simulator is free running in another thread.

PR #1111, opened 2022-10-28
TCE: standalone mode bug fix

TopiLeppanen:
fix a bug where standalone global buffers were written to AS 0 instead of global


jansol (commented):
Do the pointer arithmetic intermediate values not need the `__global__` qualifier? I.e. `kernelargs` and `temp`

PR #1117, opened 2022-11-07
Fix internal tests to work with multiple devices

TopiLeppanen:
Many tests assumed that the platform can only have 1 device. This PR fixes them so that it is now possible to run e.g.
`POCL_DEVICES="pthread pthread" make check`
and the tests pass.

Also removed CLBlast from tier1, and updated documentation on tier1 testsuites.

Added _internal_ label to some short-running tests that had no good reason to omit it. (Found by ctest -LE internal).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1759,7 +1759,7 @@ if(ENABLE_TESTS)
 
     # make check & make check_tier1
     add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} "--output-on-failure" -j ${CORECOUNT} ${COMMAND_USES_TERMINAL})
-    add_custom_target(check_tier1 COMMAND ${CMAKE_CTEST_COMMAND} "--output-on-failure" -L "'internal|piglit|PyOpenCL|conformance_suite_micro|CLBlast|shoc'" -j ${CORECOUNT} ${COMMAND_USES_TERMINAL})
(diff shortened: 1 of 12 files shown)


franz:
Couple of issues...

The assertion in the first commit - that prevents multiple devices from loading programs of other types than source - is actually there for a good reason. `POCL_DEVICES="pthread pthread"`  but for multiple devices of different type, it will fail, because if you create a program with binaries, you need a separate binary for each device type. But ofc using multiple pthread/basic devices is useful. Can you add a check that checks all devices have the same name?

Another is more a question: how is CLBlast "unstable" ?

TopiLeppanen:
> The assertion in the first commit - that prevents multiple devices from loading programs of other types than source - is actually there for a good reason. `POCL_DEVICES="pthread pthread"` but for multiple devices of different type, it will fail, because if you create a program with binaries, you need a separate binary for each device type. But ofc using multiple pthread/basic devices is useful. Can you add a check that checks all devices have the same name?

Thanks! That's a good point that I didn't think of. Added the suggested fix.

> Another is more a question: how is CLBlast "unstable" ?

@pjaaskel asked me to remove it and gave this as the reason. Personally I don't know anything about it.

pjaaskel:
@franz I disabled CLblast from my tier1 testing after observing it had race issues (IIRC). But it's years ago, don't know if they have been fixed or not.

franz:
@pjaaskel ah ok. I haven't observed any but maybe they're still there. Yeah it's fine. LGTM.

PR #1119, opened 2022-11-08
swap calloc args in enqueueMigrateMemObjects

RABijl:
small fix to prevent potential padding from messing things up


pjaaskel:
I don't understand what swapping the arguments changes here, pls explain.

RABijl:
from the description of calloc: 
```
extern void *calloc(size_t __nmemb, size_t __size)
Allocate NMEMB elements of SIZE bytes each, all initialized to 0
```
this is just to prevent bugs with alignment

RABijl:
fair point, and in theory the end result should be the same. However, we don't know what different implementations of stdlib will do with the information we are providing (i.e. nmemb and size).

pjaaskel:
Right, so what's the point of the patch then? If you want to enforce alignment, better use an API that enforces it instead.

PR #1123, opened 2022-11-14
Almaif parameter overflow

TopiLeppanen:
AlmaIF parameter would overflow if the device initialization file path was too long. Fixed it by dynamically allocating the strings instead.

Also fixed another issue with dependencies which affected the AlmaIF's poclbin test cases.

### DIFF
--- a/lib/CL/devices/almaif/XrtDevice.cc
+++ b/lib/CL/devices/almaif/XrtDevice.cc
@@ -30,19 +30,29 @@
 
 XrtDevice::XrtDevice(char *xrt_kernel_name) {
 
-  char xclbin_char[120];
-  snprintf(xclbin_char, sizeof(xclbin_char), "%s.xclbin", xrt_kernel_name);
+  unsigned xclbin_char_length = strlen(xrt_kernel_name) + 8;
+  char *xclbin_char = (char *)malloc(xclbin_char_length);
+  assert(xclbin_char);
+  snprintf(xclbin_char, xclbin_char_length, "%s.xclbin", xrt_kernel_name);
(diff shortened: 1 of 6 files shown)


lib/CL/devices/almaif/XrtDevice.cc
```diff
@@ -30,20 +30,30 @@
 
 XrtDevice::XrtDevice(char *xrt_kernel_name) {
 
-  char xclbin_char[120];
-  snprintf(xclbin_char, sizeof(xclbin_char), "%s.xclbin", xrt_kernel_name);
-
-  char kernel_char[120];
-  snprintf(kernel_char, sizeof(kernel_char), "%s:{%s_1}", xrt_kernel_name,
-           xrt_kernel_name);
+  unsigned xclbin_char_length = strlen(xrt_kernel_name) + 8;
```
pjaaskel, line 33: This file seems to use the C style although it's C++? Perhaps convert the style at some point.

pjaaskel (approved):

pjaaskel:
The build of AlmaIF (when I enable it) fails with:

```
g++: error: unrecognized debug output level ‘ -O3 -Wall -Wextra -Wno-long-long  -std=c++14’
```
The Cmake does some trickery? My CXXFLAGS is `-g -O3 -Wall -Wextra -Wno-long-long`so it likely converts it to -g'-O3 -Wall -Wextra -Wno-long-long' somehow?

pjaaskel:
(It's likely broken in master since I pulled the previous one in, I just had disabled ALMAIF build).

PR #1128, opened 2022-11-17
LLVM 15 for TCE/HSA + some fixes

franz:
all of the device drivers should now be converted + a few more SPIRV fixes.


pjaaskel:
I get this while trying to setup a non-TCE test env:
```
CMake Error at cmake/LLVM.cmake:35 (message):
  TCE is not disabled, and the llvm-config in WITH_LLVM_CONFIG
  (llvm-config-15) is different from llvm-config used by TCE
  (/home/visit0r/local/stow/llvmtce-14/bin/llvmtce-config)
Call Stack (most recent call first):
  CMakeLists.txt:491 (include)
```

```
cmake_switches="-DENABLE_CONFORMANCE=ON \
-DENABLE_TCE=OFF \
-DWITH_LLVM_CONFIG=llvm-config-15"
```

franz:
I haven't changed anything WRT enabling TCE. AFAIK it has been that way for a long time, that if TCE is available in PATH it is enabled by default. You can always disable by explicitly adding cmake option `-DENABLE_TCE=0` (even if it's in PATH).

franz:
ahh ok. Yeah there is some problem in CMake logic.

pjaaskel:
I try to enable the conformance suite, cmake fails:

```
-- Enabling testsuite conformance
-- enabling OpenCL-CTS version 3                                                                                                                                                                           
CMake Error at examples/CMakeLists.txt:116 (list):                                                                                                                                                         
  list sub-command FIND requires three arguments.                                                                                                                                                          
                                                                                                                                                                                                           

CMake Error at examples/CMakeLists.txt:119 (add_subdirectory):
  add_subdirectory not given a binary directory but the given source
  directory "/home/visit0r/src/pocl/examples" is not a subdirectory of
  "/home/visit0r/src/pocl/examples".  When specifying an out-of-tree source a
  binary directory must be explicitly specified.  
```

Not sure if it's a local script problem though.

franz:
probably an empty string ended up in ENABLE_TESTSUITES list, and was not handled correctly. Should be now fixed

PR #1132, opened 2022-11-23
Implement clEnqueueFillBuffer for TCE

TopiLeppanen:


pjaaskel:
Is there a test that could be enabled for TCE with this commit?

TopiLeppanen:
> Is there a test that could be enabled for TCE with this commit?

It does enable test-memfill for TCE

> This should also be included in the AlmaIF implementation of TCE

Oh right, I'll get that done as well

pjaaskel:
> > Is there a test that could be enabled for TCE with this commit?
> 
> It does enable test-fill-buffer for TCE

Can you add the respective tags so it gets run in run_*_tests?

TopiLeppanen:
> > > Is there a test that could be enabled for TCE with this commit?
> > 
> > 
> > It does enable test-fill-buffer for TCE
> 
> Can you add the respective tags so it gets run in run_*_tests?

It does now get run in run_tta_tests. This adds the label 'tce' to the fill-buffer test, and the run_tta_tests runs tests with the label 'tce'.
Do you mean something else than that?

pjaaskel:
> It does now get run in run_tta_tests. This adds the label 'tce' to the fill-buffer test, and the run_tta_tests runs tests with the label 'tce'. Do you mean something else than that?

AlmaIF, when it works.

TopiLeppanen:
Ok I did do that in #1137

PR #1134, opened 2022-11-23
Doxy: add initial poc for doxygen for poclu

RABijl:
This pr is an initial implementation that allows one to locally generate doxygen from their current build.

to enable in cmake: -DBUILD_DOXYGEN
gen: `make gen_doc`
open: `make open_doc`

Sometimes it seems like one set of functions can share the same the same description and this can be grouped with doxygen, however, IDEs will not pick up on this and defeat the point of this documentation in the first place.

I have tried to keep the following style rules:
* at the minimum say per argument if it is an input or output.
* documentation is done in the header file, unless it's a private function.
* include every file and use the brief description of the file. 
* if a function or macro assumes certain things, put it in the warning tag. 
* if the function is complicated, such as 'poclu_load_program_multidev', give a detailed description on how to use it.
* use \todo tags when a function might need to be changed
* brief tags aren't necessary if the function is very straightforward, but i've added them anyway.


jansol (approved):
Works a lot like `cargo doc [--open]`, nice!

jansol (changes requested):
This currently breaks if the build directory is not a direct subdirectory of the source root. There is no official way to make the inputs listed in the Doxyfile relative to the Doxyfile location but you can work around it with cmake:
https://stackoverflow.com/questions/65027031/how-to-set-input-relative-to-doxyfile

Also might as well add the rest of the relevant directories despite them not having much coverage in terms of doxygen comments yet:
* lib/
* include/ but maybe exclude include/hpp, include/CL and include/OpenCL as they are just the (usually unmodified) Khronos headers and contain a mainpage tag in opencl.hpp which can lead to some confusion
* $builddir/config.h
* $builddir/config2.h

RABijl:
@jansol I've added the changes you requested. this unfortunately does make the doxyfile unusable outside of cmake, so i've changed the name of the file to make it clear to people using it.

franz:
just minor nitpicks: in the root CMakeLists.txt, could you 1) rename the BUILD_DOXYGEN to ENABLE_DOXYGEN (to make it consistent with the rest of options) and 2) move it before the "summary" block

RABijl:
fixed the nitpicks to look cleaner

PR #1136, opened 2022-11-24
Almaif: fix argbuffer scalar sizes

TopiLeppanen:
Almaif used type_size to check for the scalar size which returns 0 if it cannot find the type from
pocl_llvm_metadata.c's type_data_map. This was happening for e.g. typedeffed scalar arguments. 

Now instead use the size coming from clSetKernelArg.

### DIFF
--- a/lib/CL/devices/almaif/almaif.cc
+++ b/lib/CL/devices/almaif/almaif.cc
@@ -1073,7 +1073,8 @@ void submit_kernel_packet(AlmaifData *D, _cl_command_node *cmd) {
     if (meta->arg_info[i].type == POCL_ARG_TYPE_POINTER) {
       arg_size += D->Dev->PointerSize;
     } else {
-      arg_size += meta->arg_info[i].type_size;
+      al = &(cmd->command.run.arguments[i]);
+      arg_size += al->size;
     }
   }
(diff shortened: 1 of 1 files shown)


franz:
Acceptable as a workaround, as long as you add some kind of "TODO FIXME workaround" into the code.

The size from clSetKernelArg is the user-provided size, and this should always be checked against the real size that PoCL figures out during compilation, so the real problem is in pocl_llvm_metadata.cc. 

The `type_size_map` is an old ugly kludge that needs to be removed, the code should figure out the argument type size properly using LLVM's `DataLayout` (`getTypeStoreSize` function).

TopiLeppanen:
Should that then be fixed in clSetKernelArg instead? And then it would be fine for the drivers to use this method?

clSetKernelArg should return this:
`CL_INVALID_ARG_SIZE` if arg_size does not match the size of the data type for an argument that is not a memory object

franz:
clSetKernelArg already does the check for `CL_INVALID_ARG_SIZE` - if it finds non-zero `type_size` in metadata. So *i think* it's safe to use in the driver, because at that point it should match the run argument size (there are some corner cases like vector-of-3, i don't remember how exactly we handle that).

PR #1142, opened 2022-11-25
ttasim: Fix name generation

nrother:
If adfFile would not contain a slash strrchr will return NULL, which causes undefined behavior in snprintf (usually printing "(null)"). We can just use the file as-in in this case.


jansol:
Can you check if AlmaIF has the same problem? Besides that, this looks like a fix that would be nice to have in the upcoming release, might as well retarget to PR to the release_3_1 branch.

nrother:
AlmaIF does not seem to have similar code, it always returns "ALMAIF TCE" as the device name.

I retargeted the PR to the release branch

jansol (approved):

PR #1143, opened 2022-11-25
Testsuite fixes

jansol:
Bump PyOpenCL to its latest release since the previous one does not build on Ubuntu 22.10.

Switch conformance test suite from the external fork to the latest upstream tag & update OpenCL headers from upstream so the updated CTS builds. NB: the CTS seems to fail to build when configured as a Release build (see KhronosGroup/OpenCL-CTS#1582)

### DIFF
--- a/examples/PyOpenCL/CMakeLists.txt
+++ b/examples/PyOpenCL/CMakeLists.txt
@@ -35,10 +35,10 @@ set(EXAMPLES_USE_GIT_MASTER 1)
 if(EXAMPLES_USE_GIT_MASTER)
   set(FETCH_SOURCE
     GIT_REPOSITORY "https://github.com/inducer/pyopencl.git"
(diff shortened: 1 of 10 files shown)


jansol:
Hm, interesting that those defines are missing in upstream headers. Should probably open an issue there if they are still part of the extension spec?

As for the CTS, I think the easiest solution would be to make it a git submodule? And yes those patches should definitely be upstreamed. One reason why I switched it to use upstream was because the fork does not build on my setup and upstream already has a fix for that particular issue.

franz:
@jansol  i'm not sure Nvidia ever bothered to upstream those, but i assume some people using pocl use them (my own commit just restores them).

WRT submodules, i'm not against in principle, but they were kinda broken in the past when used with git worktrees (which i use, since they allow me to work on/test multiple branches). I'll have to recheck.

PR #1150, opened 2022-12-09
Almaif: fix race condition in pocl_almaif_notify

TopiLeppanen:
The notify function would schedule commands for execution, which meant that it could get called recursively, which would then lead to its own event being freed. Which led to use-after-free race condition.

Removed the notify function scheduling of cmds, now the commands get scheduled only at either the almaif_submit or almaif_join.


TopiLeppanen:
Reworked this to a different fix, which seemed to fix @nrother 's problem as well. Now the functionality in almaif_notify is kept the same as before, where it executes the commands as before. 

The fix was to add a retain-release combo in common.c's pocl_broadcast function before the event is notified. Previously the almaif_notify function would sometimes destroy the event, which was causing the use-after-free bugs. Now it will not destroy it due to the event being retained before notify-function is called.

This PR also includes the BI kernel definition for minmaxloc, since I was using this branch for the recent demonstrator with @jansol.

PR #1151, opened 2022-12-15
fixing llvm-less builds

RABijl:
This is a work in progress pr regarding compiling kernels with poclcc and using them in builds that do not have llvm.

 -DHOST_DEVICE_BUILD_HASH is not used anymore and should possibly be removed. It was only used to create device hashes but would result in nested hashes when set.

Instead, for an llvm-less build; set :  `-DLLC_HOST_CPU` and `-DLLC_TRIPLE`  in cmake. keeping these loose instead of a -dhost_device_build_hash should allow for easier future cross compilation options.

On a related note; the pocl documentation should possibly be updated. `llc -mcpu=help`  could be changed to `llc --version` since on ubuntu llvm version 14.0.0 `llc -mcpu=help` errors and `llc -march=help` hangs but `llc --version` will show the default triple and host cpu. Fortunately, commands like `llc -march=arm -mcpu=help` does still work.

### working example config

build poclcc with cmake flags:
`-DENABLE_LOADABLE_DRIVERS=1`

llvm-less build:
`-DENABLE_LOADABLE_DRIVERS=1 -DENABLE_LLVM=0 -DHOST_DEVICE_BUILD_HASH=pthread-x86_64-pc-linux-gnu-broadwell -DLLC_HOST_CPU=broadwell -DLLC_TRIPLE=x86_64-pc-linux-gnu` 
ofc  the llc options are device specific and will need to be changed.

running example0 with the following env variables:
```
OCL_ICD_VENDORS=/home/rabijl/Projects/pocl/cmake-build-debug-no-llvm/ocl-vendors \
POCL_BUILDING=1 \
POCL_DEBUG=all \
POCL_DEVICES=pthread \
POCL_WORK_GROUP_SPECIALIZATION=0
```
and passing `b` as argument

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -494,9 +494,11 @@ if(ENABLE_LLVM)
   endif()
 
   if(ENABLE_HOST_CPU_DEVICES)
-  if(NOT DEFINED HOST_DEVICE_BUILD_HASH)
-      set(HOST_DEVICE_BUILD_HASH "${LLC_TRIPLE}")
-  endif()
+
+#    TODO: remove this since it's not being used anywhere
+#  if(NOT DEFINED HOST_DEVICE_BUILD_HASH)
+#      set(HOST_DEVICE_BUILD_HASH "${LLC_TRIPLE}")
+#  endif()
 
   if(INTEL_SDE_AVX512)
     set(HOST_CPU_FORCED 1 CACHE INTERNAL "CPU is forced by user" FORCE)
--- a/lib/CL/devices/almaif/almaif.cc
+++ b/lib/CL/devices/almaif/almaif.cc
@@ -505,7 +505,7 @@ unsigned int pocl_almaif_probe(struct pocl_device_ops *ops) {
 
 char *pocl_almaif_build_hash(cl_device_id /*device*/) {
   char *res = (char *)calloc(1000, sizeof(char));
-  snprintf(res, 1000, "almaif-%s", HOST_DEVICE_BUILD_HASH);
+  snprintf(res, 1000, "almaif-%s", OCL_KERNEL_TARGET);
   return res;
 }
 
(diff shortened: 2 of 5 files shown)


RABijl:
I've done a bit of experimenting with setting the LLC_HOST and LLC_TRIPLE to an arm32 cpu and the good news is that the pthread driver will compile a kernel to this. The bad news is that due to pocl's reliance on dlopen, pocl can't get get the binary size or contents needed to create a poclbin file since the compiled so is in a 32b arm format.

RABijl:
@franz and @pjaaskel  any opinions on what can be done to make this pr fit to be merged and removing the `-DHOST_DEVICE_BUILD_HASH` build option?

pjaaskel:
Can you retrieve the information somehow else than dlopen()? What does it need it for? Using ELF parser?

RABijl:
ELF parser might work, the main reason it seems dlopen is used is to get workgroup sizes. Maybe that could even be skipped if the kernel is not specialized.

pjaaskel:
The PR makes sense to me. I'm not sure what extra benefits the HOST\_DEVICE\_BUILD\_HASH would provide. For AlmaIF we might need to include the targeted subdevice in the build hash. For OpenASIP of course the ADF file is essential in the hash etc.

TopiLeppanen:
> For AlmaIF we might need to include the targeted subdevice in the build hash. For OpenASIP of course the ADF file is essential in the hash etc.

AlmaIF overrides the build_hash callback already in case OpenASIP is enabled. It doesn't know about the adf-files at pocl-compilation time anyway, so this change doesn't affect that. This hash is the one that gets set when kernel compilation is disabled (emulation device or llvm-less build on fpga with just BI kernels). Therefore in those cases it's not even needed for anything.

RABijl:
closing this pr until I have the opportunity to work on this

PR #1158, opened 2023-01-04
Almaif: qol fixes

RABijl:
* poclcc support for cl_device_type_custom
* couple new lines in error messages of almaif
* poclu_get_multiple_devices has toggle for custom devices
* poclu_get_any_device_2 will look for custom if no normal found
* device kernel hashes are also calculated for builtin kernels (pthread bi)

### DIFF
--- a/lib/CL/pocl_build.c
+++ b/lib/CL/pocl_build.c
@@ -912,8 +912,7 @@ compile_and_link_program(int compile_program,
           goto ERROR;
         }
 
-      if (program->builtin_kernel_names == NULL)
-        setup_device_kernel_hashes (program);
+      setup_device_kernel_hashes (program);
     }
 
   for (device_i = 0; device_i < program->num_devices; device_i++)
--- a/poclu/misc.c
+++ b/poclu/misc.c
@@ -83,14 +83,21 @@ poclu_get_any_device2 (cl_context *context, cl_device_id *device,
 
   err = clGetDeviceIDs (*platform, CL_DEVICE_TYPE_ALL, 1, device, NULL);
   if (err != CL_SUCCESS)
-    return err;
+    {
+
+      err = clGetDeviceIDs (*platform, CL_DEVICE_TYPE_CUSTOM, 1, device, NULL);
+      if (err != CL_SUCCESS)
+        {
+          return err;
+        }
+    }
 
   *context = clCreateContext (NULL, 1, device, NULL, NULL, &err);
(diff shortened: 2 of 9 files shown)


poclu/misc.c
```diff
@@ -113,19 +122,57 @@ poclu_get_multiple_devices (cl_platform_id *platform, cl_context *context,
...
-      = (cl_command_queue *)calloc (*num_devices, sizeof (cl_command_queue));
+      = (cl_command_queue *) calloc (*num_devices, sizeof (cl_command_queue));
 
-  err = clGetDeviceIDs (*platform, CL_DEVICE_TYPE_ALL, *num_devices, devs,
-                        NULL);
-  if (err != CL_SUCCESS)
-    return err;
+  if (num_dev_all != 0)
+    {
+      err = clGetDeviceIDs (*platform, CL_DEVICE_TYPE_ALL, *num_devices, devs,
```
TopiLeppanen, line 162: Should it instead be the _num_dev_all_ in this query? A bit nitpicky, since using either one here might work, since the spec defines:
_The number of OpenCL devices returned is the mininum of the value specified by num_entries or the number of OpenCL devices whose type matches device_type_

TopiLeppanen (commented):
Looks good to me. This is exactly the annoying workaround we have to do for CL_DEVICE_TYPE_CUSTOM since it is not included CL_DEVICE_TYPE_ALL. Related to this: https://github.com/KhronosGroup/OpenCL-Docs/issues/300

lib/CL/pocl_build.c
```diff
@@ -912,8 +912,7 @@ compile_and_link_program(int compile_program,
           goto ERROR;
         }
 
-      if (program->builtin_kernel_names == NULL)
-        setup_device_kernel_hashes (program);
+      setup_device_kernel_hashes (program);
```
TopiLeppanen, line 915: Does this also work with programs that don't actually compile their BI kernels at all? E.g. AlmaIF emulation device BI kernels? Is e.g. the build_hash[] still initialized to something reasonable for those?

TopiLeppanen (commented):
I just realized this one as well, since I noticed you made the changed for pthread's BI kernels, which are compiled behind-the-scens. Did you also test this for the non-compilation case?

poclu/misc.c
RABijl: you're right, fixed

TopiLeppanen (approved):

PR #1160, opened 2023-01-13
Basic cl_khr_subgroups implementation for CPU

pjaaskel:
The subgroup is always the X dimension, and there is only one subgroup in flight at the same time. Thus there are Z*Y SGs per WG. It passes the CTS test even though independent progress is not supported because there's only one SG in flight at the same time making progress.

Also preliminary shuffle and ballot implementations which work only with uniform execution for now.

The cross-WI data exchange is implemented via __pocl_{wg,local_mem}_alloca(), an internal function that allocates "local memory" (thread stack) dynamically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 #=============================================================================
 #   CMake build system files
 #
-#   Copyright (c) 2014-2018 pocl developers
+#   Copyright (c) 2014-2023 pocl developers
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a copy
 #   of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -777,6 +777,9 @@ option(ENABLE_CONFORMANCE "Enable conformance to OpenCL standard. \
   Enabling this option this does not guarantee conformance (depends on hardware), \
   but CMake will give errors if options that conflict with conformance are used" OFF)
 
+option(ENABLE_UNFINISHED_EXTENSIONS "Advertise extensions in device queries which \
+might work partially, but have known unfinished features." OFF)
+
```
franz, line 782: This should have a conflict with ENABLE_CONFORMANCE=ON (that is the purpose of ENABLE_CONFORMANCE flag), like e.g. lines 783-785.

In fact i'm not sure this new CMake flag is required; it's already the behavior (at least for CPU drivers) that we just advertise unfinished features when ENABLE_CONFORMANCE=OFF and disable them when it's ON.

franz (changes requested):
looks good, few minor nitpicks

PR #1162, opened 2023-01-17
Almaif standalone

TopiLeppanen:
AlmaIF: Add standalone mode for openasip backend

Works similarly to the tce-driver's standalone mode.
Produces a C program which can then be compiled with oacc
and simulated with ttasim and ghdl.

Depends on recent OpenASIP PR for AlmaIF testbench generation:
https://github.com/cpc/openasip/pull/225

This PR also moves Openasip-specific AlmaIF code to almaif/openasip.

The only non-almaif change is the commit adding support for absolute paths in add_test_pocl.cmake


lib/CL/devices/almaif/almaif.cc
```diff
@@ -33,16 +33,16 @@
...
 
-#include "AlmaifShared.hh"
-#include "AlmaifCompileTCE.hh"
 #include "AlmaifCompile.hh"
+#include "AlmaifShared.hh"
 #include "bufalloc.h"
 #include "common.h"
 #include "common_driver.h"
 #include "devices.h"
+#include "openasip/AlmaifCompileTCE.hh"
```
pjaaskel, line 45: These "system/library includes" should use <> instead of "".

lib/CL/devices/almaif/openasip/AlmaifCompileTCE.hh
```diff
@@ -32,6 +32,10 @@ int pocl_almaif_tce_initialize(cl_device_id device, const char *parameters);
 int pocl_almaif_tce_cleanup(cl_device_id device);
 void pocl_almaif_tce_compile(_cl_command_node *cmd, cl_kernel kernel,
                              cl_device_id device, int specialize);
+void pocl_almaif_tce_produce_standalone_program(AlmaifData* D,
```
pjaaskel, line 35: did you run clang-format on these files?

pjaaskel (commented):
Looks pretty good. Can you try if the clang-formatter script can improve the style (it might not work well with diffs)? Also this feature is so useful it's usage shall be documented in the user manual and its existence mentioned in CHANGES.

lib/CL/devices/almaif/almaif.cc
TopiLeppanen: It's not including something from system OpenASIP, but instead from PoCL's almaif/openasip/ directory, which is for files related to almaif-drivers "openasip-backend"

lib/CL/devices/almaif/openasip/AlmaifCompileTCE.hh
TopiLeppanen: I thought I did, but apparenly format-last-commit.sh and format-branch.sh have a slightly different file regex, so these didn't get picked up by format-last-commit.sh.
format-branch.sh has:
'(.*(\.hh$|\.cc$))|(lib/llvmopencl/.*\.h)|(lib/CL/devices/tce/.*)'
And format-last-commit.sh:
'(.*(\.hpp$|\.cc$|\.cpp$))|(lib/llvmopencl/.*\.h)'
Which one should it be? And actually what would be the preferred naming convention for (these) C++ sources and headers?

TopiLeppanen:
Force-pushed the CHANGES update and the formatting fixes

lib/CL/devices/almaif/openasip/AlmaifCompileTCE.hh
TopiLeppanen: Never mind, you seem to have fixed these in the master in the meantime. Well, now after the rebasing to master, the format-branch.sh doesn't do anything anymore, so this should now be formatted correctly.

franz: I've added the .cpp extension to `format-last-commit.sh` few weeks ago, but i forgot to add it to the `format-branch.sh`. We have some .cpp files so it's needed; actually i think we also need both .hh and .hpp; not sure why is `lib/CL/devices/tce` there.

Issue #1167, opened 2023-01-24
Failing to handle correctly guarded barrier inside a loop with work-item dependent iterations


fodinabor:
According to recent OpenCL specifications, the following kernel should produce the same output independent of the presence of the second barrier.
> [OpenCL 3.0, 6.15.8. Synchronization Functions](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#synchronization-functions):
For these functions, if any work-item in a work-group encounters a barrier, the barrier must be encountered by all work-items in the work-group before any are allowed to continue execution beyond the barrier. If the barrier is inside a conditional statement, then all work-items in the work-group must enter the conditional if any work-item in the work-group enters the conditional statement and executes the barrier. If the barrier is inside a loop, then all work-items in the workgroup must execute the barrier on each iteration of the loop if any work-item executes the barrier on that iteration.

```
__kernel
void test_kernel(global int *output) {
  __local int scratch[256]; // max wg size: 256

  const int gid_x = get_global_id(0);
  const int gid_y = get_global_id(1);
  const int gid_z = get_global_id(2);

  const int global_id =
      gid_x + get_global_size(0) * (gid_y + get_global_size(1) * gid_z);
  const int local_id =
      get_local_id(0) +
      get_local_size(0) *
          (get_local_id(1) + get_local_size(1) * get_local_id(2));

  scratch[local_id] = global_id;

  barrier(CLK_LOCAL_MEM_FENCE);

  for (size_t i = 0; i < 2 + local_id; i++) {
    scratch[local_id] += i;
    // only call the barrier if all work-items still run the loop.
    if (i < 2)
      barrier(CLK_LOCAL_MEM_FENCE);
  }

  output[global_id] = scratch[local_id];
}
```

Running this as a test case with `2 4 1 1`, the output should be the following:
```
0: 1
1: 4
2: 8
3: 13
4: 5
5: 8
6: 12
7: 17
```

When running with the default optimizations, the kernel does not terminate (optimized to a loop with no exit..), when running with `POCL_EXTRA_BUILD_FLAGS="-cl-opt-disable"`, the kernel terminates with the following results:
```
0: 10
1: 11
2: 12
3: 13
4: 14
5: 15
6: 16
7: 17
```
The "unoptimized" version shows that POCL relies on a much more strict interpretation of the OpenCL 1.0-2.x barrier restrictions:
[...]

PR #1169, opened 2023-01-25
Import Open SYCL's CBS implementation as a fallback.

fodinabor:
This can be used as a more general fallback for cases the POCL kernel compiler would fail. There's lots of room for improvement here... so far this is primarily a copy-paste of the main pass integrated into POCL (using its analyses as well):
- make use of compile-time known wg-size,
- maybe merge the work-item loop creation code (at least for when there's no barrier at all ;-))
- actually decide which path to take depending on the input kernel
- hipSYCL had a separate pass for adding the parallel access md, mark wi-loops somewhere
- this CBS impl does not partially unroll or anything
- while this version uses as much of the existing infrastructure in POCL, e.g. POCL's VariableUniformityAnalysis is used, hipSYCL's version uses a uniformity analysis that also provides knowledge about contiguous values. By tracing back contiguous values to the induction variables and uniform values instead of storing them, the number of stack arrays is reduced and more continuity information is exposed to the vectorizer. (hipSYCL experiments showed a 17% geomean speedup on select benchmarks with this). The contiguous value stuff is currently just commented out, so still some cleanup to do here.

This is a draft to give an idea of the overhead that using CBS as a fallback might incur, following up on #1167.
The performance difference to hipSYCL's fiber implementation is typically enormous, although it obviously scales with the barrier/compute ratio.
More info in the hipSYCL PR https://github.com/illuhad/hipSYCL/pull/682 and my [thesis](https://joameyer.de/hipsycl/Thesis_JoachimMeyer.pdf).

Note, this passes the full internal test suite except `different_implicit_barrier_injection_scenarios`, but that's expected since the implicit barrier passes are not in use here.


pjaaskel:
> This can be used as a more general fallback for cases the POCL kernel compiler would fail. There's lots of room for improvement here... so far this is primarily a copy-paste of the main pass integrated into POCL (using its analyses as well):

Ah! Greatness! You already made my day even though it just started :) I'll take a look at the patch as soon as I can.

>     * maybe merge the work-item loop creation code (at least for when there's no barrier at all ;-))

Or we can just fallback to the "default" WILoops in this trivial case?

>     * actually decide which path to take depending on the input kernel

This is an important piece. It needs some non-trivial CFG analysis to detect cases like you presented. But we can merge the baseline implementation before this is done as it's useful as explicitly controlled alternative as well for those tricky cases.

>     * this CBS impl does not partially unroll or anything

Overall the philosophy in WG function generation has been to rely on the basic LLVM optimizations after it as much as possible. For example, we have gone back from aggressively inlining everything to letting LLVM do more of the decisions per-target.

>     * while this version uses as much of the existing infrastructure in POCL, e.g. POCL's VariableUniformityAnalysis is used, hipSYCL's version uses a uniformity analysis that also provides knowledge about contiguous values. By tracing back contiguous values to the induction variables and uniform values instead of storing them, the number of stack arrays is reduced and more continuity information is exposed to the vectorizer. (hipSYCL experiments showed a 17% geomean speedup on select benchmarks with this). The contiguous value stuff is currently just commented out, so still some cleanup to do here.

VUA is quite primitive and would be great if it was improved. I've seen the LLVM uniformity analysis that landed years ago, which could be utilized also to (partially) replace the VUA of PoCL, but this is yet another "nice things to do" for which no one (including me) has found the time to do yet.

> This is a draft to give an idea of the overhead that using CBS as a fallback might incur, following up on #1167. The performance difference to hipSYCL's fiber implementation is typically enormous, although it obviously scales with the barrier/compute ratio. More info in the hipSYCL PR [illuhad/hipSYCL#682](https://github.com/illuhad/hipSYCL/pull/682) and my [thesis](https://joameyer.de/hipsycl/Thesis_JoachimMeyer.pdf).

OK, interesting. Will take a look at your thesis results.

> Note, this passes the full internal test suite except `different_implicit_barrier_injection_scenarios`, but that's expected since the implicit barrier passes are not in use here.

Nice! We should force the WG function method in that test to WILoops and perhaps add separate tests for CBS.

BTW I see hipSYCL can produce SPIR-Vs. Can it also use OpenCL as a runtime now?

fodinabor:
> Or we can just fallback to the "default" WILoops in this trivial case?

Counts ;-)

> VUA is quite primitive and would be great if it was improved. I've seen the LLVM uniformity analysis that landed years ago, which could be utilized also to (partially) replace the VUA of PoCL, but this is yet another "nice things to do" for which no one (including me) has found the time to do yet.

Jup, we use LLVM's SyncDependenceAnalysis + the value shape analysis built on top of that in the Region Vectorizer (https://github.com/cdl-saarland/rv/).

> BTW I see hipSYCL can produce SPIR-Vs. Can it also use OpenCL as a runtime now?

hipSYCL's SPIR-V support is currently only used in conjunction with LevelZero for Intel GPUs, so nope..

cmake/add_test_pocl.cmake
```diff
@@ -72,7 +72,8 @@ function(add_test_pocl)
   if(NOT ENABLE_ANYSAN)
     set_tests_properties("${POCL_TEST_NAME}" PROPERTIES
                          PASS_REGULAR_EXPRESSION "OK"
-                         FAIL_REGULAR_EXPRESSION "FAIL")
+                         FAIL_REGULAR_EXPRESSION "FAIL"
+                         ENVIRONMENT POCL_WORK_GROUP_METHOD=cbs)
```
pjaaskel, line 76: Does this now only test with cbs? Ideally we'd test both with cbs and wiloops (repl we can start to phase out).

lib/llvmopencl/CMakeLists.txt
```diff
@@ -72,6 +73,8 @@ set(LLVMPASSES_SOURCES "AllocasToEntry.cc"
                        "RemoveBarrierCalls.h"
                        "RemoveOptnoneFromWIFunc.cc"
                        "RemoveOptnoneFromWIFunc.h"
+                       "SubCfgFormation.cc"
```
pjaaskel, line 76: Can you rename to SubCFGFormation?

lib/llvmopencl/CanonicalizeBarriers.cc
```diff
@@ -147,8 +152,10 @@ CanonicalizeBarriers::ProcessFunction(Function &F) {
     // they just start several parallel regions. Simplifies
     // loop handling.
 
-    const bool HAS_NON_BRANCH_INSTRUCTIONS_AFTER_BARRIER =
```
pjaaskel, line 150: Please check style and naming convention. We want to clean up the code base with new commits towards using LLVM style for C++ files and GNU C for C files. The patch formatter script might help.

pjaaskel: We clean up nearby/touched parts.

lib/llvmopencl/SubCfgFormation.cc
```diff
@@ -0,0 +1,1547 @@
...
+                       << IncomingBB->getName() << " from PHI " << *Phi << "\n";
+        }
+      }
+    }
+  }
+}
+
+// // check if a contiguous value can be tracked back to only uniform values and
+// // the wi-loop indvar currently cannot track back the value through PHI
+// nodes. bool dontArrayifyContiguousValues(
```
pjaaskel, line 673: Can you comment out with #if 0 if you plan to use this code in the future and the reason why it's included in the repo?

lib/llvmopencl/SubCfgFormation.cc
```diff
@@ -0,0 +1,1547 @@
...
+
+  formSubCfgs(F, LI, DT, PDT, VUA);
+
+  return false;
+}
+
+char SubCfgFormationPassLegacy::ID = 0;
+
+// llvm::PreservedAnalyses
+// SubCfgFormationPass::run(llvm::Function &F, llvm::FunctionAnalysisManager
```
pjaaskel, line 1526: Same here (#if 0 ...).

lib/llvmopencl/SubCfgFormation.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * This file is part of hipSYCL, a SYCL implementation based on CUDA/HIP
+ *
```
pjaaskel, line 3: Please point to the origin (in github) so we can then more easily synch fixes from upstream.

tests/workgroup/CMakeLists.txt
```diff
@@ -62,6 +62,10 @@ add_test_pocl(NAME "workgroup/cond_barriers_in_for"
               EXPECTED_OUTPUT "cond_barriers_in_for_2_4_1_1.stdout"
               COMMAND "run_kernel" "cond_barriers_in_for.cl" 2 4 1 1)
 
+add_test_pocl(NAME "workgroup/cond_barrier_in_var_for"
```
pjaaskel, line 65: This should be ran only with the CBS, but the rest (that work with both) both with wiloops and cbs, right?

pjaaskel (approved):
Looks good. Thanks for your work. I did a ~30 min review as I trust the SubCFG is reviewed and battle-tested in hipSYCL so I mostly just perused it. Please also add documentation to the user manual, entry to CHANGES and make the tests such that they test both CBS and wiloops (whenever they are supposed to work) and we can pull this in!

fodinabor:
Working to integrate the feedback.

In the mean-time.. what's the recommended iteration scheme with OpenCL?
Aka, which is the dimension that should be used to access neighboring elements? x or z? Aka, what should be the innermost loop? In SYCL the convention is to use the highest dimension index, but I suspect OpenCL to use x as innermost loop?

pjaaskel:
At least to me it's always been "x". Not sure if it's "OpenCL convention", or just PoCL convention.

fodinabor:
Alright. Changed the nesting order :)
Not sure if I got all the places, where documenting this would be goodthe documentation?
Otherwise, I think it should be good to go...?

CHANGES
```diff
@@ -11,6 +11,11 @@ Notable User Facing Changes
   subgroup that always executes the whole X-dimension's WIs.
 - Added initial (incomplete) support for 
   cl_intel_required_subgroup_size for CPUs
+- Pulled in alternative work-group function mode for non-SPMD from hipSYCL:
+  Continuation-based synchronization is slightly more general than POCL's
+  current kernel compiler, but allows for fewer hand-rolled optimizations.
+  CBS is expected to work for kernels that POCL's current kernel compiler
+  does not support. For most other kernels it is expected to perform slightly worse.
```
pjaaskel, line 18: Can you add a sentence how it can be enabled?

doc/sphinx/source/using.rst
```diff
@@ -276,6 +276,12 @@ pocl.
               might avoid storing work-item context to memory.
               However, the code bloat is increased with larger
               WG sizes.
+    
+    cbs    -- Use continuation-based synchronization to execute work-items
+              on non-SPMD devices.
+              CBS is expected to work for kernels that 'loops' does not support.
+              For most other kernels it is expected to perform slightly worse.
+              Also enables the LLVM LoopVectorizer.
```
pjaaskel, line 284: Here you could advertise your master's/paper by adding a link to it for people wanting to understand how it works. Which reminds me... if you have any PoCL-comparing/related publications, you can add them to the publications.mak to advertise. I can then make it visible here: http://portablecl.org/publications.html

pjaaskel (approved):
Looks good. I'll test&push if it passes my local envs. Those two remarks you can fix later if you wish.

pjaaskel:
Fails to build for TCE, likely because of the test case name change ('_loop' appending):
```
CMake Error at tests/tce/printf/CMakeLists.txt:50 (set_tests_properties):
  set_tests_properties Can not find test to add properties to: tcele/printf
...
```
Etc.

pjaaskel (changes requested):
Fails to build for TCE, likely because of the test case name change ('_loop' appending):

```
CMake Error at tests/tce/printf/CMakeLists.txt:50 (set_tests_properties):
  set_tests_properties Can not find test to add properties to: tcele/printf
...
```
Etc.

fodinabor:
Before I try to fix that: my understanding is that TCE is usually a non-SPMD target, i.e. one would like to test Loops & CBS there?
i.e. what is the fix to apply: add `WORKITEM_HANDLER "loops"` to the `add_pocl_test` commands or set the properties on both variants?

pjaaskel:
> Before I try to fix that: my understanding is that TCE is usually a non-SPMD target, i.e. one would like to test Loops & CBS there? i.e. what is the fix to apply: add `WORKITEM_HANDLER "loops"` to the `add_pocl_test` commands or set the properties on both variants?

Correct, _for now_ OpenASIP are "MIMD" (SPMD support is WiP). But since we haven't tested with CBS there yet, you can just add loops for now and we'll expand the coverage when we test with CBS. Thanks!

doc/sphinx/source/using.rst
fodinabor: I'll decide which to add there after the IWOCL deadline ;-)

pjaaskel:
TCE passed, but this popped up next. Strangely, with the wiloops method?
```
  5/540 Test   #95: regression/test_issue_893_loops .....................................................***Failed  Error regular expression found in output. Regex=[FAIL]  2.58 sec
CMake Error at /home/visit0r/src/pocl/cmake/run_test.cmake:34 (message):
  FAIL: Test exited with nonzero code (Child aborted):
  /home/visit0r/src/pocl/test-pr-cmake-llvm-15-build/tests/regression/test_issue_893


  STDOUT:

  land.lhs.true31(EN) for.inc51.thread118 for.inc51.thread
  for.body.backedge.latchbarrier.prebarrier(EX)

  ### dumped CFG to loopy_kernel.dot.0

  

  STDERR:

  suspicious block: for.inc51.thread

  the entry is: land.lhs.true31

  test_issue_893:
  /home/visit0r/src/pocl/lib/llvmopencl/ParallelRegion.cc:399: bool
  pocl::ParallelRegion::Verify(): Assertion `0 && "Incoming edges to
  non-entry block!"' failed.
```

fodinabor:
Right. I had tested that on master once, seeing that there as well, so I thought it was prevalent there as well, but probably didn't clear the kernel cache.

Afaict, the issue is that the tests are run with `POCL_WORK_GROUP_METHOD=loops` instead of `loopvec` which has always been the case (default) so far. I assume that some of the additional set when using `loopvec` might make the difference here...?
Not sure if I can fix `loops` that just now.
Should I change the default / just run that test with `loopvec` instead of `loops`?

pjaaskel:
Yep, if that's the easy way out and test the previous status quo. Honestly I cannot recall what's the difference (any more) with these methods.

pjaaskel:
Grepping reveals that it turns some important knobs for "loopvec" only in pocl_llvm_utils.cc which should be also enabled for "loops" in general:

```C++      
      // Disable jump threading optimization with following two options from                                                                
      // duplicating blocks. Using jump threading will mess up parallel region                                                              
      // construction especially when kernel contains barriers.                                                                             
      // TODO: If enabled then parallel region construction code needs                                                                      
      // improvements and make sure it doesn't disallow other optimizations like                                                            
      // vectorization.                                                                                                                     
      O = opts["jump-threading-threshold"];
      assert(O && "could not find LLVM option 'jump-threading-threshold'");
      O->addOccurrence(1, StringRef("jump-threading-threshold"), StringRef("0"),
                       false);
      O = opts["jump-threading-implication-search-threshold"];
      assert(O && "could not find LLVM option "
                  "'jump-threading-implication-search-threshold'");
      O->addOccurrence(1,
                       StringRef("jump-threading-implication-search-threshold"),
                       StringRef("0"), false);
```
Enabling those for "loops" too might fix this.

fodinabor:
jup, fixes that as well :)
Guess loopvec as default for the tests still is sensible, as that's the default behavior of POCL.

pjaaskel:
There's a segfault with LLVM 14 with `workgroup/loop_with_two_paths_to_the_latch_cbs`. Seems to crash at kernel runtime.

pjaaskel:
Should this now work with LLVM 14 as well? Can you rebase please?

fodinabor:
Yup, tested with LLVM 14 and 15 now and rebased.

pjaaskel:
Yep, LLVM 15, 14 and 13 work, thanks! Unfortunately I see two issues with LLVM 12 (yeah, I know a CI visible in Github would be nice):
```
217/260 Test: workgroup/cond_barrier_in_var_for
...
...
Files "/home/visit0r/src/pocl/tests/workgroup/cond_barrier_in_var_for_2_4_1_1.stdout" to "/tmp/cmake_testrun_vl4H1" are different.
--- /home/visit0r/src/pocl/tests/workgroup/cond_barrier_in_var_for_2_4_1_1.stdout       2023-02-10 13:17:02.564993130 +0200
+++ /tmp/cmake_testrun_vl4H1    2023-02-10 16:03:22.608924785 +0200
@@ -1,9 +1,9 @@
 0: 1
-1: 4
-2: 8
-3: 13
+1: 2 
+2: 3
+3: 4
 4: 5
-5: 8
-6: 12
-7: 17
+5: 6
+6: 7
+7: 8
 OK
...
221/260 Testing: workgroup/conditional_barrier_cbs
...
--- /home/visit0r/src/pocl/tests/workgroup/cond_barriers_1_2_1_1_cbs.stdout     2023-02-10 13:17:02.564993130 +0200
+++ /tmp/cmake_testrun_TUc9H    2023-02-10 16:03:23.144928090 +0200
@@ -1,6 +1,6 @@
 LOCAL_ID=0 before if
-LOCAL_ID=0 inside if
 LOCAL_ID=1 before if
+LOCAL_ID=0 inside if
 LOCAL_ID=1 inside if
 LOCAL_ID=0 after if
 LOCAL_ID=1 after if

```

fodinabor:
Will take a look. Just making sure... LLVM 12.0.0 or LLVM 12.0.1?

pjaaskel:
Seems to be 12.0.0. LLVM 11 passed, but there are build errors against LLVM 10 (log attached). LLVM 9 we won't support anymore in the next release, so not testing that.
[llvm-10-errs.log](https://github.com/pocl/pocl/files/10708967/llvm-10-errs.log)

pjaaskel:
I also had to fix a couple of accel tests to fix its build, can you apply this also to your PR?

```patch
commit 39ea8c89fbd69546705d59ba6aebcfb82ae719c9 (HEAD -> master)
Author: Pekka Jääskeläinen <pekka.jaaskelainen@tuni.fi>
Date:   Fri Feb 10 14:24:55 2023 +0200

    fix accel test initializations

diff --git a/examples/accel/CMakeLists.txt b/examples/accel/CMakeLists.txt
index 50e33577..e138a0b0 100644
--- a/examples/accel/CMakeLists.txt
+++ b/examples/accel/CMakeLists.txt
@@ -185,12 +185,14 @@ endif ()

     # The actual standalone tests
     add_test_pocl(NAME "examples/example0/ttasim_standalone_${core_name}"
-        COMMAND "run_standalone_ttasim.sh####${core_name}"
-        EXPECTED_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/golden_output.txt")
+      COMMAND "run_standalone_ttasim.sh####${core_name}"
+      WORKITEM_HANDLER "loopvec"
+      EXPECTED_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/golden_output.txt")

     add_test_pocl(NAME "examples/example0/rtlsim_standalone_${core_name}"
-        COMMAND "run_standalone_rtlsim.sh####${core_name}"
-        EXPECTED_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/golden_output.txt")
+      COMMAND "run_standalone_rtlsim.sh####${core_name}"
+      WORKITEM_HANDLER "loopvec"
+      EXPECTED_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/golden_output.txt")

     set_tests_properties ("examples/example0/ttasim_standalone_${core_name}" "examples/example0/rtlsim_standalone_${core_name}"
         PROPERTIES

```

fodinabor:
Got LLVM 10 fixed (will push after verifying the other LLVM versions are still working ;-))
Not sure what's the issue with LLVM 12 though... cannot reproduce either of the test failures locally. Maybe that's something target-specific? What CPU is POCL building for there?

pjaaskel:
It actually isn't due to LLVM 12, but work-group specialization. If you disable WG specialization (like I happen to do with my LLVM 12 test script, just to poke this parameter) with `POCL_WORK_GROUP_SPECIALIZATION=0` the tests fail also for LLVM 15. Is there something _relying_ on WG specialization? We want to support also non-specialized WGs (dynamic WI loops) for general local size support for binary-only execution envs.

fodinabor:
Just as a comment: jup, `POCL_WORK_GROUP_SPECIALIZATION` did still use the loops variant in all cases, so I changed that.
Had to fix some issue down the line (we defaulted to 1024 as a size of the context arrays instead of using the dynamic group size), which again broke the test with the 15000 element private array :)
Works now, looking forward to another round of testing :)

pjaaskel:
LLVM 11 fails one test (wrong results):
```
ctest -R cond_barriers_in_for_cbs -V
...
216: -- Comparing output..
216: --- /home/visit0r/src/pocl/tests/workgroup/cond_barriers_in_for_2_4_1_1.stdout     2022-01-06 14:10:58.394596087 +0200
216: +++ /tmp/cmake_testrun_sMhps       2023-02-14 19:04:25.123503308 +0200
216: @@ -1,9 +1,9 @@
216: -0: 4
216: -1: 5
216: -2: 6
216: -3: 7
216: -4: 16
216: -5: 17
216: -6: 18
216: -7: 19
216: +0: 2
216: +1: 3
216: +2: 4
216: +3: 5
216: +4: 8
216: +5: 9
216: +6: 10
216: +7: 11
216:  OK
```

pjaaskel:
LLVM 10 is ok.

fodinabor:
Alright. Only LLVM 10 uses the (not-so-well-tested) case where CBS needs the PHIs to be demoted.
That's in good part since Open SYCL (hipSYCL) did no longer support LLVM 10 when upstreamed.

pjaaskel:
Thanks for the contribution! This should be more useful in the future when we add automated fallback when encountering too tricky CFGs to CBS.

PR #1180, opened 2023-02-25
Document the software stack below and above PoCL

pjaaskel:
Starting point for a graph-based documentation for navigating the APIs below and above PoCL. This collected info has been called for in several occasions. I'll add this to the docs after finished. 

TODO:

- [ ] Define the specific extensions needed and supported.
- [ ] Denote the maturity/testing level of the paths.

All feedback welcome.

Render with `dot -Tpdf sw-stack-graph.dot > sw-stack-graph.pdf`


pjaaskel:
sw-stack-graph

PR #1192, opened 2023-03-28
update install.rst to note specifying random hash in case of almaif devices

bojle:
`accel.rst` [had](https://github.com/pocl/pocl/commit/74b46f46be9c8205e1931983164c7f6770f96c2a) such a note. I don't know why it was not copied to current document

### DIFF
--- a/doc/sphinx/source/install.rst
+++ b/doc/sphinx/source/install.rst
@@ -283,8 +283,8 @@ The string after "HSTR:" is the device build hash.
   This is required because pocl binaries contain a device hash, and the LLVM-less
   pocl needs to know which binaries it can load.
 
-**NOTE**: If you've enabled the `almaif <http://portablecl.org/docs/html/almaif.html#usage>`_ 
-device, `HOST_DEVICE_BUILD_HASH` can be set to anything you want. Reason being, fixed function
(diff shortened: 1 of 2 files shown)


doc/sphinx/source/install.rst
```diff
@@ -283,6 +283,9 @@ The string after "HSTR:" is the device build hash.
   This is required because pocl binaries contain a device hash, and the LLVM-less
   pocl needs to know which binaries it can load.
 
+**NOTE**: If you've enabled the `almaif <http://portablecl.org/docs/html/almaif.html#usage>`_
```
pjaaskel, line 286: Can you fix the URL to a sphinx reference, not an absolute link?

pjaaskel (changes requested):

doc/sphinx/source/install.rst
bojle: sure

bojle: Do you think adding an example would be nice? like [the original](https://github.com/pocl/pocl/commit/74b46f46be9c8205e1931983164c7f6770f96c2a)?

pjaaskel: I think it's clear enough now, thanks.

bojle: great!

PR #1198, opened 2023-04-03
builtin_kernels.hh: allow POCL_CDBI_LAST to be automatically assigned

bojle:
POCL_CDBI_LAST is used as the size parameter for pocl_BIKernels. Since, integers in enums are automatically assinged a value of previous-integer + 1, POCL_CDBI_LAST's value should be assigned in such a way as opposed to being hard-coded.

Furthermore, comment on the top vaguely directs to add new kernel definitions to the end of the enum. Adding anything after POCL_CDBI_LAST will go undetected as it is used for size. This is corrected.

### DIFF
--- a/lib/CL/devices/builtin_kernels.hh
+++ b/lib/CL/devices/builtin_kernels.hh
@@ -4,7 +4,7 @@
 /*
   steps to add a new builtin kernel:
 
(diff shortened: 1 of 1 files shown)


pjaaskel (approved):
LGTM

Issue #1196, opened 2023-04-03
Error on aarch64 architecture (ARM)


jandrovins:
I have the following error running `OCL_ICD_VENDORS=/pocl-prefix/etc/OpenCL/vendors ./bin/poclcc -l`: 

> `./bin/poclcc: symbol lookup error: /pocl-prefix/lib64/pocl/libpocl-devices-pthread.so: undefined symbol: pocl_ignore_sigfpe_for_thread`

I am using the Khronos ICD loader (https://github.com/KhronosGroup/OpenCL-ICD-Loader), although I  guess is not very relevant.

Using `grep` I found that the `pocl_ignore_sigfpe_for_thread` function is declared in [lib/CL/pocl_util.h line 63](https://github.com/pocl/pocl/blob/c141fb77ab7e8e219848edcc5576be2bc2ba3213/lib/CL/pocl_util.h#L63)  and defined at [lib/CL/devices/signal_handlers.c line 153](https://github.com/pocl/pocl/blob/c141fb77ab7e8e219848edcc5576be2bc2ba3213/lib/CL/devices/signal_handlers.c#L153). From the last file, this function is inside an `#ifdef __x86_64__` so it makes sense that I do not have it on my ARM installation.

Not sure what to do. Should I just delete the `#ifdef __x86_64__` and expect it to work on my ARM?

jandrovins:
I have just noticed that in known issues (http://portablecl.org/docs/html/conformance.html#known-issues-in-pocl-things-to-be-aware-of) it is mentioned that POCL do not really manages exceptions for integer division by 0 on architectures different from x86. 

I commented line [134 on ../lib/CL/devices/pthread/pthread_scheduler.c](https://github.com/pocl/pocl/blob/c141fb77ab7e8e219848edcc5576be2bc2ba3213/lib/CL/devices/pthread/pthread_scheduler.c#L134) and it is now apparently working. May be this line should be enclosed in a `#ifdef __x86_64__` too.

pjaaskel:
Sounds right. Please send a PR.

jandrovins:
Will do. In a couple of days

franz:
Yep, the ifdef is the correct fix. The SIGFPE code was recently changed and haven't been tested on ARM.

PR #1205, opened 2023-04-13
Fix runtime error 'undefined symbol: pocl_ignore_sigfpe_for_thread' on aarch64

jandrovins:
Solves #1196

### DIFF
--- a/lib/CL/devices/pthread/pthread_scheduler.c
+++ b/lib/CL/devices/pthread/pthread_scheduler.c
@@ -131,7 +131,9 @@ pthread_scheduler_init (cl_device_id device)
       PTHREAD_CHECK (pthread_create (&scheduler.thread_pool[i].thread, NULL,
                                      pocl_pthread_driver_thread,
                                      (void *)&scheduler.thread_pool[i]));
+#ifdef __x86_64__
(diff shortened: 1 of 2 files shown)


buni-rock:
As you have already spotted, the declaration of the  **pocl_ignore_sigfpe_for_thread** is located under the  **#ifdef x86_64**. With your pull request, you have done the same for the call of the function, but the definition of this function is still available in the header file. I have suggested to guard the definition of the function in the header file so that **definition/declaration/function calls** are consistently guarded.

pjaaskel:
Thanks. Can you add a mention of this fix to CHANGES, it might interest others trying to use PoCL in ARM?

franz:
@pjaaskel  i think this should not be in CHANGES - this bug was introduced in March commit 94546bc787371a78a855c2f01187cf, so it's not a change from previous release. Pretty sure PoCL 3.1 release works (reasonably) on ARM64, [it was tested](https://github.com/pocl/pocl/wiki/Release-testing-of-pocl-3.1).

jandrovins:
Added guard to https://github.com/pocl/pocl/blob/c141fb77ab7e8e219848edcc5576be2bc2ba3213/lib/CL/pocl_util.h#L63 as suggested by @buni-rock

PR #1266, opened 2023-06-20
Fixes

anbe42:
Make the `LLC_HOST_CPU` behavior more consistent.
 - If it is set by the user, this should always set `HOST_CPU_FORCED` (including if `LLC_HOST_CPU = LLC_HOST_CPU_AUTO`)
 - If it is set by CMake, CMake would find the cached value on a subsequent run and turn on `HOST_CPU_FORCED` on the next run, therefore use a different variable internally. Feel free to use a better variable name than `_LLC_HOST_CPU`.


franz:
Hi, thanks for the contribution. Do you have any more fixes you plan to submit for 4.0 release ?

anbe42:
I think that's it for now.

anbe42:
Perhaps that patch could be upstreamed, fixing another longstanding i386 test failure:
https://salsa.debian.org/opencl-team/pocl/-/blob/main/debian/patches/test_structs_as_args.patch
(needs to be refreshed and simplified and cleaned up for release_4_0)
I had hoped to get that fixed via the headers, but there seems to be no progress, so maybe we can work around it in the pocl test.

And then I have that one as well:
https://salsa.debian.org/opencl-team/pocl/-/blob/main/debian/patches/timeout.patch
(some test was miscompiled (may have been a llvm issue on an exotic architecture)) and got into an infinite loop ... causing the build process to be forcibly terminated (by the builder infrastructure) after a few hours with a log missing the interesting final part.)

And some wishlist idea: can we have something that allows me something like
`cmake -DVERSION_SUFFIX_VENDOR_TEXT=+debian ...`
and then in CMakeLists.txt `set(VERSION_SUFFIX "${VERSION_SUFFIX_FIXED_TEXT}${VERSION_SUFFIX_VENDOR_TEXT}")`
(right now I'm patching CMakeLists.txt to add my `+debian` to `VERSION_SUFFIX_FIXED_TEXT`)
(since I have some behavior changing patches applied, I'd like to mark that with a vendor suffix in the internal version number)

franz:
> Feel free to use a better variable name than _LLC_HOST_CPU

Thanks, i've renamed the variable and merged the PR.

PR #1268, opened 2023-06-20
cpu selection

anbe42:
These patches have been part of the Debian packages for more than a year already with no known problems.

With my distribution maintainer hat on ... I like the PoCL distro build (performance wise, makes good use of the cpu) and at the same time dislike compiling stuff for the native target since it may make debugging target dependent bugs difficult (not that I'm aware of any such bug, yet). Also the tests only cover one of the many targets of the distro build - that best matches the buildd CPU.
Targets that require only a subset of the features provided by the host CPU should be testable on that machine, too. E.g. a CPU capable of AVX512 can test SSE2 (only), AVX, AVX2, ... too.

I've therefore implemented some changes ... allowing overriding the CPU detection at runtime (but only to targets supported on the host CPU, falling back to the host CPU).

For each kernellib variant that gets built, tie it to a target cpu (i.e. the one used as target for building the kernellib) and use that cpu as code generation target instead of `native`. The differences should only be marginal, since the kernellib selection tries to do a best match for the host cpu. But it significantly reduces the possible runtime code generation targets and would allow testing all of them with only a minimal number of (physical) CPUs (ideally an Intel one and an AMD one).

Running the tests for all targets is currently only part of the Debian build scripts, I haven't tried doing that in CMake, yet.
(Effectively its a loop over all targets, setting an environment variable and running the tests for each target.)

I can also rebase that on release_4_0 if you prefer.

### DIFF
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -1660,13 +1660,19 @@ pocl_init_default_device_infos (cl_device_id dev)
 #ifdef ENABLE_LLVM
 
   dev->llvm_target_triplet = OCL_KERNEL_TARGET;
-#ifdef HOST_CPU_FORCED
+#if defined(KERNELLIB_HOST_DISTRO_VARIANTS)
+  dev->kernellib_name = pocl_get_distro_kernellib_name ();
+  dev->llvm_cpu = pocl_get_distro_cpu_name (dev->kernellib_name);
+#elif defined(HOST_CPU_FORCED)
+  dev->kernellib_name = OCL_KERNEL_TARGET_CPU;
   dev->llvm_cpu = OCL_KERNEL_TARGET_CPU;
 #else
+  dev->kernellib_name = NULL;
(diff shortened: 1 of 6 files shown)


pjaaskel:
Thanks. Let's postpone this to the next release as it's a bit intrusive and a new feature, not a bug fix.

This reminds me of the old chat we had a long time ago about making the kernel compilation more portable by (optionally) always building the called builtins from sources instead of linking in the target specific bitcode. It might be an interesting alternative when going forward, but requires splitting the builtins to more files, indexing them, and adding an LLVM pass that extracts all required builtins from the kernel bitcode to minimize the compiler frontend overheads (the naive solution would just always compile in everything).

franz:
Taken a quick look, nothing really stands out, except...  we should avoid using `POCL_ABORT` for error handling, **especially** in device discovery / device init phase. The much nicer strategy is to 1) log an error 2) set the device->available to CL_FALSE. PoCL's `clGetDeviceIDs` will not return unavailable devices.

PR #1272, opened 2023-06-29
Misc fixes

anbe42:
unfortunately I only spotted the two badly named symbols after the 4.0 release ...

work around another test failing on i386 since this does not seem to get fixed in the headers


pjaaskel (approved):
LGTM

PR #1276, opened 2023-07-17
Corrected clGetCommandBufferInfoKHR with CL_COMMAND_BUFFER_QUEUES_KHR query

shajder:
removed unnecessary `&` which causes crashes during `memcpy` operation


jansol (approved):
Good catch! Looking at the code it appears that `CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR` a few lines below is affected by the same problem so if you don't mind fixing that here as well it'd be nice to get that squared away in the same go.

PR #1290, opened 2023-08-29
Enforce requirement of c++11 compiler for device library

JayFoxRox:
C++ required here (other similar cases exist), for creating an `std::vector` in arguments:

https://github.com/pocl/pocl/blob/5a99e12d0bb78427ad948ac368589135507f1b59/lib/CL/devices/builtin_kernels.cc#L23-L25

This fails on the default settings for some compilers / toolchains with older standard.
On macOS, [the brew package manager actually used a different compiler for building pocl](https://github.com/Homebrew/homebrew-core/blob/191b3918d53ae9f1527315755f7282bc0c6a7022/Formula/p/pocl.rb#L34-L44).

### DIFF
--- a/lib/CL/devices/CMakeLists.txt
+++ b/lib/CL/devices/CMakeLists.txt
@@ -23,6 +23,9 @@
 #
 #=============================================================================
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
 if(ENABLE_LOADABLE_DRIVERS)
 
   function(add_pocl_device_library name)


franz:
Thanks. C++11 requirement is already set in the top-level `CMakeLists.txt`, i think it's OK if you just add `CMAKE_CXX_STANDARD_REQUIRED` there instead of `lib/CL/devices`.

pjaaskel:
Good catch. The intention is that we could build most of the core PoCL runtime with C99, so in this case only if the device needs built-in-kernels it would require the C++11 support. Something more to fix in the "Nano-PoCL"-build.

JayFoxRox:
> Thanks. C++11 requirement is already set in the top-level CMakeLists.txt, i think it's OK if you just add CMAKE_CXX_STANDARD_REQUIRED there instead of lib/CL/devices.

I think it's the opposite - I probably only have to set the `CMAKE_CXX_STANDARD` again after this:

https://github.com/pocl/pocl/blob/5a99e12d0bb78427ad948ac368589135507f1b59/lib/CL/CMakeLists.txt#L25-L30

I'll probably try that in the next days (am a bit busy, but plan to finish this PR)

franz:
@JayFoxRox  hmm i forgot about that hack. Unfortunately it is required (for all LLVM related sources), but it should only unset the standard for the `lib_cl_llvm` library, not the entire `lib/CL`. If you could change the code to save, unset, call `add_library(lib_cl_llvm ..)` & then re-set the `CMAKE_CXX_STANDARD` variable, that would be ideal.

Another thing that needs to be removed is `set(CMAKE_CXX_STANDARD 17)` from `lib/llvmopencl/CMakeLists.txt`, that doesn't belong there at all.

franz:
This has been fixed in main branch.

PR #1296, opened 2023-09-04
remote: Improve testing infrastructure & coverage

jansol:
This adds a pair of wrapper scripts for running (some of) the existing tests with the remote driver using either one or two server daemons (which in turn use the PoCL cpu driver as their native driver).

To kick off slightly more systematic testing, this script is applied to the set of the tests under `tests/runtime` that passed without further modifications.


examples/example1/CMakeLists.txt
```diff
@@ -94,3 +94,16 @@ set_tests_properties( "examples/example1_poclbin"
     ENVIRONMENT "POCL_KERNEL_CACHE=0"
     LABELS "poclbin;internal;level0"
     DEPENDS "pocl_version_check")
+
+if(ENABLE_REMOTE_CLIENT AND ENABLE_REMOTE_SERVER AND ENABLE_HOST_CPU_DEVICES)
+add_test(NAME "remote/example1"
```
pjaaskel, line 99: Indent missing?

tools/scripts/run_remote_tests
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# run_remote_tests - Runs tests against the remote driver, using the PoCL cpu
```
pjaaskel, line 2:
```suggestion
# run_remote_tests - Runs tests against the remote driver, using the PoCL-CPU
```

tools/scripts/test_remote_runner_multi.sh
```diff
@@ -1,25 +1,27 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pjaaskel, line 2: License text missing.

tools/scripts/test_remote_runner_multi.sh
```diff
@@ -1,25 +1,27 @@
...
 
 BUILD_DIR=$1
+TEST_BINARY=$2
+shift 2
 
-PORT1=$((3000 + $RANDOM % 6000))
-PORT2=$(($PORT1 + 10))
+PORT1=11000
```
pjaaskel, line 7: Is this better than random?

tools/scripts/test_remote_runner_single.sh
```diff
@@ -1,8 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pjaaskel, line 2: License

pjaaskel (approved):
Hm. I thought the ASAN subdevice failure was fixed? Did you not include the fix yet? Perhaps include it in this one?

tools/scripts/test_remote_runner_multi.sh
jansol: We can't guarantee there won't be collisions with random anyway, and this should make errors more deterministic and therefore easier to spot.

jansol:
The ASAN problem *was* fixed, this seems to be different.

jansol:
Couldn't repro the TSAN failure locally but fixed some mildly fishy albeit not obviously problematic accesses in the function the error pointed towards. Also switched the basic tests to RelWithDebInfo instead of Release builds so if it happens again the backtraces should hopefully be a bit more informative.

pjaaskel:
Hm. a subgroups CTS build failure all of a sudden?

PR #1302, opened 2023-09-14
proxy: add pocl extensions to rename_opencl.h

RABijl:
This allows access to pocl extensions in scenarios where you make use of rename_opencl.h

### DIFF
--- a/include/rename_opencl.h
+++ b/include/rename_opencl.h
@@ -152,4 +152,7 @@
 #define clCommandFillBufferKHR POclCommandFillBufferKHR
 #define clCommandFillImageKHR POclCommandFillImageKHR
 
+/* PoCL extensions */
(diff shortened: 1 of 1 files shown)


include/rename_opencl.h
```diff
@@ -152,4 +152,9 @@
 #define clCommandFillBufferKHR POclCommandFillBufferKHR
 #define clCommandFillImageKHR POclCommandFillImageKHR
 
+/* PoCL extensions */
+#define clSetContentSizeBufferPoCL POclSetContentSizeBufferPoCL
```
pjaaskel, line 156: Only this one is the main branch, please remove the rest, we might not publish them at all, they are from our experiments.

pjaaskel (changes requested):

include/rename_opencl.h
RABijl: i've removed them as requested

Issue #1311, opened 2023-09-23
examples: cmake of earlier version complains  for `-E rm` option 


zhoumin2:
I found that there were some failures when cmake of version ```3.13.4``` was used to run `make test` as following:
```bash
CMake Error at /home/user/projects/pocl/cmake/multi_exec_test.cmake:12 (message):
  FAIL: Command exited with nonzero code (1):

  STDOUT:



  STDERR:

  CMake Error: cmake version 3.13.4

  Usage: /usr/bin/cmake -E <command> [arguments...]

  Available commands:

    capabilities              - Report capabilities built into cmake in JSON format
    chdir dir cmd [args...]   - run command in a given directory
    compare_files file1 file2 - check if file1 is same as file2
    copy <file>... destination  - copy files to destination (either file or directory)
    copy_directory <dir>... destination   - copy content of <dir>... directories to 'destination' directory
    copy_if_different <file>... destination  - copy files if it has changed
    echo [<string>...]        - displays arguments as text
    echo_append [<string>...] - displays arguments as text but no new line
    env [--unset=NAME]... [NAME=VALUE]... COMMAND [ARG]...
                              - run command in a modified environment
    environment               - display the current environment
    make_directory <dir>...   - create parent and <dir> directories
    md5sum <file>...          - create MD5 checksum of files
    sha1sum <file>...         - create SHA1 checksum of files
    sha224sum <file>...       - create SHA224 checksum of files
    sha256sum <file>...       - create SHA256 checksum of files
    sha384sum <file>...       - create SHA384 checksum of files
    sha512sum <file>...       - create SHA512 checksum of files
    remove [-f] <file>...     - remove the file(s), use -f to force it
    remove_directory dir      - remove a directory and its contents
    rename oldname newname    - rename a file or directory (on one volume)
    server                    - start cmake in server mode
    sleep <number>...         - sleep for given number of seconds
    tar [cxt][vf][zjJ] file.tar [file/dir1 file/dir2 ...]
                              - create or extract a tar or zip archive
    time command [args...]    - run command and display elapsed time
    touch file                - touch a file.
    touch_nocreate file       - touch a file but do not create it.
    create_symlink old new    - create a symbolic link new -> old


```
It seems that cmake of version ```3.13.4``` supports `cmake -E remove` instead of `cmake -E rm` and cmake of higher version supports both commands.

PR #1312, opened 2023-09-23
examples: fix cmake error caused by `-E rm` option

zhoumin2:
#1311


franz:
Hello, thanks for the PR. The bad news is that the fix is not correct; the 'remove' command of CMake does not recognize '-r' (recursive) option, so it does nothing and returns an error. The good news is that the correct fix is even easier, you can just remove the relevant lines with `-DCMD2=...` from the cmake files. The use of `Testing/Cache` directory was removed in commit f0c52a100f3.

PR #1317, opened 2023-10-02
remote+pocld: deduplicate hostname resolution and socket options code

### LINKED ISSUE #1304: pocld silently fails to receive connections when using 'localhost' instead of 127.0.0.1
From the client side, both the hostname and the ip work.

jansol:
This should unify the behaviour everywhere, particularly when passing `localhost` or `ip6-localhost` as the server address. Also deduplicated the client socket options setting code while I was at it.

Fixes #1304

### DIFF
--- a/lib/CL/devices/remote/CMakeLists.txt
+++ b/lib/CL/devices/remote/CMakeLists.txt
@@ -24,7 +24,12 @@
 #=============================================================================
 
 if(MSVC)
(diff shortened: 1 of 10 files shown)


pjaaskel (approved):

pjaaskel:
LLVM 16 fails `conformance_main_c11_atomics_micro_3 (SEGFAULT)`, seems the atomics is still flaky.

PR #1319, opened 2023-10-05
CTS fixes

franz:


CMakeLists.txt
pjaaskel, line 1: Can you postpone this commit to the beginning of the next release cycle and instead add a deprecation note to the ReleaseNotes that we are going to remove it in the next release?

lib/CL/pocl_llvm_utils.cc
```diff
@@ -366,7 +366,7 @@ std::string CurrentWgMethod;
 
 static bool LLVMInitialized = false;
 static bool LLVMOptionsInitialized = false;
-
+static bool LLVMUseGlobalContext = true;
```
pjaaskel, line 369: Does the separate LLVM context per context break with the supported LLVM versions? Do we still need the global context, isn't one per cl context a better option if it works?

lib/kernel/printf.c
```diff
@@ -38,9 +38,19 @@
  *
  * For debugging
  * Use as: DEBUG_PRINTF((fmt, args...)) -- note double parentheses!
-
-int printf(const char *format, ...);
-#define DEBUG_PRINTF(args) (printf args)
+ *
```
pjaaskel, line 41: Can you use #if 0, slightly nicer.

examples/PyOpenCL/CL_UNORM_INT24.patch
pjaaskel, line 1: Is this a PyOpenCL path which should upstream?

pjaaskel (changes requested):

CMakeLists.txt
franz: OK

examples/PyOpenCL/CL_UNORM_INT24.patch
franz: yeah we should

lib/CL/pocl_llvm_utils.cc
franz: It shouldn't break with any LLVM version, but it's not necessarily always better. It can lead to extreme memory usage, if the program creates lots of cl_context objects and doesn't clean them up. IIRC this was a problem with at least PyOpenCL.

pjaaskel: Ah right, LLVM can be large.

lib/kernel/printf.c
franz: fixed

examples/PyOpenCL/CL_UNORM_INT24.patch
franz: filed [PyOpenCL issue 705](https://github.com/inducer/pyopencl/issues/705)

PR #1323, opened 2023-10-10
basic: don't assume other threads won't notify events

jansol:
Events are notified in locked state so unlock them before scheduling which tries to notify them again which in turn tries to lock the event.


franz:
LGTM

franz (approved):

PR #1325, opened 2023-10-12
A couple of small conformance fixes to program queries

### LINKED ISSUE #1313: clGetProgramInfo returns the wrong number of binaries
according to the spec found [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_program_object_queries), CL_PROGRAM_BINARIES should return an array of the same size as the array returned from CL_PROGRAM_DEVICES. However, if there are two devices and only one is built, the number of binaries returned is one. Below is an example program that highlights the problem, I've tried this with two basic and two remote devices and got the same result. I tested this on the main branch.

```
#include <stdio.h>
#include <string.h>
#include <assert.h>
#include <CL/cl.h>

int main ()
{

  const char *source = " __kernel \n"
                       "void copy(__global int *input, __global int *output)\n"
                       "{\n"
                       "    int index = get_global_id(0);\n"
                       "    output[index] = input[index];\n"
                       "}";

  int status;
  cl_uint num_devs;
  cl_device_id devices[2];
  cl_platform_id platform;
  char result_array[256];
  cl_context context;
  cl_program program;

  status = clGetPlatformIDs (1, &platform, &num_devs);

  status = clGetDeviceIDs (platform, CL_DEVICE_TYPE_ALL, 2, devices, &num_devs);

  for(int i =0; i < num_devs; i++) {
      clGetDeviceInfo (devices[i], CL_DEVICE_NAME, 256 * sizeof (char), result_array, NULL);
      printf ("device name: %s\n", result_array);
  }

  cl_context_properties cps[] = {CL_CONTEXT_PLATFORM, (cl_context_properties) platform,
                                 0};
  context = clCreateContext (cps, 2, devices, NULL, NULL, &status);

  size_t len = strlen (source);
  program = clCreateProgramWithSource (context, 1, &source, &len, &status);

  status = clBuildProgram (program, 1, devices, NULL, NULL, NULL);

  printf("devices: %i \n", num_devs);
  size_t param_value_size_ret;

  clGetProgramInfo (program, CL_PROGRAM_DEVICES, 0, NULL, &param_value_size_ret );
  printf ("%lu, %lu\n", param_value_size_ret, param_value_size_ret/sizeof (cl_device_id));

  int devices_size = param_value_size_ret/sizeof (cl_device_id);

  clGetProgramInfo (program, CL_PROGRAM_BINARIES, 0, NULL, &param_value_size_ret );
  printf ("%lu, %lu\n", param_value_size_ret, param_value_size_ret/sizeof (unsigned char *));

  int binaries_size = param_value_size_ret/sizeof (unsigned char *);

  assert(binaries_size == devices_size);

  return 0;
}
```

On a related note, trying to get program binaries like the snippet below results in a segfault.

```
  size_t param_value_size = 2 * sizeof(unsigned char *);
  void * param_value = malloc(param_value_size);

  clGetProgramInfo (program, CL_PROGRAM_BINARIES, param_value_size, param_value, &param_value_size_ret );
```

### LINKED ISSUE #1321: Local devices are passed to the remote in clBuildProgram
When building a program with clBuildProgram containing both local and remote devices, they are both passed to the remote, even though the remote should see only the remote device. It triggers this assert https://github.com/pocl/pocl/blob/00a5c19359959dfeac9c57c41160b2e55e5a8118/pocld/shared_cl_context.cc#L1151 (builtProgramDevices has size 2, binaries has size 1). The solution could be to modify the client code to filter out client-only devices before passing them to the remote.

cc @jansol

jansol:
Fixes #1313
Fixes #1321

### DIFF
--- a/lib/CL/clGetProgramBuildInfo.c
+++ b/lib/CL/clGetProgramBuildInfo.c
@@ -87,9 +87,6 @@ POname(clGetProgramBuildInfo)(cl_program            program,
     
   case CL_PROGRAM_BUILD_LOG:
     {
-      POCL_RETURN_ERROR_ON((program->build_status == CL_BUILD_NONE),
-                           CL_INVALID_PROGRAM,
-                           "Program was not built");
       if (program->builtin_kernel_names != NULL)
         {
           POCL_RETURN_GETINFO_STR ("");
--- a/lib/CL/clGetProgramInfo.c
+++ b/lib/CL/clGetProgramInfo.c
@@ -32,43 +32,76 @@
 
 static void get_binary_sizes(cl_program program, size_t *sizes)
 {
-  unsigned i;
-  for (i=0; i < program->num_devices; i++)
+  unsigned assoc_i, dev_i;
+  for (assoc_i = 0; assoc_i < program->associated_num_devices; assoc_i++)
     {
-      if (program->devices[i]->ops->build_poclbinary)
-        program->devices[i]->ops->build_poclbinary (program, i);
+      int program_device = 0;
+      for (dev_i = 0; dev_i < program->num_devices; ++dev_i)
+        {
+          if (program->devices[dev_i] == program->associated_devices[assoc_i])
+            {
+              program_device = 1;
+              break;
+            }
+        }
+      if (!program_device)
+        {
+          sizes[assoc_i] = 0;
+          continue;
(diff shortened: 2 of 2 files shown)


franz:
LGTM

PR #1329, opened 2023-10-20
Pocld socket rework

jansol:
Move from per-client per-socket threads to a single thread that `poll()`s all client sockets. Additionally requests are now read incrementally to avoid large requests from blocking smaller commands.

### DIFF
--- a/pocld/pocld.cc
+++ b/pocld/pocld.cc
@@ -37,6 +37,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+#include <poll.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/un.h>
@@ -66,9 +67,10 @@
 #include "rdma.hh"
 #endif
 
-VirtualContextBase *createVirtualContext(client_connections_t conns,
-                                         ClientHandshake_t handshake);
-void startVirtualContextMainloop(VirtualContextBase *ctx);
+#ifndef POLLRDHUP
+#define PULLRDHUP 0
+#endif
+#define POLLFD_ERROR_BITS (POLLHUP | POLLERR | POLLNVAL | POLLRDHUP)
 
 #define PERROR_CHECK(cond, str)                                                \
   do {                                                                         \
@@ -77,6 +79,10 @@ void startVirtualContextMainloop(VirtualContextBase *ctx);
       return 1;                                                                \
     }                                                                          \
   } while (0)
+
+VirtualContextBase *createVirtualContext(client_connections_t conns,
+                                         ClientHandshake_t handshake);
+void startVirtualContextMainloop(VirtualContextBase *ctx);
 
 /***************************************************************************/
 /***************************************************************************/
@@ -116,8 +122,8 @@ in_addr_t find_default_ip_address() {
     freeifaddrs(ifa);
   } else
     POCL_MSG_ERR("getifaddrs() failed or returned no data.\n");
-
 #endif
+
   return listen_addr.s_addr;
 }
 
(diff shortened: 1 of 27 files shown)


pocld/request.hh
```diff
@@ -0,0 +1,111 @@
...
+#include <cstring>
+
+#ifdef __GNUC__
+#pragma GCC visibility push(hidden)
+#endif
+
+class Request {
+
+public:
+  uint32_t req_size;
```
pjaaskel, line 38: I know these are not new attributes, but again, would help readability if they were documented.

pjaaskel (commented):
As the communication code is becoming a bit more complex, it would be useful to have a centralized comment of how all of it plays together. I'll test the branch locally.

pocld/pocld.cc
```diff
@@ -237,323 +239,575 @@ int listen_rdmacm_events(rdmacm::EventChannelPtr cm_channel,
...
+          return false;
+        } else if (ev & POLLIN) {
+          struct sockaddr client_address;
+          socklen_t client_address_length = sizeof(client_address);
+          /* NOTE: address length MUST be initialized to the size of the storage
+           * given as the addr argument */
+          int newfd = accept(pfd.fd, &client_address, &client_address_length);
+          if (newfd > 0) {
+            // XXX: rework (eliminate?) handshakes and just store new fds
+            hack = newfd;
```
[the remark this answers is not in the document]
jansol: It's returning the new fd from the function via out parameter so it can be paired with the other fd ("command" and "stream") of the same client. Once requests contain session IDs this is not necessary (and we won't have race conditions when multiple clients try to connect simultaneously)

pjaaskel:
pocld is nicely not actively polling any longer, which is very nice. But `remote/example1_spirv` fails.

jansol:
`remote/example1_spirv` passes locally for me.

pjaaskel:
Or can we use fuser or similar too to wait until the port is free. Or my previous suggestion of adding the binding retry logic to pocld instead?

pocld/pocld.cc
```diff
@@ -239,33 +239,90 @@ int listen_rdmacm_events(rdmacm::EventChannelPtr cm_channel,
...
+   * (command, stream) pairs have been obtained, the client handshake is
+   * performed and the fds are associated with a new or existing client context
+   * based on the handshake.
+   *
+   * After the special case of the listener sockets, if there are further events
+   * in the poll result, the respective fds are handed to Result::read for
+   * reading a piece of the next command sent over that socket. Similar to the
+   * pollfd list, the function keeps a list of in-flight Requests for this
+   * purpose. If there are any read errors or poll results indicating that a
+   * socket was closed, the corerspondign fd is pushed into a list for cleanup.
```
pjaaskel, line 302:
```suggestion
   * socket was closed, the corresponding fd is pushed into a list for cleanup.
```

pocld/pocld.cc
```diff
@@ -239,33 +239,90 @@ int listen_rdmacm_events(rdmacm::EventChannelPtr cm_channel,
...
+  void waitForExit() {
+    if (client_sockets_th.joinable())
+      client_sockets_th.join();
+  }
+
   /* returns client context on success, nullptr on error */
   VirtualContextBase *performClientSetup(int command_fd, int stream_fd);
 
 private:
+  /** File descriptor for the socket that listens for incoming cliend
```
pjaaskel, line 319:
```suggestion
  /** File descriptor for the socket that listens for incoming client
```

pocld/pocld.cc
```diff
@@ -578,6 +633,16 @@ void PoclDaemon::readAllClientSocketsThread() {
...
     /* We always put our client listener sockets in the list first so let's
      * handle accepting incoming connections separately here. */
+    /*
+     * In order to associate the correct "command" and "stream" socket fds with
+     * the same context, we hold onto the last opened fd of each listener socket
+     * (via the out parameter in accept_new_connection out) and once we have
+     * both, create a new context with them (or update and existing one,
+     * depending on handshake). This is obviously a race condition and should be
+     * eliminated, but that will require redesigning the handshake process so
+     * until that happens please try to avoid several clients connecting at the
```
pjaaskel, line 643: Something to put high up in the laundry list.

tools/scripts/test_remote_runner_multi.sh
```diff
@@ -102,4 +103,6 @@ kill -9 $EXAMPLE_PID 1>/dev/null 2>&1
 kill -9 $POCLD_PID1 1>/dev/null 2>&1
 kill -9 $POCLD_PID2 1>/dev/null 2>&1
 
+wait -f
```
pjaaskel, line 106: Good to know of such a thing!

pjaaskel (changes requested):

PR #1331, opened 2023-10-23
clGetContextInfo: param_value_size_ret may be NULL

johningve:
According to the OpenCL spec, param_value_size_ret may be NULL and should be ignored in that case. This commit fixes one case that did not check if the pointer is NULL before writing to it.


pjaaskel:
Thanks!

PR #1336, opened 2023-11-03
command buffer fix + CTS fix

franz:


jansol:
:+1: I was sure the command node cloning was subtly wrong but never figured out what the correct form was.

PR #1348, opened 2023-11-11
HSA fixes

Oblomov:
This is a draft/WIP patchset to update the HSA device, which I'm throwing out more as a RFC than as an actual candidate for merging. So far the biggest stopgap is that I'm unable to actually compile the kernel library because compilation fails due `size_t` being undefined. I'm not familiar enough with the kernel library compilation mechanism to fix this, so any recommendation in this regard is welcome.

(My aim is to reach support for my iGP in PoCL, and this approach sounds a bit more approachable than a HIP/ROCm device built as a clone of the CUDA device, even though I'm also slowly working on that too.)


pjaaskel:
Thanks for working on AMD support. To be honest, I'm not sure if the best starting point is to build on the old HSA driver or create a new ROCr driver or such. I haven't followed recently how closely AMD still follows the HSA interfaces as specified by HSAF.  

For the compilation problem, I'm not sure which problem you are seeing. If size_t is not defined by default (if you compile in C mode and you don't include stddef.h) you might want to define that to a 64b unsigned int type somewhere early.  I'd just try to copy the CUDA build configurations for the bitcode lib and start building from there towards the AMD HSA lib.

Oblomov:
The first 5 (or 7?) commits are independent from AMD specifically and probably worth looking into regardless, at least if the HSA device still has any relevance at all (some of the changes are just “unbreak host-side compilation due to function signature changes” (I'm getting the impression this device isn't tested often ;-)).

I'll confess to be extremely ignorant of how the bitcode library is built in the first place. I'll try to see what the CUDA device does and see if I can find inspiration.

PR #1353, opened 2023-11-15
Initial steps for CG SVM on PoCL-R

pjaaskel:
Still a WiP so disabled in code meanwhile. It currently works only if both the server and the client manage to map the SVM pool to the same address range. This seems to happen often enough for this state to be useful for testing and further development.

### DIFF
--- a/lib/CL/devices/bufalloc.c
+++ b/lib/CL/devices/bufalloc.c
@@ -1,12 +1,13 @@
 /* OpenCL runtime/device driver library: custom buffer allocator
 
    Copyright (c) 2011-2020 pocl contributors
+                 2023 Pekka Jääskeläinen / Intel Finland Oy
 
    Permission is hereby granted, free of charge, to any person obtaining a copy
-   of this software and associated documentation files (the "Software"), to deal
-   in the Software without restriction, including without limitation the rights
-   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-   copies of the Software, and to permit persons to whom the Software is
+   of this software and associated documentation files (the "Software"), to
+   deal in the Software without restriction, including without limitation the
+   rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+   sell copies of the Software, and to permit persons to whom the Software is
    furnished to do so, subject to the following conditions:
 
    The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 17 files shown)


pocld/shared_cl_context.cc
```diff
@@ -495,10 +512,10 @@ SharedCLContext::SharedCLContext(cl::Platform *p, unsigned pid,
                                  ReplyQueueThread *s, ReplyQueueThread *f) {
   p->getDevices(CL_DEVICE_TYPE_ALL, &CLDevices);
 
-  cl_context_properties properties[] = {
+  cl_context_properties Properties[] = {
```
jansol, line 515: There's a bunch of other local C++ variables that are getting capitalized in this PR too, what's the logic behind that?

jansol (approved):
LGTM, although I am not too familiar with SVM. Some minor comments.

lib/CL/devices/remote/remote.c
```diff
@@ -233,6 +306,120 @@ pocl_remote_update_event (cl_device_id device, cl_event event)
...
+ * device regions where the SVM allocations are mapped. This should be called
+ * after a point where we have initialized enough of the device that we can use
+ * it for buffer creation.
+ *
+ * \param [inout] device the device to update the pool info to.
+ * \returns zero on a successful mapping.
+ *
+ */
+static int
+setup_svm_memory_pool (cl_device_id device)
```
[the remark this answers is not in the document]
pjaaskel: Perhaps. I haven't really thought through the scenario with multiple devices yet, but indeed if those devices are able to create a shared CG SVM region together then it definitely is the way to go.

pocld/shared_cl_context.cc
pjaaskel: I'm unsure about this. PoCL-D was not following the PoCL C++ coding standard (LLVM-style) when it was added and I'm leaning towards converting it step-by-step to the style (like with the rest of the PoCL) code space instead of adding a 3rd coding style just for PoCL-D. But we can do the latter as well if you think it's not worth having mixed styles over "the migration".

lib/CL/devices/bufalloc.c
```diff
@@ -166,7 +169,7 @@ append_new_chunk (memory_region_t *region,
   DL_APPEND (region->chunks, region->last_chunk);
 
 #ifdef DEBUG_BUFALLOC
-  printf ("#### after append_new_chunk (%x, %u)\n", region, size);
+  printf ("#### after append_new_chunk (%p, %lu)\n", region, size);
```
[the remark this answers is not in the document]
pjaaskel: Will that in the next PR.

pocld/shared_cl_context.cc
jansol: Nah I was just not aware that LLVM style requires capitalized variable names. Apparently clang-format does not enforce this either. But since it is in the style guide let's keep it in mind and convert the code to adhere to it as we go.

PR #1354, opened 2023-11-20
Implement cl_khr_icd 2.0.0 loader managed dispatch

Kerilk:
This implements the loader managed dispatch strategy presented here: https://github.com/KhronosGroup/OpenCL-Docs/issues/1003

### DIFF
--- a/lib/CL/CMakeLists.txt
+++ b/lib/CL/CMakeLists.txt
@@ -181,6 +181,7 @@ set(POCL_LIB_SOURCES  "clCreateContextFromType.c"
                    "clEnqueueMemcpyINTEL.c"
                    "clEnqueueMemAdviseINTEL.c"
                    "clEnqueueMigrateMemINTEL.c"
+                   "clGetFunctionAddressForPlatformKHR.c"
 )
 
 if(ANDROID)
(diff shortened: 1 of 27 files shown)


pjaaskel:
Thanks Brice! I'll launch the CIs. I added some questions in the [extension PR](https://github.com/KhronosGroup/OpenCL-Docs/pull/1005#pullrequestreview-1745888806). @franz can you add an ICD2-compliant ICD loader to one of the CIs?

franz:
I can add it as soon as it lands in ocl-icd's ubuntu ppa.

Kerilk:
Thanks for the reviews. Not sure when this will be ready. I want to make sure I can implement instance layers on top of this (WIP is here https://github.com/Kerilk/pocl/tree/instance).
Also, I don't know if pocl is unloadable. I know it does not cleanup cleanup the `pocl_devices` global variable, so I would lean on the side of no. I will add a query to the platform `CL_PLATFORM_UNLOADABLE_KHR` and have it return no for now.

franz:
@Kerilk it does not cleanup everything, but it does have cleanup - *some* drivers implement the cleanup hooks, and also there is code to unload LLVM (as much as possible). It's disabled by default, IIRC there were some problems with reliability, and there wasn't any official API to unload the drivers / platforms, only the compiler. However if you try `export POCL_ENABLE_UNINIT=1` and in the program call `clUnloadPlatformCompiler` after releasing all `cl_*` objects, PoCL should (at least with CPU driver) stop the driver threads and unload almost everything (except some static variables), to the point that both Valgrind and AddressSanitizer show no memory leaks. It should be possible to release remaining static variables in PoCL (if it's needed), but obviously we can't do anything about static variables of libLLVM or other libraries PoCL links to.

pjaaskel:
There was some changes planned still to the icd2 in the OpenCL WG so let's revert this one. @franz can you do it in your next PR (my local PoCL repo is unclean)?

PR #1364, opened 2023-11-29
Somewhat usable remote CG SVM

pjaaskel:
Not targeting for 5.0, but let's aim merging this in soon after the branch is done.

SVM is still enabled by luck (if the moon and both side mmaps are in the correct phase) only, but when the client and server-side offsets match, migrations, memcpys etc. seem to work for the simple test case by using the shadow cl_mem approach that reuses PoCL's buffer migration handling. Should be good enough for further testing to see where distributing chipStar fails next and start building mitigations for the offset != 0 case.

Also enabled FG system SVM for CPU which for some reason was not advertised. Let's see if any of the CTS tests explode with that.

### DIFF
--- a/include/pocl.h
+++ b/include/pocl.h
@@ -96,7 +96,7 @@ typedef struct pocl_mem_identifier
   int is_pinned;
 
   /* The device-side memory address (if known). If is_pinned is true, this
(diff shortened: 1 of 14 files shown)


pjaaskel:
I'm preparing a better PR.

PR #1376, opened 2023-12-20
add shared lib version of pocl for android

RABijl:
This pr does the following things:
* disables a couple of things not available on android
* renames the shared lib to libpocl.so when on android
* finds and adds logging library for android
* adds some explicit imports of libraries to pocl_networking.c that were not found otherwise
* updates some function declarations in the proxy driver
* adds a new cmake option called `PROXY_USE_LIBOPENCL_STUB` which allows the proxy device to link that instead of the system libopencl library.
* add a new cmake option called `RENAME_POCL` which does many things similar to what `ENABLE_PROXY_DEVICE` used to do.
* the proxy driver can now be used without renaming all of pocl's ocl functions by using `PROXY_USE_LIBOPENCL_STUB`
* fixes the ordering of a couple of includes on test files that would fail during the build phase if the proxy device was enabled.

### DIFF
--- a/lib/CL/devices/devices.c
+++ b/lib/CL/devices/devices.c
@@ -534,8 +534,7 @@ pocl_init_devices ()
     sleep (delay);
 #endif
 
-
-#ifdef __linux__
+#if defined(__linux__) && !defined(__ANDROID__)
 
 #ifdef ENABLE_HOST_CPU_DEVICES
   if (pocl_get_bool_option ("POCL_SIGFPE_HANDLER", 1))
--- a/lib/CL/devices/pthread/pthread_scheduler.c
+++ b/lib/CL/devices/pthread/pthread_scheduler.c
@@ -565,7 +565,7 @@ pocl_pthread_driver_thread (void *p)
   assert (scheduler.local_mem_size > 0);
   td->local_mem = pocl_aligned_malloc (MAX_EXTENDED_ALIGNMENT,
                                        scheduler.local_mem_size);
-#ifdef __linux__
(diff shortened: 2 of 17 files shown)


lib/CL/devices/devices.c
```diff
@@ -535,7 +535,7 @@ pocl_init_devices ()
 #endif
 
 
-#ifdef __linux__
+#if  defined(__linux__) && !defined(__ANDROID__)
```
pjaaskel, line 538: two spaces before defined

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -565,7 +565,7 @@ pocl_pthread_driver_thread (void *p)
   assert (scheduler.local_mem_size > 0);
   td->local_mem = pocl_aligned_malloc (MAX_EXTENDED_ALIGNMENT,
                                        scheduler.local_mem_size);
-#ifdef __linux__
+#if  defined(__linux__) && !defined(__ANDROID__)
```
pjaaskel, line 568: same

pjaaskel (approved):
LGTM. But please remember to update the Android docs (after you're finished with the JOCL work).

lib/CL/devices/pthread/pthread_scheduler.c
RABijl: fixed

lib/CL/devices/devices.c
RABijl: fixed

PR #1381, opened 2023-12-22
pocld: reorder code to be a bit more modular

jansol:
Should make future changes easier and more contained. Also replaces some remaining ad-hoc socket communication with proper Request/Reply messages to clean up the connection setup a bit. If/when client RDMA is brought back I'd like to do a similar Request/Reply there to set up that connection independently from the context setup like is now done for the fast/slow sockets.


pjaaskel:
Since you a refactoring the server-side, a couple of aspects we should look at and discuss:

* Lifetime/roles of the virtual_cl_context and shared_cl_context. They are not well documented (can you add some docs in this PR?) and I find myself constantly wondering what are the purposes of the both when trying to understand the code. Is shared_cl_context supposed to be shared across multiple client contexts? It doesn't seem to work that way currently (like I assumed when implementing the CG SVM).
* Freeing of the resources. Currently it seems to be virtual_cl_context is never freed (and by implication shared_cl_context neither), which leads to memory leaks which are critical in a server. I understand the complications due to the reconnect support, but we should add a timeout for how long to retain the context server side waiting for the client to reappear. Some of the resources could be freed at least when the client frees the client-side context (related to the above point).
* This discussion opens another (security related) question at the system architecture level: Shall we assume a single pocld instance to be per (authenticated) user, thus it _should_ be able to share a device context and associated data safely, thus we _should_ have the logic for shared_cl_context outlined in the first bullet to minimize unnecessary initializations between application runs. If we assume multiple (potentially hostile) users share a single pocld instance, it might get tricky to secure (e.g. there are a lot of asserts now which can be used to crash the server easily). So, shall there be an authenticator/proxy/session handler daemon before passing the user's request to the actual pocld instance instead of handling the auth in pocld?
 
Let's continue the discussion f2f/teams when you get back from holidays. I think we should reach clarity for these aspects before refactoring a lot back-and-forth.

jansol:
The code path for destroying a client session & freeing its resources has been there for a long time, there just hasn't been any way to actually send the shutdown message from the client when the application exits so it's not getting triggered.

pjaaskel:
Pls rebase.

lib/CL/devices/remote/communication.c
```diff
@@ -1480,57 +1467,54 @@ find_or_create_server (const char *address_with_port, unsigned port,
...
+      POCL_MSG_ERR ("Could not connect to server\n");
+      POCL_MEM_FREE (d);
+      return NULL;
+    }
 
   DL_APPEND (servers, d);
 
 #ifdef ENABLE_RDMA
-  d->use_rdma = hsr.rdma_supported;
+  d->use_rdma = 0; // hsr.m.create_session.rdma_supported;
```
pjaaskel, line 1517: Is this change on purpose? Disables RDMA?

jansol: This only disables RDMA between the client and server, which has been broken for ages and has been forced off earlier in the code for an almost equally long time.

pjaaskel:
Rebase + the remote test failure seems valid? L0 seems to be still fragile.

PR #1383, opened 2024-01-05
PoCL-R: Somewhat usable Coarse Grain SVM trial 2 + misc fixes

pjaaskel:
Implemented initial CG SVM support for PoCL-R (when mmap start addresses can be made to match):

Rework the internal SVM implementation to use a cl_mem shadow buffer. The shadow buffer will be then passed to standard cl_mem APIs for mapping, unmapping, memcpy, migrate etc.

This unifies all the automated migration, content size buffer, etc. handling code to avoid special casing for SVM buffers.

Converted PoCL-D/SVM to use SubBuffers in the server-side for managing the chunks: Using parent Buffers doesn't work as a wrapper since the cl_mems are expected to point to the beginning of the SVM buffer, which leads to undefined results when wrapping chunks from the large SVM allocation.

With this, chipStar 1.1-RC1 can run at least the simple_kernel and MatrixMultiply correctly (when "the mmoon" happens to land in the right phase). 

chipStar 1.1-RC3 added the runtime linking for enhanced portability feature, which requires cl{Compile,Link}Program support which the remote driver currently lacks.

It also starts to fix memleaks on pocld-side (see [comment](https://github.com/pocl/pocl/pull/1381#issuecomment-1878300371)) since the SVM pool can be a huge chunk of VM to leak. 

This PR requires the fixes I submitted in #1382 and includes its commits.

### DIFF
--- a/lib/CL/devices/remote/communication.c
+++ b/lib/CL/devices/remote/communication.c
@@ -1111,7 +1111,8 @@ pocl_remote_writer_pthread (void *aa)
           uint32_t msg_size = request_size (cmd->request.message_type);
 
           POCL_MSG_PRINT_REMOTE ("WRITER THR: WRITING MSG, TYPE: %u  ID: %zu  "
-                                 "EVENT: %zu  SIZE: %u + %zu + %zu + %zu\n",
+                                 "EVENT: %zu  SIZE: msg_size: %u + waitlist: "
+                                 "%zu + extra: %zu + extra2: %zu\n",
                                  cmd->request.message_type,
                                  cmd->request.msg_id, cmd->event_id, msg_size,
                                  cmd->req_waitlist_size * sizeof (uint64_t),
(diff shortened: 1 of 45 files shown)


lib/CL/devices/remote/communication.c
```diff
@@ -2034,22 +2035,23 @@ pocl_network_free_device (cl_device_id device)
 // SYNCHRONOUS
 
 cl_int
-pocl_network_create_buffer (remote_device_data_t *ddata, uint32_t mem_id,
-                            uint32_t mem_flags, uint64_t mem_size,
+pocl_network_create_buffer (remote_device_data_t *ddata, cl_mem mem,
```
jansol, line 2038: Since we are passing a whole cl_mem object now, it may be a good idea to assert that at least the most critical fields have been set to some sensible values. Either here or in remote.c.

Also CL_MEM_USES_SVM_POINTER is mentioned in messages.h as required if host_ptr is to be set, so that would be good to assert here as well.

pocld/pocld.cc
```diff
@@ -785,11 +790,14 @@ void PoclDaemon::readAllClientSocketsThread() {
           close(fd);
           std::swap(open_client_fds[i], open_client_fds.back());
           open_client_fds.pop_back();
+          std::swap(socket_contexts[i], socket_contexts.back());
```
jansol, line 793: Moving this line to be before the comment that explains it seems a bit odd

poclu/misc.c
```diff
@@ -556,7 +556,7 @@ poclu_load_program_multidev (cl_context context, cl_device_id *devices,
     {
       TEST_ASSERT (device != NULL);
       strcat (final_opts, " -x spir -spir-std=1.2");
-      err = clGetDeviceInfo (device, CL_DEVICE_EXTENSIONS, 1024, extensions,
+      err = clGetDeviceInfo (device, CL_DEVICE_EXTENSIONS, 10000, extensions,
```
jansol, line 559: This is getting a bit uncomfortably large for a stack-allocated array with an arbitrary hardcoded size. Maybe query the actual size & allocate dynamically?

jansol (approved):
Looks fairly straightforward

PR #1385, opened 2024-01-09
minor fixes

franz:


lib/CL/CMakeLists.txt
```diff
@@ -302,6 +302,10 @@ add_library("${POCL_LIBRARY_NAME}" ${LIBPOCL_OBJS})
 
 harden("${POCL_LIBRARY_NAME}")
 
+if(UNIX AND (NOT APPLE))
+  target_link_options("${POCL_LIBRARY_NAME}" PRIVATE "LINKER:--exclude-libs,ALL")
```
pjaaskel, line 306: Did you check that we get a nice set of (PoCL-only) symbols exported with this in libpocl.so even when static LLVM linkage is done?

CMakeLists.txt
```diff
@@ -894,30 +894,19 @@ endif()
 
 set(SPIRV OFF)
 
-if(ENABLE_LLVM)
-  option(ENABLE_SPIR "Enable SPIR support (default ON when available)" ON)
+# drivers that consume SPIRV directly don't need LLVM or LLVM-SPIRV;
+# those that consume SPIRV indirectly need both, but it might be useful to disable LLVM-SPIRV translator
```
pjaaskel, line 898: Reminds me that we should now drop the old LLVM versions since it's quite early in the release cycle. Can you create a single commit that does it and cleans up some LLVM version macro checks?

pjaaskel (approved):
LGTM

lib/CL/CMakeLists.txt
franz: yep i checked with `nm --extern-only libpocl.so | grep -i llvm` it now lists a much shorter list of symbols, and they are pocl's LLVM function names/types which have POCL_EXPORT.

CMakeLists.txt
franz: sure

PR #1393, opened 2024-01-17
remote: fix casting issues for Android

RABijl:


lib/CL/devices/remote/communication.c
```diff
@@ -2061,7 +2061,9 @@ pocl_network_create_buffer (remote_device_data_t *ddata, cl_mem mem,
 
   nc.request.m.create_buffer.flags = mem->flags;
   nc.request.m.create_buffer.size = mem->size;
-  nc.request.m.create_buffer.host_ptr = mem->mem_host_ptr;
+  // see https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Pointer_002dInteger-Conversion.html
```
pjaaskel, line 2064: That link doesn't seem to suggest a double cast, but just note on the use of the uintptr_t if possible (in this case we want fixed size fields so it's not). What difference does the extra cast make here?

lib/CL/devices/remote/remote.c
```diff
@@ -480,7 +480,9 @@ pocl_remote_init (unsigned j, cl_device_id device, const char *parameters)
       strncpy (exts_w_pinned + strlen (device->extensions) + 1,
                CL_POCL_PINNED_BUFFERS_EXTENSION_NAME,
                strlen (CL_POCL_PINNED_BUFFERS_EXTENSION_NAME));
-      free (device->extensions);
+      // the const char * to void * cast is fine here since this value is
```
pjaaskel, line 483: Does an Android compiler complain about the implicit cast here?

lib/CL/devices/remote/communication.c
pjaaskel: That's a better reference, but still: "If the **result cannot be represented in the target type,** the behavior is undefined". We know uint64_t can store the pointer bits, thus it should be defined? I'm not opposed to the change, it just seems strange to my eyes and I think we have multiple other spots casting ptrs directly to u64. Android's default compiler warns/errors about this?

RABijl: The compiler errors about the implicit cast, it probably will do just fine with only a cast to uint64_t. @jansol mentioned that this was safer and that something similar should probably happen in other places of the code as well.

RABijl: also, if it adds any clarity, I checked the compiler we are using for android and it is llvm 17.0.2. The automated tests that build the remote client only build for llvm 16.

jansol: There are actually *two* corner cases we need to watch out for here:
* Pointer to integer casts are implementation defined so some compilers treat pointers as intrinsically signed types [and will sign-extend when going from 32bit pointers to 64bit types](https://gcc.gnu.org/onlinedocs/gcc/Arrays-and-pointers-implementation.html). By first converting to an unsigned type of the same size we avoid the surprise sign extension.
* Undefined behaviour if the pointer size is larger than uint64_t (admittedly not something we'll run into for a good while yet)

In other words while the C spec says "pointers can be cast to any integer type", it mostly means that such code will *compile*, not that it will work as expected. The well-defined way is going through uintptr_t or intptr_t (officially since C99). I only learned about this last week myself, I wonder if the subtle intrigues of the C language will ever cease to surprise me :smile: 

There are indeed other places in the code where pointers are cast directly to uint64_t and those should also be converted to make a round trip through uintptr_t.

pjaaskel: I don't get why this makes any difference. You say that the target int types' signedness doesn't matter, it might sign extend even if the target is unsigned? Does it really matter? If we are going to use the pointer again in the 32b device, the upper bits will be dropped?

jansol: I'm saying that the compiler may treat pointers as signed types and choose to do bit extension before sign conversion. Or it may not. The only way we have to be (somewhat) sure is to explicitly tell it which one to do by asking for an intermediate step through (u)intptr_t.

If the pointers in question were always the same size it would probably not matter. But if say, your 32bit microcontroller with 32bit pointers connects to pocld running on a 64-bit machine, this is suddenly not the case any more.

pjaaskel: Yep - I understand, but I repeat my question: When would it matter since when the bits will be treated as a pointer again in the 32b machine, the upper bits are gone again? If we do pointer arithmetics on the host side with that 32b ptr casted to the 64b uint, only then it should matter, right? Anyhow, it should not harm (except the untrained eye) so just let's merge.

jansol: Yes, if the pointer is only ever treated as a pointer on the 32bit machine and not touched in its 64bit representation it likely does not matter. But if I recall correctly this particular instance is meant for SVM pointers... which would be accessed by kernels on the 64bit machine, as a 64bit pointer.

pjaaskel: True!

PR #1397, opened 2024-01-22
pocld: Fix pocld crash when running without listenning address

xvim:
The original code uses the ifa_addr member to get the address of an interface in the find_default_ip_address() function. However, for some interfaces, such as WireGuard, ifa_addr may be NULL. Therefore, it is necessary to check if ifa_addr is NULL, and if it is, skip the interface.

getifaddrs() man page: [https://man7.org/linux/man-pages/man3/getifaddrs.3.html]


jansol (approved):
Good catch!

PR #1399, opened 2024-01-26
Small fixes

jansol:
The most trivial changes from #1381


jansol:
I am starting to seriously doubt the well-being of the runner in charge of remote tests...

PR #1400, opened 2024-01-26
ci: bump 3rd party actions to get rid of deprecated nodejs dependency

jansol:
Let's try to get this done before it *actually* becomes a problem.

Failing CI runs are caused by those machines having a too old version of github-runner.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -54,7 +54,7 @@ jobs:
             config: chipstar
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: CMake
         id: cmake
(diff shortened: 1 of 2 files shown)


pjaaskel:
@jansol do you mean that we (@franz) should update the CI runners by hand?

jansol:
Yeah. Just updating the runner daemon should be enough but that can't be done remotely.

franz:
I was under the impression that the runners update themselves. According to [documentation](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) they should:

> When a new version is released, the runner application automatically updates itself when a job is assigned to the runner, or within a week of release if the runner hasn't been assigned any jobs.

I haven't manually updated any of them, so it seems 3/4 do update themselves and the rest don't, weird.

jansol:
They might have automatic updates disabled. Or that feature was only added after they were set up?

PR #1405, opened 2024-01-29
Remote session/handshake rework

jansol:
Let's try this again. Remaining bits of #1381.


pjaaskel:
Are the remote tests executed with parallel ctest runs? I still get occasional remote port binding etc. issues if I run 'make check' instead of 'ctest -j1'.

jansol:
I don't know if `make check` runs multiple instances of ctest in parallel? They are definitely run serially as they should with `ctest -j20` though.

PR #1418, opened 2024-02-15
tests: Add queue creation with hints

jansol:
A basic smoke test that checks that queue creation behaves correctly both with valid and some invalid (combinations of) hints.

These are platform extensions so the test is tagged to all drivers in order to fish out any unexpected breakage from incorrect assumptions, aside from the hardcoded internal limit of 10 property list items. That one should not be a problem *yet* and I'm not sure how deep reaching of a change that would be to fix so that is not addressed yet.


jansol:
Openasip is failing because that runner does not seem to have OpenCL 3.0 headers installed. We could either update the installed headers on the runner to something released after 2020 or make it use our bundled CL headers (not sure how?).

pjaaskel:
I think we use system headers to ensure compatibility with the system ICD loader. But in this case we could just not use ICD and use the PoCL-shipped headers. Right @franz ?

franz:
> But in this case we could just not use ICD and use the PoCL-shipped headers

Yes we could probably get the CI to pass with `-DENABLE_ICD=0`, but that doesn't fix the breakage. The policy for a while has been 1) libpocl is always built with 3.0 internal headers 2) all tests and examples must deal with 1.2 to 3.0 headers (because they could be linked against any ICD). So it should either use ifdefs, or be disabled in CMake if headers < 3.0 (like e.g. `test_program_scope_vars.cpp`).

I suppose we could also completely drop 1.2 support at some point, since we PoCL only supports LLVM versions 14+, and all have good support for 3.0. We might need to fix some queries for 1.2 devices but should be relatively easy.

pjaaskel:
OK, discussed locally: We'll deprecate buillding against 1.2 system headers/ICD in the next release notes and remove it in the next after that. For now, thus @jansol you need to fix the test case so it builds with either.

jansol:
Alright, given that this specifically tests an extension that requires 3.0 I'll disable it from CMake then.

PR #1423, opened 2024-02-20
Lttng tracing

RABijl:
This redoes the LTTng trace parsing in Python. Parses both PoCL and PoCLd. Here is a sample trace that can be opened in chrome tracing: 
[sample.json](https://github.com/pocl/pocl/files/14341325/sample.json)


RABijl:
also feel free to squash the commits

jansol:
Those regexes look a lot like they are matching JSON objects. Could a proper JSON parser perhaps make sense of the input? That would likely be more robust & ergonomic to work with.

pjaaskel:
@jansol probably could, but I think we have plenty of more important tasks to do in the project pipeline, so I suggest we go with this for now and consider switching to a JSON parser if there's a need (problems start to appear) instead of spending much more time on this.

RABijl:
it's matching against strings like these: `{ context_id = 0x5, kernel_id = 0x19, kernel_name = "pocl.dnn.detection.u8" }`, which python ( at least by default) does not parse. A nice thing about regex is that the result contains the expected keys and returns a parsing error. Otherwise i'd have to check if every key exists, or risk the program crashing on a malformed line.

PR #1426, opened 2024-02-21
Vsock

xvim:


pocld/daemon.cc
```diff
@@ -415,6 +417,134 @@ int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports) {
...
+		}
+#endif
+
+		return 0;
+	}
+  return -1;
+}
+
+int PoclDaemon::launchvsock(std::string ListenAddress, struct ServerPorts &Ports) {
+  ListenPorts = {Ports};
```
pjaaskel, line 469: Here also, please check the LLVM C++ naming convention and style.

pjaaskel (changes requested):
Thanks for this useful contribution. On top of fixing my feedback I added, can you add a description of the enhancements (from your commit perhaps) to the next release notes in file https://github.com/pocl/pocl/blob/main/doc/sphinx/source/notes_6_0.rst?

pocld/cmdline.c
```diff
@@ -37,6 +37,7 @@ const char *gengetopt_args_info_help[]
         "  -a, --address=STRING     Listen address",
         "  -p, --port=INT           Listen port",
         "  -v, --log_filter=STRING  Program log category filter",
+        "  -s, --vsock              Whether use VSOCK rather than TCP",
```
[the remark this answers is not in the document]
xvim:
```
static void
init_args_info (struct gengetopt_args_info *args_info)
{

  args_info->help_help = gengetopt_args_info_help[0];
  args_info->version_help = gengetopt_args_info_help[1];
  args_info->address_help = gengetopt_args_info_help[2];
  args_info->port_help = gengetopt_args_info_help[3];
  args_info->log_filter_help = gengetopt_args_info_help[4];
  args_info->use_vsock_help = gengetopt_args_info_help[5];
}
```
put gengetopt_args_info_help in an alphabetical then modify init_args_info?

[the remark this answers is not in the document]
pjaaskel: Actually, it seems to use a generator (check the file header comment) for the .h and .c, thus you should edit cmdline.input and generate these files with `gengetopt --input=cmdline.input --include-getopt`.

pocld/daemon.hh
```diff
@@ -114,6 +125,8 @@ public:
 
   /* returns nullptr on error */
   VirtualContextBase *performSessionSetup(int fd, Request *R);
+  /* whether to use vsock for communication */
+  bool UseVsock;
```
pjaaskel, line 129: Shouldn't this be in the private section?

doc/sphinx/source/notes_6_0.rst
```diff
@@ -44,6 +44,14 @@ server-side. This is a work-in-progress, but is usable for testing
 client apps and libraries that require CG SVM as it seems to work
 often enough.
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Basis for the vsock support
```
pjaaskel, line 48:
```suggestion
Vsock support
```

Is it only a 'basis' or isn't it fully working already?

doc/sphinx/source/notes_6_0.rst
```diff
@@ -44,6 +44,14 @@ server-side. This is a work-in-progress, but is usable for testing
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Adds support for vsock communication to PoCL-Remote, vsock is a
```
pjaaskel, line 51:
```suggestion
Adds support for vsock communication to PoCL-Remote. vsock is a
```

doc/sphinx/source/notes_6_0.rst
```diff
@@ -44,6 +44,14 @@ server-side. This is a work-in-progress, but is usable for testing
...
+high-performance, low-latency, secure, and scalable network communication
+protocol that is ideal for virtualized environments
```
pjaaskel, line 53:
```suggestion
protocol that accelerates guest-host communication in virtualized environments.
```

CMakeLists.txt
```diff
@@ -249,6 +249,8 @@ option(ENABLE_POCL_BUILDING "When OFF, env var POCL_BUILDING has no effect. Defa
 
 option(ENABLE_PRINTF_IMMEDIATE_FLUSH "[currently only applies to CPU drivers] if enabled, printf is flushed immediately when encountered, instead of after all NDRange workgroups are finished" ON)
 
+option(ENABLE_VSOCK "Enable vsock transport in the remote driver" OFF)
```
pjaaskel, line 252: Can you also add a test for the required header/library?

pocld/cmdline.h
```diff
@@ -54,12 +54,18 @@ extern "C"
...
+    char *use_vsock_orig;        /**< @brief vsock flag original value
+                                   given at command line.  */
 
     unsigned int help_given;       /**< @brief Whether help was given.  */
     unsigned int version_given;    /**< @brief Whether version was given.  */
     unsigned int address_given;    /**< @brief Whether address was given.  */
     unsigned int port_given;       /**< @brief Whether port was given.  */
     unsigned int log_filter_given; /**< @brief Whether log_filter was given. */
+    unsigned int
+        use_vsock_given; /**< @brief Whether use VSOCK rather than TCP. */
```
pjaaskel, line 68: This looks strange. Can you move the comment to the previous line instead?

pjaaskel (changes requested):
Getting there...

lib/CL/pocl_networking.c
```diff
@@ -170,3 +171,18 @@ pocl_remote_client_set_socket_options (int socket_fd, int bufsize, int is_fast)
 
   return 0;
 }
+
+int
+pocl_parse_cid (const char *str, unsigned int *cid)
+{
+  char *endptr;
+  unsigned long t = strtoul (str + 6, &endptr, 10);
```
jansol, line 179: This looks rather dangerous to the uninitiated. Please add some comments to document what is going on.

lib/CL/devices/remote/communication.c
```diff
@@ -543,47 +547,76 @@ pocl_network_connect (remote_server_data_t *data, int *fd, unsigned port,
   *fd = -1;
   unsigned addrlen = 0;
 
+#ifdef ENABLE_VSOCK
+  if (strncmp (data->address, "vsock-", 6) == 0)
```
jansol, line 551: Would it be possible / sensible to handle this with `pocl_resolve_address`?

pocld/daemon.cc
```diff
@@ -415,6 +421,137 @@ int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports) {
...
+  pid_t ServerPid = getpid();
+  int Error = 0;
+  NumListenFds = 0;
+  int One = 1;
+
+  struct sockaddr_vm ServerAddrCommand, ServerAddrStream;
+  socklen_t VsockAddrLen;
+  struct sockaddr VsockAddr;
+
+  Error = convertToVsockAddress(ListenAddress, ListenPorts.command, &VsockAddr,
```
jansol, line 488: Same here, would it be possible / sensible to merge this with `pocl_resolve_address` so we can reuse the code path that just takes a `struct sockaddr*` without caring what kind of address it actually is?

pocld/daemon.hh
xvim: vsock can not support TCP_NODELAY, so need a public flag , seems no need to add set/get function

pocld/daemon.cc
```diff
@@ -415,6 +421,137 @@ int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports) {
   return 0;
 }
 
+#ifdef ENABLE_VSOCK
+static int convertToVsockAddress(const std::string &VsockContextId, int Port,
+                                 struct sockaddr *Sockaddr,
+                                 socklen_t *SockaddrLen) {
+  if (Sockaddr == nullptr || SockaddrLen == nullptr)
+    return -1;
```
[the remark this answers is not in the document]
xvim: SockaddrLen  set to  sizeof(struct sockaddr_vm) in convertToVsockAddress

pocld/daemon.cc
xvim: getaddrinfo canot support AF_VSOCK, it seems more sensible to merge this with pocl_resolve_address, I will try fill getaddrinfo manually

pocld/daemon.cc
```diff
@@ -415,6 +421,137 @@ int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports) {
...
+      svm->svm_flags = VMADDR_FLAG_TO_HOST;
+    }
+#endif
+
+    return 0;
+  }
+  return -1;
+}
+
+int PoclDaemon::launchVsock(std::string ListenAddress,
```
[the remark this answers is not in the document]
xvim: About two or three weeks ago, the launch function had the following parameters:
```
int PoclDaemon::launch(struct sockaddr_storage &base_addr,
                      socklen_t base_addrlen,
                      struct ServerPorts &ports)
```
I have a launch implementation that supports vsock, please refer to the following link:

[https://github.com/xvim/pocl/blob/vsock-lagacy/pocld/daemon.cc]

After the launch parameters were changed, it is not so easy to use only one launch. I am not sure if you accept the code similar above?

pocld/daemon.cc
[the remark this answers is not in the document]
jansol: But this function does not allocate Sockaddr, so the size of the allocation must be passed in from the caller and this function must check that the allocation is large enough before writing the actual used size to SockaddrLen. Note that `struct sockaddr` is NOT large enough to hold arbitrary socket addresses. For that you need to allocate at least the size of `struct sockaddr_storage`. This is also how getaddrinfo & related APIs work - `struct sockaddr` is basically treated as a sizeless "virtual" type that is only used to achieve a unified interface without resorting to `char*` or `void*`.

pocld/daemon.cc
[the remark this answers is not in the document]
jansol: Right, I changed that while refactoring so that the daemon could handle DNS names such as "localhost" resolving to multiple different addresses (e.g. IPv4 and IPv6) simultaneously. The way it's meant to work is that it simply takes all results and tries to listen on each produced sockaddr separately. If at least one of the resolved addresses can be successfully listened on that is considered a success.

What I care about is keeping this functionality. I don't have any attachment to the exact function signature of launch so long as this remains possible, hence why I suggested looking into `std::variant` or a tagged union so that it can be passed around as is and resolved to one or more results in the shared resolver.

pocld/daemon.hh
```diff
@@ -73,7 +73,8 @@ public:
    * and finally launches the main I/O thread running
    * `readAllClientSocketsThread()`
    */
-  int launch(std::string ListenAddress, struct ServerPorts &ports);
+  int launch(std::string ListenAddress, struct ServerPorts &ports,
+             const bool &UseVsock = false);
```
pjaaskel, line 77: Why a const ref? Pass by value should be fine?

pjaaskel (changes requested):

pocld/daemon.cc
```diff
@@ -574,8 +619,9 @@ void PoclDaemon::readAllClientSocketsThread() {
             IncompleteRequests.push_back(new Request());
             FdsChanged = true;
             /* XXX: Set these based on CreateOrAttachSession request instead? */
-            pocl_remote_client_set_socket_options(newfd, Params.BufSize,
-                                                  Params.IsFast);
+            if (!UseVsock) /* vsock don't support TCP_NODELAY */
+              pocl_remote_client_set_socket_options(newfd, Params.BufSize,
```
jansol, line 623: Same here, if any of the other socket options make sense for vsock, I'd rather adjust this function to cope with vsock and call it for all socket types.

jansol (approved):
Looks good. I'm not too familiar with socket options on vsock. If possible, I'd like to unify that code path between the socket types, but if it does not make sense then it can stay as it is.

A couple of small things and then this is good to go as far as I'm concerned.

pocld/daemon.cc
```diff
@@ -267,7 +273,8 @@ static std::string find_default_ip_address() {
   return std::string(listen_addr ? listen_addr : "127.0.0.1");
 }
 
-int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports) {
+int PoclDaemon::launch(std::string ListenAddress, struct ServerPorts &Ports,
+                       const bool UseVsock) {
```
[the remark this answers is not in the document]
pjaaskel: For strings and structs in general I can buy it to reduce copying, but for plain old small datatypes that should fit to registers it's difficult to see why.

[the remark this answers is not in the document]
jansol: This boolean here is not a reference though, just const, making the compiler yell at you if you accidentally try to change it within the function which would likely have "fun" consequences. That aside, all of this would probably be better passed as an argument to the `PoclDaemon` constructor (which currently just uses the default constructor).

lib/CL/devices/remote/communication.c
```diff
@@ -570,16 +574,38 @@ pocl_network_connect (remote_server_data_t *data, int *fd, unsigned port,
...
       CL_INVALID_DEVICE, "socket() returned errno: %i\n", errno);
+
+  if (ai->ai_family == AF_VSOCK)
+    {
+      host_freeaddrinfo (ai);
+    }
+  else
+    {
+      freeaddrinfo (ai);
+      err = pocl_remote_client_set_socket_options (socket_fd, bufsize,
```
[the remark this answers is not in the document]
xvim: VSOCK's implementation for level SOL_SOCKET is the same as other sockets. The latest kernel additionally supports SO_ZEROCOPY, and for level == AF_VSOCK, it also supports several:

SO_VM_SOCKETS_BUFFER_SIZE
SO_VM_SOCKETS_BUFFER_MAX_SIZE
SO_VM_SOCKETS_BUFFER_MIN_SIZE
SO_VM_SOCKETS_CONNECT_TIMEOUT_NEW
SO_VM_SOCKETS_CONNECT_TIMEOUT_OLD
We can clearly see this logic in the code here:

https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/net/vmw_vsock/af_vsock.c?id=e27090b1413ff236ca1aec26d6b022149115de2c#n1642

The kernel directory linux/tools/testing/vsock has some examples of calling setsockopt, so my current idea is to add an address family in the current version to distinguish between IPPROTO_TCP and AF_VSOCK, and then further adjust these options in the future, such as:

You think SO_RCVBUF/SO_SNDBUF is not wise in VSOCK
Is it also a problem that NETWORK_BUF_SIZE_FAST is fixed to 4K for systems with a page size of 64k?

pocld/daemon.cc
```diff
@@ -501,7 +546,7 @@ VirtualContextBase *PoclDaemon::performSessionSetup(int fd, Request *R) {
   return ctx;
 }
 
-void PoclDaemon::readAllClientSocketsThread() {
+void PoclDaemon::readAllClientSocketsThread(const bool UseVsock) {
```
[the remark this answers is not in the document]
jansol: Oh nice, not needing the flag at all thanks to using the address family of the client instead makes the latest version much cleaner!

pocld/daemon.cc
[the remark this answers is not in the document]
pjaaskel: Yes, in general I'd agree on the 'const', but my gut says that 99% of args are not typically (meant to be) modified within the function and still they do not mark them const (just clutters the arg list).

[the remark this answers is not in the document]
xvim: so change args to bool UseVsock ?

lib/CL/devices/remote/communication.c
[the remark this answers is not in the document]
jansol: Thanks for taking the time to look into this!
Yes, making the exact options being set conditional on the address family is exactly what I was getting at.

I don't have any concrete knowledge on the SO_RCVBUF/SO_SNDBUF bit, but intuitively I would assume that if AF_VSOCK works by simply sharing memory between processes, it is likely most efficient if each buffer is an exact multiple of the memory page size. Looking at the code it looks like it defaults to 256KiB so maybe it's best to simply leave that entirely untouched for now? We can always adjust it if someone takes the time to benchmark that and finds that other values work better.

SO_ZEROCOPY sounds like it could be useful, since AF_VSOCK essentially just shuffles memory around between processes. https://lwn.net/Articles/913969/ seems to suggest that at least for the "slow" socket (buffer data bulk transfers) it would be beneficial.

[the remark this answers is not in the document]
xvim: The next stage will involve performance research on pocl vsock. We will submit a pull request once the research is completed.

lib/CL/pocl_networking.c
```diff
@@ -124,7 +218,31 @@ pocl_remote_client_set_socket_options (int socket_fd, int bufsize, int is_fast)
...
+                        -1, "setsockopt(SO_RCVTIMEO) returned errno: %i\n",
+                        errno);
+  POCL_RETURN_ERROR_ON ((setsockopt (socket_fd, SOL_SOCKET, SO_SNDTIMEO,
+                                     (const char *)&tv, sizeof (tv))),
+                        -1, "setsockopt(SO_SNDTIMEO) returned errno: %i\n",
+                        errno);
+
+#ifdef ENABLE_VSOCK
+  if (ai_family != AF_VSOCK)
+    {
```
jansol, line 232: It'd be cleaner to only `#ifdef` the if statement here (and its closing brace further down) rather than all the code in it, since the current solution causes unnecessary duplication of the TCP socket options.

Maybe even just do an early return if this is a vsock socket for now?

jansol (changes requested):

lib/CL/pocl_networking.c
```diff
@@ -124,7 +209,32 @@ pocl_remote_client_set_socket_options (int socket_fd, int bufsize, int is_fast)
...
+                                     (const char *)&tv, sizeof (tv))),
+                        -1, "setsockopt(SO_RCVTIMEO) returned errno: %i\n",
+                        errno);
+  POCL_RETURN_ERROR_ON ((setsockopt (socket_fd, SOL_SOCKET, SO_SNDTIMEO,
+                                     (const char *)&tv, sizeof (tv))),
+                        -1, "setsockopt(SO_SNDTIMEO) returned errno: %i\n",
+                        errno);
+
+#ifdef ENABLE_VSOCK
+  if (ai_family != AF_VSOCK)
```
jansol, line 222: This is now skipping the TCP options if the socket IS a TCP one...

jansol (changes requested):
Looks great now, aside from one small fib in the latest change!

pjaaskel (approved):

jansol (approved):

xvim:
sorry push again, still has one defect, addrinfo *ai has been freed, so use server.ss_family replace.

jansol:
The CI failure looks like the one fixed by #1433 so it should disappear when rebasing on latest main.

pjaaskel:
@xvim pls rebase once more

PR #1432, opened 2024-02-22
add new cl_pocl_command_buffer extension APIs

franz:


lib/CL/clCommandReadBufferPOCL.c
```diff
@@ -0,0 +1,64 @@
+/* OpenCL runtime library: clCommandReadBufferPOCL()
+
+   Copyright (c) 2022-2023 Michal Babej / Intel Finland Oy
```
pjaaskel, line 3: Also some other headers have an old year range.

pjaaskel (changes requested):
Thanks for dusting this off. Can you add a mention of these to the release notes?
Also, the *Rect functions do not seem to use the SVM cl_mem shadow buffer refactoring. I wonder do the implicit migrations work with PoCL-R?

franz:
> Also, the *Rect functions do not seem to use the SVM cl_mem shadow buffer refactoring. 

Yes, this is a rebased & slightly cleaned up code from ~1 year ago, before the SVM cl_mem was merged.

> I wonder do the implicit migrations work with PoCL-R?

The ones you have changed should still work, these new ones do not, since they use the SVM pointer directly. But it should work with CPU devices. Also missing are tests; i did limited testing with chipStar graph PR when the code was written, but we should  have more tests (possibly in OpenCL-CTS).

jansol:
#1433 should fix the RDMA CI

PR #1433, opened 2024-02-28
pocld: zero-initialize Requests harder

jansol:
This should fix the recent RDMA CI issues.


pjaaskel:
Can you document this a bit better somewhere (commit msg or the private region in the class?): Why does it affect only RDMA?

jansol:
It technically affects "plain" remote too, that machine just got lucky.

pjaaskel (approved):

PR #1436, opened 2024-03-11
PoCL-R multidevice SVM plus other small improvements (pls check the commits)

pjaaskel:


jansol:
> `sort client.log server.log > combined.log`

The remote test runners could/should perhaps be updated to actually do this and print the combined log at the end? If there is a way to disambiguate between the two server instances at run time (include PID as part of the prefix?) that would make debugging CI errors a lot easier.

pjaaskel:
> > `sort client.log server.log > combined.log`
> 
> The remote test runners could/should perhaps be updated to actually do this and print the combined log at the end? If there is a way to disambiguate between the two server instances at run time (include PID as part of the prefix?) that would make debugging CI errors a lot easier.

For the latter suggestion we'd need to make the log prefix an environment variable or a pocld cmdline arg. Something for later.

PR #1439, opened 2024-03-27
Make memobj locking more aggressive in create_migration_commands

jansol:
With multiple threads submitting commands to multiple devices there seems to be a race condition wrt the last migration event of a mem object. As a simple solution, just keep the object locked until the end of pocl_create_migration_commands.

### DIFF
--- a/lib/CL/pocl_util.c
+++ b/lib/CL/pocl_util.c
@@ -907,8 +907,6 @@ FINISH_VER_SETUP:
         }
     }
 
-  POCL_UNLOCK_OBJ (mem);
-
   /*****************************************************************/
 
   /* enqueue a command for export.
@@ -1008,14 +1006,13 @@ FINISH_VER_SETUP:
        * which only read from buffers */
       if (readonly)
         {
-          POCL_LOCK_OBJ (mem);
           mem->last_event = last_migration_event;
-          POCL_UNLOCK_OBJ (mem);
           POname (clReleaseEvent) (final_event);
         }
       else /* because explicit event */
         POname (clReleaseEvent) (last_migration_event);
     }
+  POCL_UNLOCK_OBJ (mem);
 
   return CL_SUCCESS;
 }


jansol:
I'm pretty sure the failures are unrelated to this fix. Seems like something is wrong with the CMakeLists.txt for pulling in the examples' sources? And on the other runners I guess they are cached / not cleaned up between runs so it looks fine there.

jansol:
Aside from the CMake problems in tier1 and cts tests, there's a bunch of spirv-related failures I can't do much about at the moment: ubuntu appears to have updated llvm-spirv-18 to LLVM 18.1, which bumps the so version, but the llvm-18 packages still predate that so llvm-spirv fails to find the llvm libraries it needs, causing SPIR-V tests to fail on LLVM 18.

Remote is failing the SVM test and there are some messages about a potential double free. I'm not familiar with that code but I'm fairly certain it's unrelated to this PR.

franz:
I think most of the CMake problems are caused by missing `mkdir -p /home/github/examples/source`.

However, to run `tier1` configuration on `rancher`, it'll also need the sources for the `tier1` testsuites present in `/home/github/examples/source`.

jansol:
Isn't CMake supposed to create those folders & download the necessary sources?

pjaaskel (approved):
Looks good (after the tests pass). Somewhat more fine-grained locking than in my workaround of my local tree.

franz:
> Isn't CMake supposed to create those folders & download the necessary sources?

The top directory is expected to be found, it isn't created (perhaps it should be). For the testsuite subdirectories, it depends on the testsuite. For some, the sources exist only as compressed package, and are hard/impossible to find on the internet anymore.

franz:
The last commit (`Try some more possibilities for llvm-spirv`) might be useful, but in case we find unversioned `llvm-spirv`, we need to actually *verify* its version matches the LLVM major version, by parsing `llvm-spirv --version`.

In my experience, the amount of churn in both LLVM and the translator is such that even a difference of one major version has very high probability of breaking things. Since LLVM 15 there have been opaque-type and opaque-pointer changes which are 100% incompatible, and then compatibility can break even on "trivial" things like debuginfo as seen [here.](https://github.com/CHIP-SPV/chipStar/issues/456)

jansol:
Added check for llvm-spirv version.

And as a drive-by fix updated our PyOpenCL scripts since they were failing the tier1 build on some CI machines. (the virtualenv was missing some packages, IIRC they used to be present by default but no longer are)

jansol:
@franz I believe the CI box that is failing because of the PyOpenCL update is one maintained by you. (it refuses to update the git checkout since there are local changes i.e. the now-obsolete patch)

PR #1441, opened 2024-04-03
cl_ext_buffer_device_address

pjaaskel:
For implementation also in RustiCL and usage from chipStar.

The actual extension to write and propose after these happen.

- [ ] change log entry

### DIFF
--- a/include/CL/cl_ext_pocl.h
+++ b/include/CL/cl_ext_pocl.h
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2021 Tampere University
- *               2023 Pekka Jääskeläinen / Intel Finland Oy
+ *               2023-2024 Pekka Jääskeläinen / Intel Finland Oy
  *
  * PoCL-specific proof-of-concept (draft) or finalized OpenCL extensions.
  *
@@ -60,42 +60,89 @@ typedef CL_API_ENTRY cl_int
 
 #endif
 
-/* cl_pocl_pinned_buffers (experimental stage) */
+/* cl_ext_buffer_device_address (experimental stage)
 
-#ifndef cl_pocl_pinned_buffers
-#define cl_pocl_pinned_buffers 1
-#define CL_POCL_PINNED_BUFFERS_EXTENSION_NAME "cl_exp_pinned_buffers"
-
-/* TODO: We need also platform/runtime extension due to the new buffer
-   creation flags. */
-
-/* clCreateBuffer(): A new cl_mem_flag CL_MEM_PINNED:
-
-   This flag specifies that the buffer must be persistently allocated
-   in the device's physical memory for its lifetime. That is, the buffer's
-   device address will remain the same and the space is reserved until
-   the buffer is freed. The device-specific buffer content updates are
-   still performed by implicit or explicit buffer migrations performed by
-   the runtime or the client code. If any of the devices in the context
-   does not support pinning, an error (TO DEFINE) is returned.
+   TODO:
+   * Perhaps a better name could be simply 'cl_ext_device_pointers'?
 */
-#define CL_MEM_PINNED (1ul << 31)
 
-/* clGetMemObjectInfo(): A new query CL_MEM_DEVICE_PTR:
+#ifndef cl_ext_buffer_device_address
+#define cl_ext_buffer_device_address 1
 
-Returns a list of pinned device addresses for a buffer allocated
-with CL_MEM_PINNED. If the buffer was not created with CL_MEM_PINNED,
-returns CL_INVALID_MEM_OBJECT.
+/* We need both a platform and device extension due to the new buffer
+   creation flag. */
+
+/* clCreateBuffer(): A new cl_mem_flag CL_MEM_DEVICE_ADDRESS_EXT:
+
+   This flag specifies that the buffer must have a single fixed address
+   for its lifetime, which should be unique at least across the devices
+   of the context, but not necessarily with the host (virtual memory) as
+   with SVM or USM.
+
+   The flag might imply that the buffer will be "pinned" permanently to
+   a device's memory, but might not be necessarily so, as long as the address
+   range of the buffer remains constant.
+
+   The address is guaranteed to remain the same until the buffer is freed and
+   the address can be queried via clGetMemObjectInfo().
+
+   The device-specific buffer content updates must be still performed by
+   implicit or explicit buffer migrations performed by the runtime or the
+   client code. If any of the devices in the context does not support
+   such allocations, an error (CL_INVALID_VALUE) is returned.
+
 */
-#define CL_MEM_DEVICE_PTRS 0xff01
+#define CL_MEM_DEVICE_ADDRESS_EXT (1ul << 31)
 
-typedef struct _cl_mem_pinning
-{
-  cl_device_id device;
-  void *address;
-} cl_mem_pinning;
+/* Experimental multi-device allocation support flags. */
 
-/* cl_pocl_pinned_buffers */
+/* If combined with CL_MEM_DEVICE_ADDRESS_EXT, each device can have their
+   own (fixed) device-side address and copy of the created buffer which are
+   synchronized implicitly. The main difference to a default cl_mem allocation
+   is then that the addresses are queriable with CL_MEM_DEVICE_PTRS_EXT and
+   the per-device address is guaranteed to be the same for the entire-lifetime
+   of the cl_mem.
+
+   Otherwise, each compatible device in the context can access a shared
+   allocation in one of the device's global memory using the same address. */
+#define CL_MEM_DEVICE_PRIVATE_EXT (1ul << 30)
+
+/* clGetMemObjectInfo(): A new cl_mem_info type CL_MEM_DEVICE_PTR_EXT:
+
+   Returns the device address (as a void*) for a buffer allocated with
+   CL_MEM_DEVICE_ADDRESS_EXT. If the buffer was not created with the flag,
+   returns CL_INVALID_MEM_OBJECT.
+*/
+#define CL_MEM_DEVICE_PTR_EXT 0xff01
+
+typedef void *cl_mem_device_address_EXT;
+
+/* Returns the device-address pairs for a buffer allocated with
+   CL_MEM_DEVICE_ADDRESS_EXT | CL_MEM_DEVICE_PRIVATE_EXT.
+*/
+#define CL_MEM_DEVICE_PTRS_EXT 0xff02
+
+  typedef struct _cl_mem_device_address_pair_EXT
+  {
+    cl_device_id device;
+    cl_mem_device_address_EXT *address;
+  } cl_mem_device_address_pair_EXT;
+
+/* clSetKernelExecInfo(): CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT:
+
+   Similar to CL_KERNEL_EXEC_INFO_SVM_PTRS except for CL_MEM_DEVICE_ADDRESS_EXT
+   device pointers: If a device pointer accessed by a kernel is not passed as
+   an argument, it must be set by this property.
+*/
+#define CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT 0x11B8
+
+  /* A new function clSetKernelArgDevicePointerEXT() for setting raw device
+     pointers as kernel arguments. */
+
+  typedef cl_int (CL_API_CALL *clSetKernelArgDevicePointerEXT_fn) (
+      cl_kernel kernel, cl_uint arg_index, const void *arg_value);
+
+/* cl_ext_buffer_device_address (experimental stage) */
 #endif
 
 /***********************************
@@ -497,6 +544,8 @@ clEnqueueSVMMemFillRectPOCL (cl_command_queue  command_queue,
                              const cl_event *  event_wait_list,
                              cl_event *        event);
 
+extern CL_API_ENTRY cl_int CL_API_CALL clSetKernelArgDevicePointerEXT (
+    cl_kernel kernel, cl_uint arg_index, const void *arg_value);
 
 #endif // CL_NO_PROTOTYPES
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1316,7 +1316,7 @@ set(HOST_DEVICE_EXTENSIONS "cl_khr_byte_addressable_store cl_khr_global_int32_ba
 cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics \
 cl_khr_local_int32_extended_atomics cl_khr_3d_image_writes \
 cl_khr_command_buffer cl_khr_subgroups cl_intel_unified_shared_memory \
-cl_exp_pinned_buffers")
+cl_ext_buffer_device_address")
 
 # Host CPU device: list of OpenCL 3.0 features that are always enabled
 # TODO: __opencl_c_atomic_scope_all_devices works with CPU device but not others
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -198,9 +198,11 @@
 
 #define LLVM_OPT "@LLVM_OPT@"
 
+#define LLVM_LINK "@LLVM_LINK@"
+
 #cmakedefine LLVM_MAJOR @LLVM_VERSION_MAJOR@
 
-// minimum LLVM version at which to enable new PM
+/* minimum LLVM version at which to enable new PM */
 #define MIN_LLVM_NEW_PASSMANAGER 14
 
 #cmakedefine LLVM_BUILD_MODE_DEBUG
--- a/doc/sphinx/source/notes_6_0.rst
+++ b/doc/sphinx/source/notes_6_0.rst
@@ -20,6 +20,33 @@ any particular behavior (or lack thereof) they are treated as a no-op. However
 specifying them no longer causes `clCreateCommandQueueWithProperties` to return
 an error.
 
+===================================================
+Experimental cl_ext_buffer_device_address prototype
+===================================================
+
+This new extension prototype enables allocating `cl_mem` buffers with client-accessible
+physical addresses which is guaranteed to be fixed for the lifetime of the buffer.
+The main difference to coarse-grain SVM allocations is that all
+SVM allocations require always the virtual address address to match the device address,
+thus mapping the buffer address range also to the vmem even though its contents
+are managed only via explicit memcopies by the application.
+
+Although it's a very simple incremental extension to the basic `clCreateBuffer()` API,
+it enables implementing `hipMalloc()` HIP/CUDA and `omp_target_alloc()` OpenMP
+allocation calls when the application doesn't require a unified address space.
+
+There is also a prototype implementation of the extension in `Rusticl/Mesa <https://gitlab.freedesktop.org/karolherbst/mesa/-/commit/fa5f51da728dcaf277b0919e90e0400859f290bb>`_.
+
+`chipStar <https://github.com/CHIP-SPV/chipStar>`_ can optionally
+use the extension, if neither Unified Shared Memory (Intel extension) nor
+OpenCL 2.0+ Coarse-Grain SVM is supported by the OpenCL device/platform,
+and the HIP/CUDA application doesn't require unified address space, but
+explicitly specifies the memory copy directions.
+
+The actual extension text is yet to write and the extension can
+change without notification as we get feedback and more experience from
+using it.
+
 ===========================
 Driver-specific features
 ===========================
--- a/lib/CL/CMakeLists.txt
+++ b/lib/CL/CMakeLists.txt
@@ -194,6 +194,7 @@ set(POCL_LIB_SOURCES  "clCreateContextFromType.c"
                    "clCommandWriteImagePOCL.c"
                    "clEnqueueSVMMemcpyRectPOCL.c"
                    "clEnqueueSVMMemfillRectPOCL.c"
+                   "clSetKernelArgDevicePointer.c"
 )
 
 if(ANDROID)
--- a/lib/CL/clCreateBuffer.c
+++ b/lib/CL/clCreateBuffer.c
@@ -1,25 +1,26 @@
 /* OpenCL runtime library: clCreateBuffer()
 
    Copyright (c) 2011-2013 Universidad Rey Juan Carlos
-                           Pekka Jääskeläinen / Tampere University of Technology
-   
+                           Pekka Jääskeläinen / Tampere University of Tech.
+                 2024 Pekka Jääskeläinen / Intel Finland Oy
+
    Permission is hereby granted, free of charge, to any person obtaining a copy
-   of this software and associated documentation files (the "Software"), to deal
-   in the Software without restriction, including without limitation the rights
-   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-   copies of the Software, and to permit persons to whom the Software is
+   of this software and associated documentation files (the "Software"), to
+   deal in the Software without restriction, including without limitation the
+   rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+   sell copies of the Software, and to permit persons to whom the Software is
    furnished to do so, subject to the following conditions:
-   
+
    The above copyright notice and this permission notice shall be included in
    all copies or substantial portions of the Software.
-   
+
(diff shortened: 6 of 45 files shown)


include/CL/cl_ext_pocl.h
```diff
@@ -60,42 +60,42 @@ typedef CL_API_ENTRY cl_int
...
-   device address will remain the same and the space is reserved until
-   the buffer is freed. The device-specific buffer content updates are
-   still performed by implicit or explicit buffer migrations performed by
-   the runtime or the client code. If any of the devices in the context
-   does not support pinning, an error (TO DEFINE) is returned.
+   This flag specifies that the buffer must have a single fixed address
+   for its lifetime, which should be unique at least across the devices
+   of the context, but not necessarily with the host (virtual memory) as
+   with SVM or USM. The flag might imply that the buffer is "pinned"
+   permanently to a device's memory, but might not be necessarily so,
```
linehill, line 77: Will indirect buffer accesses from kernels work in case the buffers happen not to be pinned to devices’ memory? Maybe this extension needs a new clSetKernelExecInfo() parameter, `CL_KERNEL_EXEC_INFO_<something>_PTRS`, to let drivers know about indirectly accessed buffers?

pjaaskel: Yes, that might be the most convenient and perf. portable way. The original semantics of this extension included "pinned" which as a side effect ensured consistency also for indirect accesses, but I wanted to relax that for additional implementation flexibility.

Another question is multi-device random access. For that we can add a new property for clCreateBufferWithProperties() later. So the initial prototype would be defined to work for single-device allocations only.

karolherbst:
I think there is merit in having to states of operations here. Implementing this for a single device is pretty trivial, but still has value as you can rely on an address pointing to the same memory across invocations and that address to remain forever valid. The other mode is Userspace being in control of placing memory within the GPUs vm, which is more or less the requirement to support this extension across multiple devices.
So a driver could run into the situation where each device can support this extension within a single device context, but maybe not in a multi device one.

pjaaskel:
> I think there is merit in having to states of operations here. Implementing this for a single device is pretty trivial, but still has value as you can rely on an address pointing to the same memory across invocations and that address to remain forever valid. The other mode is Userspace being in control of placing memory within the GPUs vm, which is more or less the requirement to support this extension across multiple devices. So a driver could run into the situation where each device can support this extension within a single device context, but maybe not in a multi device one for a certain combination of devices.

Makes sense. Initially we can limit this to single-device contexts (since clCreateBuffer() by default allocates for all devices in the context) and can then expand with two properties for clCreateBufferWithProperties() along the lines of 1) **CL_MEM_TARGET_DEVICE**: a convenience feature for creating single device allocations even if there are multiple devices in the context, making the buffer accessible only from kernels launched on the device defined with the property. 2) **CL_MEM_MULTI_DEVICE_ACCESS**: a list of target devices from which the buffer must be randomly accessible. clCreateBuffer() can then default to 2) if it's called with a context with more than one device and to 1) if with only one. How does this sound?

karolherbst:
> > I think there is merit in having to states of operations here. Implementing this for a single device is pretty trivial, but still has value as you can rely on an address pointing to the same memory across invocations and that address to remain forever valid. The other mode is Userspace being in control of placing memory within the GPUs vm, which is more or less the requirement to support this extension across multiple devices. So a driver could run into the situation where each device can support this extension within a single device context, but maybe not in a multi device one for a certain combination of devices.
> 
> Makes sense. Initially we can limit this to single-device contexts (since clCreateBuffer() by default allocates for all devices in the context) and can then expand with two properties for clCreateBufferWithProperties() along the lines of 1) **CL_MEM_TARGET_DEVICE**: a convenience feature for creating single device allocations even if there are multiple devices in the context, making the buffer accessible only from kernels launched on the device defined with the property. 2) **CL_MEM_MULTI_DEVICE_ACCESS**: a list of target devices from which the buffer must be randomly accessible. clCreateBuffer() can then default to 2) if it's called with a context with more than one device and to 1) if with only one. How does this sound?

Maybe it makes sense to add a device property to query if a device supports participating in a multi device buffer address thing.

pjaaskel:
@karolherbst Yep, a device/platform-query will be needed at that point.

I updated the "draft spec" in the header comment, the example CPU implementation and the test case. It now has the indirect ptr support. I also noticed a couple of things that have to be updated in the spec related to passing pointers inside the buffers to kernels using sub-buffers which I reflected in the notes.  I believe now this extension should have the minimal features for a minimal hipMalloc() implementation.

include/CL/cl_ext_pocl.h
```diff
@@ -60,42 +60,73 @@ typedef CL_API_ENTRY cl_int
...
+
+   Sub-buffers created from buffers allocated with the flag are guaranteed
+   to be synchronized fully as a single entity, regardless of the offset value
+   passed. That is, when the parent buffer has been allocated with
+   CL_MEM_DEVICE_ADDRESS, setting the size of the sub-buffer to 0 does not
+   return CL_INVALID_BUFFER_SIZE, but denotes that the sub-buffer's
+   parent buffer's contents might be accessed anywhere by the kernel.
+   This is to ensure pointer arithmetics can be done freely within the parent
+   allocation on the kernel side although the kernel would get a pointer to
+   a middle passed in as an argument.
```
karolherbst, line 100: I don't think this is such a good idea, because that's really isn't a sub-buffer anymore.

Maybe adding `clSetKernelArgDeviceAddressPointer` (similar to `clSetKernelArgSVMPointer`) would allow for a better programming model instead here, as you only need to keep track of your pointers and the runtime can map that to the allocations internally as this tracking needs to happen for `CL_KERNEL_EXEC_INFO_DEVICE_PTRS` already anyway.

This somewhat makes it a new SVM variant, but that doesn't have to be a bad thing if the API gets duplicated. Maybe it can be merged, but for that I'd wait for WG review and what people there decide.

pjaaskel: Yes, creating a new `clSetKernelArgDeviceAddressPointer` is the cleaner alternative, but since it's much more intrusive spec-vise (adds a new function to the ICD dispatch table, etc.), I'd go with the sub-buffer approach in the draft version and perhaps introduce the new function in the final version of the extension. Is the sub-buffer approach a lot of work to implement in RustiCL/Mesa?

karolherbst: Also as long as it stays an extension you'd go through `clGetExtensionFunctionAddress` anyway, so as long as it's not considered to become a core spec it doesn't really matter all that much. But maybe you could also allow `clSetKernelArgSVMPointer` to be used, because it's really fundamentally the same. Could also be renamed to a more generic name via an union in a later CL spec version.

I doubt the amount of work changes that much, it just feels too hacky for my test and makes device-address-buffer sub-buffers objects you'd need to special case in a couple of places. Where if you design it like `clSetKernelArgSVMPointer` runtimes could just share a bunch of code, because it's essentially the same for the runtime, just instead of tracking pointer -> SVM alloc, you'd track pointer -> cl_mem internally or some base + sub classes.

It also allows application to set arbitrary pointers as arguments which would remove the need to constantly allocate new sub-buffers as you can just point to a random pointer directly.

Though maybe there is also merit in creating sub-buffers and using some of the memory Copy and Map APIs on them.

So maybe it's fine to simply bring it up in WG discussions and see how others feel about this.

pjaaskel: Yes, you are right. I didn't remember that the `clGetExtensionFunctionAddress` indirection helps here. I'll add the new setarg API, makes it much cleaner. It think has to be a new API as we must be able to differentiate between dev ptrs and vm ptrs in case the device supports both?

My current plan with this extension is to prototype it as an 'ext' extension in PoCL (CPU obviously works already, I'm working on L0 implementation right now) and (if I understood your intentions right) in RustiCL, with chipStar as the first example client app. This way we support a lot of CUDA/HIP apps already using only the simple extension. Then if it seems more widely useful (e.g. a lot of OpenMP offload apps should work with it alone) with positive feedback, bring it to OpenCL WG for potential 'khr' promotion discussions.

karolherbst: > It think has to be a new API as we must be able to differentiate between dev ptrs and vm ptrs in case the device supports both?

Depends on how runtimes are willing to figure this out. Those allocations can't overlap so you can easily tell them apart, but of course there is runtime overhead you wouldn't have when having two APIs.

> My current plan with this extension is to prototype it as an 'ext' extension in PoCL (CPU obviously works already, I'm working on L0 implementation right now) and (if I understood your intentions right) in RustiCL, with chipStar as the first example client app. This way we support a lot of CUDA/HIP apps already using only the simple extension. Then if it seems more widely useful (e.g. a lot of OpenMP offload apps should work with it alone) with positive feedback, bring it to OpenCL WG for potential 'khr' promotion discussions.

sounds like a plan

tests/runtime/test_pinned_buffers.cpp
```diff
@@ -34,21 +34,42 @@
...
 )raw";
 
-void *getDeviceAddressFromHost(cl::Buffer &Buf) {
-  cl_mem_pinning Pinning;
+// A kernel that accesses another buffer indirectly.
+static char IndirectAccess[] = R"raw(
+
+  __kernel void indirect_access (__global long* in_addr,
+                                 __global int* out) {
+    *out = **(int**)in_addr;
```
karolherbst, line 53: For runtimes not supporting generic pointers this needs to be `*out = **(int __global* __global*)in_addr;` or whatever style you prefer to specify the address space (I just put them right before the `*` because that feels more natural to me).

pjaaskel:
The latest update addressess @karolherbst's points and implements the extension for the Level Zero backend as well. I think the latest update address your points. @linehill note. @franz this also fixes USM and chipStar for Level Zero and includes the two commits from you.

karolherbst:
Rusticl implementation: https://gitlab.freedesktop.org/karolherbst/mesa/-/commits/pinned_memory/?ref_type=heads

I'm not entirely happy how I'm doing things there, especially as cross device memory is kinda broken in rusticl atm, but at least for single device context it should work as expected.

I have tested that on my Apple M2 and on my x86_64 Intel laptop. Haven't tested if it works as expected with AMD, but I don't see a good reason why it shouldn't. If there are bugs feel free to report it to me, however I'll be at IWOCL this week and won't be able to look into any bugs I can't hit on my MacBook.

pjaaskel:
@karolherbst nice - thanks! @linehill ^ let's test chipStar on AMD/NVIDIA GPUs on that Rusticl branch.

pjaaskel:
Some updates:
 * Renamed `CL_KERNEL_EXEC_INFO_DEVICE_PTRS` to `CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT` 
 * I was reminded gently by a CI bot that PoCL-R multidevice SVM implementation relies on per-device internal allocations with fixed addressess, so I added a first multi-device mode flag `CL_MEM_DEVICE_PRIVATE_EXT` which allows each device to create their own mirror buffers, like default cl_mem, as long as the addresses are fixed.
 * Test case fix (passed value instead of a pointer to it).
 * clSetKernelArgInfo() fix for CPU.

pjaaskel:
Passed all but the ARM CIs which are down. Change log: https://github.com/pocl/pocl/pull/1441/commits/0813788983652bd41a866a19be5200ebb308dcd3

pjaaskel:
Hmh. Github UI got stuck. Says "merge in progress". The main seems to be updated so closing.

linehill:
@karolherbst, I tried to run POCL's [test_device_address](https://github.com/pocl/pocl/blob/main/tests/runtime/test_device_address.cpp)(*) on rusticl on llvmpipe and radeonsi and both of them segfaults [here](https://github.com/pocl/pocl/blob/4aaa03176d05b9a4105051f4e0722057820bc7d8/tests/runtime/test_device_address.cpp#L156).

*: With minor local modifications to make it compile without POCL specific stuff.

Backtrace from llvmpipe:
```
(gdb) run
Starting program: /mnt/md0/linehill/ws-mesa/test_device_address 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
bt
[New Thread 0x7fffe4651640 (LWP 40668)]
[New Thread 0x7fffdbfff640 (LWP 40669)]
[New Thread 0x7fffe3e50640 (LWP 40670)]
[New Thread 0x7fffe364f640 (LWP 40671)]
[New Thread 0x7fffe2e4e640 (LWP 40672)]
[New Thread 0x7fffe264d640 (LWP 40673)]
[New Thread 0x7fffe1e4c640 (LWP 40674)]
[New Thread 0x7fffe164b640 (LWP 40675)]
[New Thread 0x7fffe0b29640 (LWP 40676)]
llvmpipe (LLVM 17.0.6, 256 bits) OpenCL 3.0 : suitable
[New Thread 0x7fffe02e6640 (LWP 40677)]
[New Thread 0x7fffdb6bd640 (LWP 40678)]

Thread 1 "test_device_add" received signal SIGSEGV, Segmentation fault.
0x00007ffff480f091 in llvmpipe_set_global_binding (pipe=0x5555556335d0, first=0, count=1, resources=0x55555a0222e0, handles=0x7fffffffce58) at ../../mesa/src/gallium/drivers/llvmpipe/lp_state_cs.c:1764
1764	   if (first + count > cs->max_global_buffers) {
(gdb) bt
#0  0x00007ffff480f091 in llvmpipe_set_global_binding (pipe=0x5555556335d0, 
    first=0, count=1, resources=0x55555a0222e0, handles=0x7fffffffce58)
    at ../../mesa/src/gallium/drivers/llvmpipe/lp_state_cs.c:1764
#1  0x00007ffff421e6c3 in mesa_rust::pipe::context::PipeContext::set_global_binding (self=0x555556de8778, res=..., out=...)
    at ../../mesa/src/gallium/frontends/rusticl/mesa/pipe/context.rs:449
#2  0x00007ffff41e2fe0 in rusticl::core::device::{impl#1}::bda (
    self=0x7fffffffd040, res=0x5555585df158)
    at ../../mesa/src/gallium/frontends/rusticl/core/device.rs:150
#3  0x00007ffff41c4aaa in rusticl::core::memory::MemBase::new_buffer (
    context=..., flags=2147483681, size=64, host_ptr=0x7fffffffd9a0, props=...)
    at ../../mesa/src/gallium/frontends/rusticl/core/memory.rs:428
#4  0x00007ffff41f4741 in rusticl::api::memory::create_buffer_with_properties (
    context=0x5555556c9150, properties=0x0, flags=2147483681, size=64, 
    host_ptr=0x7fffffffd9a0)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:299
#5  0x00007ffff41f49f7 in rusticl::api::memory::create_buffer (
    context=0x5555556c9150, flags=2147483681, size=64, host_ptr=0x7fffffffd9a0)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:309
#6  0x00007ffff41f48ae in rusticl::api::memory::cl_create_buffer (
    context=0x5555556c9150, flags=2147483681, size=64, 
    host_ptr=0x7fffffffd9a0, errcode_ret=0x7fffffffd764)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:302
#7  0x0000555555559250 in cl::Buffer::Buffer (this=0x7fffffffd868, 
    context=..., flags=<optimized out>, size=<optimized out>, 
    host_ptr=<optimized out>, err=0x0) at /usr/include/CL/opencl.hpp:3916
#8  0x00005555555579f5 in main () at test_device_address.cpp:179
```
Backtrace from radeonsi:
```
(gdb) run
Starting program: /mnt/md0/linehill/ws-mesa/test_device_address 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
bt
[New Thread 0x7fffe4651640 (LWP 40178)]
[New Thread 0x7fffe3b2f640 (LWP 40179)]
[New Thread 0x7fffe332e640 (LWP 40180)]
[New Thread 0x7fffe2b2d640 (LWP 40181)]
[New Thread 0x7fffe1e52640 (LWP 40182)]
[New Thread 0x7fffe1590640 (LWP 40183)]
[New Thread 0x7fffe0d8f640 (LWP 40184)]
AMD Radeon Pro VII (radeonsi, vega20, LLVM 17.0.6, DRM 3.56, 5.15.0-67-generic) OpenCL 3.0 : suitable
[New Thread 0x7fffc3fff640 (LWP 40185)]
[New Thread 0x7fffe058e640 (LWP 40186)]
[New Thread 0x7fffc37fe640 (LWP 40187)]

Thread 1 "test_device_add" received signal SIGSEGV, Segmentation fault.
0x00007ffff46c0eac in si_set_global_binding (ctx=0x555555752c10, first=0, n=1, resources=0x555556c026b0, handles=0x7fffffffce58) at ../../mesa/src/gallium/drivers/radeonsi/si_compute.c:351
351	   if (first + n > program->max_global_buffers) {
(gdb) bt
#0  0x00007ffff46c0eac in si_set_global_binding (ctx=0x555555752c10, first=0, 
    n=1, resources=0x555556c026b0, handles=0x7fffffffce58)
    at ../../mesa/src/gallium/drivers/radeonsi/si_compute.c:351
#1  0x00007ffff421e6c3 in mesa_rust::pipe::context::PipeContext::set_global_binding (self=0x555555646cb8, res=..., out=...)
    at ../../mesa/src/gallium/frontends/rusticl/mesa/pipe/context.rs:449
#2  0x00007ffff41e2fe0 in rusticl::core::device::{impl#1}::bda (
    self=0x7fffffffd040, res=0x555556c1aec8)
    at ../../mesa/src/gallium/frontends/rusticl/core/device.rs:150
#3  0x00007ffff41c4aaa in rusticl::core::memory::MemBase::new_buffer (
    context=..., flags=2147483681, size=64, host_ptr=0x7fffffffd9a0, props=...)
    at ../../mesa/src/gallium/frontends/rusticl/core/memory.rs:428
#4  0x00007ffff41f4741 in rusticl::api::memory::create_buffer_with_properties (
    context=0x555555571f20, properties=0x0, flags=2147483681, size=64, 
    host_ptr=0x7fffffffd9a0)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:299
#5  0x00007ffff41f49f7 in rusticl::api::memory::create_buffer (
    context=0x555555571f20, flags=2147483681, size=64, host_ptr=0x7fffffffd9a0)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:309
#6  0x00007ffff41f48ae in rusticl::api::memory::cl_create_buffer (
    context=0x555555571f20, flags=2147483681, size=64, 
    host_ptr=0x7fffffffd9a0, errcode_ret=0x7fffffffd764)
    at ../../mesa/src/gallium/frontends/rusticl/api/memory.rs:302
#7  0x0000555555559250 in cl::Buffer::Buffer (this=0x7fffffffd868, 
    context=..., flags=<optimized out>, size=<optimized out>, 
    host_ptr=<optimized out>, err=0x0) at /usr/include/CL/opencl.hpp:3916
#8  0x00005555555579f5 in main () at test_device_address.cpp:179
```

karolherbst:
@linehill seems like they both kinda have the same code here and rely on a "bound program" to give a pointer to a memory object.

It's one of the part why I don't like the current code, but I wasn't aware how those drivers implement those bits. It's a very old interface and needs to be reworked anyway, so maybe I'll be focusing on that to implement this extension once I'm back fro IWOCL.

pjaaskel:
@karolherbst https://github.com/CHIP-SPV/chipStar/pull/830 is a PR for chipStar to use that extension when USM/SVM is not available, if you want to run more test cases.

include/CL/cl_ext_pocl.h
```diff
@@ -60,42 +60,89 @@ typedef CL_API_ENTRY cl_int
...
+/* clGetMemObjectInfo(): A new cl_mem_info type CL_MEM_DEVICE_PTR_EXT:
 
-   This flag specifies that the buffer must be persistently allocated
-   in the device's physical memory for its lifetime. That is, the buffer's
-   device address will remain the same and the space is reserved until
-   the buffer is freed. The device-specific buffer content updates are
-   still performed by implicit or explicit buffer migrations performed by
-   the runtime or the client code. If any of the devices in the context
-   does not support pinning, an error (TO DEFINE) is returned.
+   Returns the device address (as a void*) for a buffer allocated with
```
karolherbst, line 112: I think returning `void*` is invalid here. Or at least needs some more comments. GPUs with a 64 bit VM can be used in 32 bit applications and we either require those pinned buffers to be allocated in the lower 32 bits or we relax that and make it a `cl_ulong`.

pjaaskel: Yes, it crossed my mind, but then I started to doubt if that's a realistic configuration these days or something worth trying to support. Is it? I'm not sure. Host void* allows 32b/32b (cl_ulong would force 32b systems to use 64b here), and 64b/32b systems to function, but the combination of 32b/64b host/device is indeed a corner case.

karolherbst: It's hit by any 32 bit application on a modern system. However I doubt that applications still being 32 bit only would start using this extension.

But anyway, I think this would require a notice on the ext like "This requires a 64 bit host to function" or something like that. Or it simply moves to use `cl_ulong` as supporting this use case is generally pretty cheap. But I think `void*` should be changed to some int type anyway, so that applications don't assume it's a host accessible pointer or something.

In either case, I think it needs to be mentioned, so people don't ask later.

pjaaskel: I agree. Anyhow I think 32b/32b (where cl_ulong would be unnecessary) can be a valid ("deeply embedded") case. I'll go with a documentation fix for now and we can discuss more when we propose this to Khronos docs.

include/CL/cl_ext_pocl.h
```diff
@@ -60,42 +60,89 @@ typedef CL_API_ENTRY cl_int
...
 
-typedef struct _cl_mem_pinning
-{
-  cl_device_id device;
-  void *address;
-} cl_mem_pinning;
+  typedef struct _cl_mem_device_address_pair_EXT
+  {
+    cl_device_id device;
+    cl_mem_device_address_EXT *address;
```
karolherbst, line 128: this needs to be `cl_mem_device_address_EXT address;` otherwise this ends up as a pointer to a pointer or `void **address` but I think once this is a int type this shouldn't matter anymore. Just something I hit while implementing it again.

pjaaskel:
Pls check #1449. I ended up changing the address datatype too. @linehill heads-up for chipStar client code.

karolherbst:
properly implemented the extension: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/28785

verified to work on radeonsi:
```
$ RUSTICL_ENABLE=radeonsi run_local_mesa ./tests/runtime/test_device_address 
AMD Radeon RX 6700 XT (radeonsi, navi22, LLVM 17.0.6, DRM 3.57, 6.8.0-rc5+) OpenCL 3.0 : suitable
OK
```

pjaaskel:
Excellent! @linehill time to test with chipStar again?

karolherbst:
I'm sure missing global program var support might still be an issue... I should create a mesa branch with all those things put together at some point again.

pjaaskel:
OK, lack of global support (esp. if it still supports the majority of the cases) shouldn't affect most of the programs. Abort/assert is currently implemented via global variable signaling in chipStar, but if compiled without assertions we should be able run a lot of apps already.

linehill:
FYI, updated chipStar (https://github.com/CHIP-SPV/chipStar/pull/830). 36 out of 87 HeCBench benchmarks successfully ran on Radeon PRO VIII (vega20/gfx906) through rusticl.

karolherbst:
> FYI, updated chipStar ([CHIP-SPV/chipStar#830](https://github.com/CHIP-SPV/chipStar/pull/830)). 36 out of 87 HeCBench benchmarks successfully ran on Radeon PRO VIII (vega20/gfx906) through rusticl.

what llvm version did you target? I ran into https://github.com/CHIP-SPV/chipStar/issues/429#issuecomment-2066356579 myself in a lot of tests

pjaaskel:
I created a chipStar discussion for this topic, I suggest to continue there: https://github.com/CHIP-SPV/chipStar/discussions/833

include/CL/cl_ext_pocl.h
pjaaskel: I sent a PR of the extension text for review feedback: https://github.com/KhronosGroup/OpenCL-Docs/pull/1159 @karolherbst @linehill  pls provide feedback.

PR #1448, opened 2024-04-16
Some fixes

pjaaskel:
Please check the commits list.

### DIFF
--- a/lib/CL/devices/level0/pocl-level0.cc
+++ b/lib/CL/devices/level0/pocl-level0.cc
@@ -1393,7 +1393,7 @@ cl_int pocl_level0_set_kernel_exec_info_ext(
 
   assert(Kernel->data[ProgramDeviceI] != nullptr);
   Level0Kernel *L0Kernel = (Level0Kernel *)Kernel->data[ProgramDeviceI];
-  assert(L0Kernel != NULL);
+  assert(L0Kernel != nullptr);
   switch (param_name) {
   case CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM: {
     if (Dev->svm_caps & CL_DEVICE_SVM_FINE_GRAIN_SYSTEM)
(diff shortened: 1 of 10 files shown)


lib/CL/devices/level0/pocl-level0.cc
```diff
@@ -1393,7 +1393,7 @@ cl_int pocl_level0_set_kernel_exec_info_ext(
 
   assert(Kernel->data[ProgramDeviceI] != nullptr);
   Level0Kernel *L0Kernel = (Level0Kernel *)Kernel->data[ProgramDeviceI];
-
+  assert(L0Kernel != NULL);
```
jansol, line 1396:
```suggestion
  assert(L0Kernel != nullptr);
```

Using the NULL macro in C++ code is considered non-idiomatic AFAIK.

jansol (approved):

lib/CL/pocl_llvm_build.cc
```diff
@@ -418,14 +418,19 @@ int pocl_llvm_build_program(cl_program program,
 
   std::string AllBuildOpts = ss.str();
 
+  // Convert the -g to more specific debug options understood by the frontend.
+  AllBuildOpts = std::regex_replace(
+      AllBuildOpts, std::regex("-g(\\s+)"),
```
franz, line 423: AFAICT this regex needs at least a whitespace before, otherwise it could match things we don't want to match, e.g.
 `... -I/some/path/ending/with/letter-g ...`. And ideally we should not use regex for this kind of replacement at all, so a TODO would be nice here...

pjaaskel:
```
PoCL-D: in fn virtual int SharedCLContext::freeBuffer(BufferId_t, bool) at line 2195:  *** ERROR ***  |     ERROR |  Did not find the SVM chunk to free at 0x7fee43fff100. A double free attempt?
```
says the PoCL-R bot, but I cannot repro this locally. Some race somewhere, maybe.

PR #1456, opened 2024-05-03
PoCL-R: Simple USM indirect implementation and other fixes

pjaaskel:


pjaaskel:
The macOS cases are failing on main as well due to the builders updated to arm64. @franz will push a commit that will skip those tests for macOS as well.

PR #1478, opened 2024-05-22
CUDA: relax CL_DEVICE_MAX_MEM_ALLOC_SIZE limit

linehill:
Relax the limit from 1/4 * CL_DEVICE_GLOBAL_MEM_SIZE to currently available free memory reported by cuMemGetInfo() at device initialization time.

Motivation: enable a HeCBench case, running on chipStar/OpenCL->PoCL/CUDA->RTX3060, to allocate ~4GiB buffers.

Edit: was: Relax the limit from 1/4 to 1/2 * CL_DEVICE_GLOBAL_MEM_SIZE.


franz:
I think we can go much higher than 1/2.  @pjaaskel @isuruf what do you think ?

pjaaskel:
Yes, I don't see why not. Let's match what the NV propr. driver reports?

linehill:
The limit is now the free memory reported by CUDA.

PR #1486, opened 2024-05-27
L0 sub-buffers fix

pjaaskel:
It didn't add the offset of the sub-buffers when calling the kernel.

Fixes the conformance test and the new subbuffer test I add with this PR.


pjaaskel:
The SYCL CI has a reliability prob currently.

PR #1487, opened 2024-05-27
update GH workflow with new oneAPI CI

franz:


franz:
rancher is currently offline, should be restored soon.

PR #1491, opened 2024-06-03
Move command enqueues out of critical section

gtkiku:
Fixes possible deadlock where `pocl_create_migration_commands()` and `pocl_free_event_memobjs()` may try to acquire the same memory object lock. Occurs in `example1` when compiled with

```
  cmake -DENABLE_ALMAIF_DEVICE=ON -DENABLE_POCL_BUILDING=ON ..
```


jansol:
Could you move the debug messages as well so they stay next to the code they are referring to?

jansol (approved):

PR #1495, opened 2024-06-04
Sub-buffer rework and a bunch of cleanups and fixes

pjaaskel:
* Use the sub-buffer like it was a regular buffer wherever feasible instead of pretending it is the parent buffer. This enables fine-grain implicit migrations of the sub regions. The migrations are done per sub-buffer now and when migrating a parent buffer with sub-buffers, the "holes" (without client sub-buffers) are migrated using implicit extents using hidden sub-buffers.
* Adds PoCL-R sub-buffer support.
* A new debug aid for dumping all commands/events to a .dot task graph. Proven very useful while debugging the implicit migration code.

No parallel sub-buffer access yet. The implicit migration still creates an overly cautious dependency chain. We need to analyze the deps backwards and retain some history of commands to be able to remove unnecessary migrations and deps.


franz:
Some of the CI failures (`mkdir: cannot create directory ‘/home/github’: Permission denied`) should be fixed by rebasing the PR on main branch, however some seem like genuine problems that need fixing:

```
../../../lib/CL/clCreateSubBuffer.c: At top level:
../../../lib/CL/clCreateSubBuffer.c:202:8: error: ‘clCreateSubParent’ undeclared here (not in a function)
  202 | POsym (clCreateSubParent)
```

tests/runtime/test_subbuffers.cpp
```diff
@@ -145,10 +154,7 @@ int TestOutputDataDecomposition() {
 
     std::vector<int> AfterSubBufCContents(NumData);
 
-    for (size_t i = 0; i < Queues.size(); ++i)
-      Queues[i].finish();
-
-    Queues[0].enqueueReadBuffer(CBuffer, CL_TRUE, 0, sizeof(cl_int) * NumData,
+    Queues[0].enqueueReadBuffer(CBuffer, CL_FALSE, 0, sizeof(cl_int) * NumData,
```
jansol, line 157: Seems this explicitly does *not* synchronize between queues since they use separate non-overlapping sub-buffers? It's probably a good idea to explain this at the start of the file, since it's likely not something most people would expect in an OpenCL application.

tools/scripts/test_remote_runner_multi.sh
```diff
@@ -92,7 +92,7 @@ set +e
 
 RESULT=3
 WAIT=1
-while [ $WAIT -le 10 ]; do
+while [ $WAIT -le 1000 ]; do
```
jansol, line 95: Maybe remove the echo from the next line, this gets very spammy.

tools/scripts/test_remote_runner_single.sh
```diff
@@ -74,7 +74,7 @@ set +e
 
 RESULT=3
 WAIT=1
-while [ $WAIT -le 10 ]; do
+while [ $WAIT -le 1000 ]; do
```
jansol, line 77: Ditto about the echo

pocld/common.hh
```diff
@@ -402,10 +402,11 @@ struct EventPair {
...
 
 // Type for the buffer allocation identifier, a running number starting from
-// 0. Note: In some parts functions, the input can be declared as a buffer
-// id which might be treated as a raw 64b SVM pointer instead of a cl_mem
-// buffer identifier, in case referring to SVM buffers.
-typedef uint64_t BufferId_t;
+// 0 with 0 denoting a NULL/illegal buffer.
+// Note: Currently in some parts functions, the input can be declared as
+// a buffer id which might be treated as a raw 64b SVM pointer instead of
+// a cl_mem buffer identifier, in case referring to SVM buffers (WiP).
```
jansol, line 408: Probably worth turning this into a doc comment (`/// ...` or `/** ... */`) while at it. Then it should get displayed in the LSP tooltip when hovering over the type name elsewhere in the code.

pocld/shared_cl_context.cc
```diff
@@ -2117,16 +2121,39 @@ int SharedCLContext::createBuffer(BufferId_t BufferID, size_t Size,
...
-      new cl::Buffer(ContextWithAllDevices, Flags, Size, HostPtr, &Err));
+  clBufferPtr Buf;
+
+  if (ParentID == 0) {
+    Buf = clBufferPtr(
+        new cl::Buffer(ContextWithAllDevices, Flags, Size, HostPtr, &Err));
+  } else {
+    std::unique_lock<std::mutex> Lock(BufferMapMutex);
+    assert(BufferIDmap.find(ParentID) != BufferIDmap.end());
+    clBufferPtr &Parent = BufferIDmap[ParentID];
```
jansol, line 2141: This unnecessarily re-does the lookup. Better to store the result from find in a variable before the assert instead and then do `clBufferPtr &Parent = FindResult->second` afterwards.

Also, since the BufferIDMap is not actually modified here I think it's safe to release the lock immediately after the find & assert?

lib/CL/devices/remote/communication.c
```diff
@@ -2082,10 +2083,15 @@ pocl_network_create_buffer (remote_device_data_t *ddata, cl_mem mem,
 
   nc.request.m.create_buffer.flags = mem->flags;
   nc.request.m.create_buffer.size = mem->size;
-  // see https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Pointer_002dInteger-Conversion.html
-  // for reason behind the double cast
+  /* See https://www.gnu.org/software/c-intro-and-ref/manual/html_node/
```
jansol, line 2086: This breaks the link, should it be surrounded in `<>` to keep it intact during formatting?

lib/CL/devices/remote/remote.c
```diff
@@ -2512,6 +2537,8 @@ pocl_remote_driver_pthread (void *cldev)
   _cl_command_node *cmd = NULL;
   _cl_command_node *finished = NULL;
 
+  /* Sleep so we have time to run the task graph dumper in main(). */
+  /* sleep (2); */
```
jansol, line 2541: Leftover debug code?

lib/CL/pocl_cl.h
```diff
@@ -445,27 +458,45 @@ struct pocl_device_ops {
...
+  /****** Memory management APIs. */
+
+  /* Allocate a buffer in the device's global memory. */
   cl_int (*alloc_mem_obj) (cl_device_id device, cl_mem mem_obj, void* host_ptr);
-  /* free a device buffer */
+  /* Free a device buffer. */
   void (*free) (cl_device_id device, cl_mem mem_obj);
 
-  /* return >0 if driver can migrate directly between devices.
+  /* Optional: Allocate/register a sub-buffer in the device's global memory.
```
jansol, line 468: Doc comments start with `/**` to distinguish them from "normal" comments

lib/CL/pocl_util.c
```diff
@@ -1059,13 +1071,205 @@ pocl_create_migration_commands (cl_device_id dev, cl_event *ev_export_p,
...
+                         sub_buf->id, size, origin);
+}
+
+/**
+ * Generates "implicit" sub-buffers that covers the parts of the parent
+ * buffer that are not covered by user-created sub-buffers.
+ *
+ * The implicit sub-buffers are used to implicitly synchronize the parent
+ * buffer's uncovered parts when accessing the parent buffer after modifying
+ * the sub-buffer.
```
jansol, line 1123: Was there some clarification on whether these are actual OpenCL sub-buffers somewhere? Also how/where are they managed?

jansol (changes requested):
Looks promising, I quite like it!

lib/CL/pocl_util.c
```diff
@@ -1117,95 +1338,158 @@ pocl_create_command_full (_cl_command_node **cmd,
...
-              pocl_release_mem_host_ptr (buffers[i]->size_buffer);
-              POname (clReleaseEvent) (size_events[i]);
-              size_events[i] = NULL;
-            }
-        }
+        if (mi->buffer->size_buffer != NULL)
+          {
+            /* BLOCK until size buffer has been imported to host mem! No event
+             * exists if host import is not needed. */
+            if (size_events[i] != NULL)
```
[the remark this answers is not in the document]
pjaaskel: My gut feeling is that content size and sub-buffers are typically used for similar purposes of saving unnecessary data transfers (the former also for dynamic content size) and thus not mixed, thus I'd implement the buf content size only for client buffers. My understanding is that this should work like that already? If the user passes a buffer (sub-buffer or a parent one), it should track its size. Or perhaps I didn't catch what you meant fully.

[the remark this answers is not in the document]
jansol: If we have a content size buffer we'll want to not migrate regions past the content size even if the kernel accessed them, no?

My understanding is that sub-buffers get somewhat automatically created for top-level buffers that are migrated? But I didn't see anywhere that accounts for content size of the parent when dealing with sub-buffers.

lib/CL/pocl_util.c
```diff
@@ -1312,47 +1610,50 @@ pocl_create_recorded_command (_cl_command_node **cmd,
...
-      for (unsigned i = 0; i < num_buffers; ++i)
-        {
-          POname (clRetainMemObject) (memobjs[i]);
-        }
-    }
+  (*cmd)->migr_infos = buffer_usage;
 
+  pocl_buffer_migration_info *migr_info = NULL;
+  LL_FOREACH (buffer_usage, migr_info)
+    POname (clRetainMemObject) (migr_info->buffer);
```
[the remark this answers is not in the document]
pjaaskel: Yep, the latter.

pocld/shared_cl_context.cc
pjaaskel: I've understood STL doesn't guarantee thread safety even for reads of the container in case there might be mutual updates. Better be safe than sorry with potential races and not micro-optimize prematurely.

pjaaskel:
@jansol thanks for the review. Addressed/fixed most of them.

lib/CL/pocl_util.c
```diff
@@ -1117,95 +923,136 @@ pocl_create_command_full (_cl_command_node **cmd,
...
+      }
+
+    LL_FOREACH (buffer_usage, mi)
+      {
+        /* The last events of the parent buffers can be now released as the
+           sub-buffer references to the events should hold them alive. */
+        if (mi->buffer->parent == NULL && mi->buffer->sub_buffers != NULL)
+          {
+            int new_ref;
+            POCL_RELEASE_OBJECT_UNLOCKED (mi->buffer->parent->last_updater,
```
jansol, line 1014: This only decrements the refcount and does NOT actually release anything. The next line seems to assume it does, though?

lib/CL/pocl_util.c
```diff
@@ -2415,39 +2266,24 @@ pocl_copy_event_node (_cl_command_node *dst_node, _cl_command_node *src_node)
               src_node->command.svm_fill.pattern_size);
       break;
 
-#if 0
-    /* these cases are currently not handled in pocl_copy_event_node,
+    /* These cases are currently not handled in pocl_copy_event_node,
```
jansol, line 2269:
```suggestion
    /* These cases are currently not handled in pocl_copy_command_node,
```

pjaaskel:
Cleaned up the history. Now (at least) the internal tests pass with and without the top commit reverted.

pjaaskel:
Yey. Finally all green. I'll see if I can launch the longer test runs on it before pulling in.

pjaaskel:
I think we need to get rid of the github CIs, they are too moody.

pjaaskel:
The [full CTS run](https://github.com/pocl/pocl/actions/runs/9891957165/job/27323502129) fails two tests cases: The first one should get fixed with a CTS update on the CI; it doesn't recognize the multi-device CB extension but that is fixed in our CTS branch. The image error I fixed locally. It seemed to be a genuine error as it computed min buffer size wrongly for 1D images. The [patch](https://github.com/pocl/pocl/pull/1495/commits/21d94b57fdf89e706ef0f8d497438bd2bf416f34) is in this PR, but I don't know what it still fails in the bot. It works locally.

The MacOS issues are due to the old flakiness which could be due to some custom LLVM version. 

I'm pulling this in. It's easy to revert if things go ugly.

PR #1499, opened 2024-06-12
rename llvmGetPassPluginInfo() to pocl_llvmGetPassPluginInfo()

anbe42:


franz:
I think this particular function cannot be renamed. From [LLVM docs](https://llvm.org/doxygen/PassPlugin_8h.html):

> llvmGetPassPluginInfo()
> The public entry point for a pass plugin.
> 
> When a plugin is loaded by the driver, it will call this entry point to obtain information about this plugin and about how to register its passes. This function needs to be implemented by the plugin, see the example below:

OTOH i don't think SVMOffset pass is required unless we're compiling pocld (remote server), so the entire pass could be conditionally disabled.

anbe42:
Oh, that symbol is LLVM API and not PoCL API ...

But in that case I don't think it belongs into a public shared library. Just imagine another OpenCL ICD brings the same symbol. Which one is going to win? (Still assuming both ICDs are built against the same llvm version, I really don't want to open the can of worms when mixing different llvm versions in one process.)

Feel free to ignore this for 6.0, I don't consider it  blocker.

pjaaskel:
Yes, you are right we shouldn't link this nor the other LLVM passes we register to the pass manager to libpocl.so. In this case the pass is even called from the command line (opt) so it shouldn't be linked to libpocl.so, but I started to wonder if we have a similar problem with the kernel compiler passes in general.

pjaaskel:
...yeah I guess that's the problem here. I've implemented that hook since that's the way opt registers passes loaded dynamically since I use opt in the pocld code. Other LLVM passes used by the kernel compiler are registered internally when libpocl.so is loaded and we call pass manager directly for the kernel compilation.

The clean solution here would be to use the PM in PoCL-D directly and not export that hook or compile the SVMOffset pass to a separate dynlib which is loaded by the opt instead of the whole libpocl.so it does not.  It should be OK to postpone this as this hook is supposed to work like this (defined weakly - which is dlopene'd last should get the function defined).

PR #1504, opened 2024-06-20
remote: Rework reconnection code

jansol:
Apparently the old reconnection code was reliant on undefined behaviour that does not work very well on Linux, which became apparent while benchmarking reconnection. Implemented a more correct solution with pipes for signaling between thread pairs instead, as well as a couple of drive-by fixes.


pjaaskel (changes requested):
Some // comments in C files. It also would be useful to document the reconnect mechanism somewhere centrally (as a code comment) as it's a bit complex now.

jansol:
I'll see about a proper explanation of the socket thread logic later, possibly as a follow-up PR.

pjaaskel:
Pls rebase.

PR #1505, opened 2024-06-26
refactor level zero device setup + minor fixes

franz:
the new device setup should better support devices which don't have all features of GPU (some of the zeGet...Properties return error)


pjaaskel:
Pls drop [2665c1e](https://github.com/pocl/pocl/pull/1505/commits/2665c1e44d2b6513f18da50f54efa460c7437507) and rebase.

PR #1514, opened 2024-07-18
Almaif: update OpenASIP-specific doc

TopiLeppanen:
Inspired by recent questions about the AlmaIF-driver, I wrote a clarifying section to AlmaIF-driver's documentation specific to OpenASIP, and how we currently support it in the AlmaIF-driver.


TopiLeppanen:
> Thanks, this should be useful for adopters. Some minor suggestions added.

Thank you, very sensible suggestions. Applied.

PR #1518, opened 2024-07-23
Fixes to make OpenVINO run on the L0 driver

pjaaskel:
Small patches to make OpenVINO run the tiny-llama-1b-chat model infer. via the L0 driver.

An example run:
```
samples/cpp/multinomial_causal_lm/multinomial_causal_lm ~/src/openvinino_notebooks/notebooks/llm-chatbot/tiny-llama-1b-chat/INT4_compressed_weights/ "Why is the Sun yellow?"
PoCL Level0
6 warnings generated.
12 warnings generated.
The answer, as you would expect, is because sunlight contains an abundance of atoms of iron (Fe), a naturally occurring element that is red-coloured because of the presence of one of its elements.
 
12. Why are there stars with blue-grey hues? Blue-grey colours are produced by iron-rich regions of the Sun, which contain large numbers of elements like nickel (Ni), cobalt (Co), and magnesium
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ apt install -y python3-dev libpython3-dev build-essential ocl-icd-libopencl1 \
     llvm-${LLVM_VERSION}-dev
 ```
 
+For LLVM 18 and above also `libpolly-${LLVM_VERSION}-dev libzstd-dev` are
+needed for static LLVM linkage.
+
 If your distro does not package the version of LLVM you wish to build against
 you might want to set up the
 [upstream LLVM package repository](https://apt.llvm.org/).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -45,6 +45,9 @@ apt install -y python3-dev libpython3-dev build-essential ocl-icd-libopencl1 \
     llvm-${LLVM_VERSION}-dev
 ```
 
+For LLVM 18 and above also `libpolly-${LLVM_VERSION}-dev libzstd-dev` are
```
franz, line 48: This is actually needed also for earlier LLVM(s) when using static linking, i've just checked and LLVM 17 definitely requires libpolly-17-dev; earlier LLVMs possibly but i haven't checked.

franz: i'll fix it manually

Issue #1525, opened 2024-07-26
Kernel execution fails with SIGBUS (Misaligned address error)


fjarri:
MacOS 14.5, pocl 6.0 (installed via homebrew)

`kernel.cl`: https://gist.github.com/fjarri/f86652b38cc61969d0c7933ad3f27d36
`test.c`: https://gist.github.com/fjarri/7f8fb6de256bba5292044807a64dc825
(sorry for the sloppy C code, the example was reduced to MRE from a large Python codebase)
```
$ gcc test.c -lpocl -L/opt/homebrew/Cellar/pocl/6.0/lib`
$ /a.out
Platform: Portable Computing Language
Device: cpu
Before synchronizing the queue
fish: Job 1, './a.out' terminated by signal SIGBUS (Misaligned address error)
``` 

As far as I understand, the kernel should simply copy `input` into `output`. All the memory accesses are within the buffer ranges. The kernel works as expected if:
- one removes the local mem barrier (even though local mem is not even used)
- explicit `local_work_size` when invoking the kernel is substituted with `NULL` (that is, if the local work size choice is automatic)
- `local_work_size` and `global_work_sze` are less than 1024+512. 
- the unrolled memory loads/saves are substituted with loops

pjaaskel:
Due to the barrier PoCL generates a lot of context data, for each of the compute i increment, in addition to the user array, which leads to large stack consumption. On MacOS the stack is limited to 640KB (? pls verify with 'ulimit -s') which easily gets exhausted by this case.

```
pjaaskel@pjaaskel-mobl1:~/src/pocl/build-cpu$ ulimit -s 640
pjaaskel@pjaaskel-mobl1:~/src/pocl/build-cpu$ tests/regression/test_issue_1525
Segmentation fault
```
Works here otherwise (Linux, x86, LLVM 17). Where did you get that 'Misaligned address error' from? It would hint towards unaligned access, but who knows what exhausting the stack could cause...

The correct fix would be to return CL_OUT_OF_RESOURCES from clEnqueueNDRangeKernel() when detecting the stack will exceed the process limit: _"CL_OUT_OF_RESOURCES if there is a failure to queue the execution instance of kernel on the command-queue because of insufficient resources needed to execute the kernel. For example, the explicitly specified local_work_size causes a failure to execute the kernel because of insufficient resources such as registers or local memory."_

[...]

PR #1526, opened 2024-07-27
Add regression test for #1525 (sigbus crash)

fjarri:
Add regression test for #1525 

Not sure if it will work on CI, may be dependent on the specific device and its parameters (maximum workgroup size, memory alignment size etc).


pjaaskel:
You are likely running out of stack due to the excessive amount of WI context data this one wants to create due to the barrier in between.

fjarri:
> Is this with LLVM 16? 

The `homebrew` metadata claims that it was built with 18.1.8. 

> What if you force the alignment of that local array by __attribute__((align(256)) or such?

You mean `aligned`? `align` doesn't seem to be recognized by the compiler. In any case, it does not affect the crash.

pjaaskel:
The test is in and should pass now thanks to remat reducing the context size.

PR #1535, opened 2024-08-15
WG-vectorizer: Re-enable vectorization

pjaaskel:
The vectorizers were accidentally not enabled when transitioning to the new PM. This commit re-enables them and exposes some new inefficiencies (to be continued...).

Yeah, we should also add a perf.regression test or at least a "smoke test" for the WG vectorization.


franz:
The `populateModulePM` was outside the `PoCLModulePassManager` class because it's also called by nvidia (to avoid code duplication).

WRT the other failures, i suspect they are caused by reusing the PassBuilder instance (`PassBuilder &PB = *PassB.get()`... this is why the `SEPARATE_OPTIMIZATION_FROM_POCL_PASSES` was introduced), and additionally the code is now registering new on-stack instances (FAM, LAM, MAM etc) with the reused PB instance which already had these xAMs registered once.

pjaaskel:
The ARM64 LLVM 17 barriered_for_loops crash might be an LLVM issue since it disappears with LLVM 18. I think we could even drop the LLVM 17 ARM64 bot as we should move on to LLVM 19 soon, but I'll disable that test now.

franz:
its seems the `test_alignment_with_dynamic_wg_114_cbs` is still failing with `LLVM 14 - OpenMP` CI config

pjaaskel:
This is new and interesting:

```
240/268 Test  #340: piglit_cl_api_get_event_info .................................Subprocess aborted***Exception:   0.31 sec
## Test: clGetEventInfo (/srv/github2/actions-runner/examples/build_tier1/piglit/src/piglit/tests/cl/api/get-event-info.c) ##

# Running on:
#   Platform: Portable Computing Language
#   Device: cpu
#   OpenCL version: 2.0
CL_EVENT_COMMAND_QUEUE 
CL_EVENT_COMMAND_TYPE 
CL_EVENT_REFERENCE_COUNT 
CL_EVENT_COMMAND_EXECUTION_STATUS 
CL_EVENT_CONTEXT 
# Result:
PIGLIT: {"result": "pass" }
PTHREAD ERROR in pocl_update_event_finished():2215: Invalid argument (22)
```
I wonder do we free events too early somewhere or something.

PR #1539, opened 2024-08-21
Poclr messages

RABijl:
some message related additions as part of the upstreaming process

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,7 +304,7 @@ option(ENABLE_EXAMPLES "Build examples. Defaults to ON" ON)
 
 option(ENABLE_RDMA "Enable usage of RDMA libraries for memory allocations. Requires libRDMAcm and libverbs" OFF)
 
-option(ENABLE_TRAFFIC_MONITOR "Enable network traffic monitoring & logging in remote device" OFF)
+option(ENABLE_TRAFFIC_MONITOR "Enable network traffic monitoring & logging in remote device" ${ENABLE_REMOTE_CLIENT})
 
 OPTION(RENAME_POCL "rename PoCL's ocl functions to PO<ocl_function>. Allows an user to call both PoCL and another ocl application" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,8 +304,6 @@ option(ENABLE_EXAMPLES "Build examples. Defaults to ON" ON)
 
 option(ENABLE_RDMA "Enable usage of RDMA libraries for memory allocations. Requires libRDMAcm and libverbs" OFF)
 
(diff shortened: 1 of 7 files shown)


lib/CL/devices/remote/communication.c
```diff
@@ -1321,6 +1321,29 @@ traffic_monitor_pthread (void *arg)
   free (a);
   return NULL;
 }
+
+__attribute__ ((visibility ("default"))) void
```
pjaaskel, line 1325: What's the attribute for?

pjaaskel (changes requested):

lib/CL/devices/remote/communication.c
RABijl: the rest of the file has the visibility hidden and this function is used by clients to get info about the amount of data being sent.

jansol: The functions that are actually meant to be part of the ABI are defined in the header with POCL_EXPORT but I didn't want to advertise this function like that originally when it was just for some internal testing. If we want to integrate this more "officially" the function should be declared in the corresponding header with the appropriate macros.

pjaaskel: I already discussed the PoCL-specific (clGetDeviceInfo device query) extension with Yash. I think that's the clean way to go. Pls discuss with @Yashvardhan-A

RABijl: I've made the change by using clGetDeviceInfo

CMakeLists.txt
```diff
@@ -304,7 +304,7 @@ option(ENABLE_EXAMPLES "Build examples. Defaults to ON" ON)
 
 option(ENABLE_RDMA "Enable usage of RDMA libraries for memory allocations. Requires libRDMAcm and libverbs" OFF)
 
-option(ENABLE_TRAFFIC_MONITOR "Enable network traffic monitoring & logging in remote device" OFF)
+option(ENABLE_TRAFFIC_MONITOR "Enable network traffic monitoring & logging in remote device" ${ENABLE_REMOTE_CLIENT})
```
pjaaskel, line 307: What's the reason making this optional? The data is accumulated unconditionally so there should not be additional runtime overheads here.

pjaaskel (changes requested):

CMakeLists.txt
RABijl: The code was originally written as an optional thing and spawns an extra thread for these stats. I've kept the option to turn it off if so desired.

pjaaskel: But does it impact the overhead. It's not disabling the collection of the status from what I see? I prefer not to have options that are not clearly useful as they clutter the code and also make it more difficult test the different (build) paths.

PR #1544, opened 2024-08-28
poclr: fix build on arch linux

RABijl:
Building on arch linux had the unfortunate
issue of not finding std::find and std::for_each
unless explicitly importing algorithm.
There was also a case of a size and null pointer
being swapped which caused a casting error.


jansol (approved):
Default implicit includes should be banned, they keep showing up to bite with every single compiler update.

franz:
The TSan failure is an actual bug in my recent code, i'll fix it soon

PR #1546, opened 2024-08-30
lttng: get mem_obj info from command instead

RABijl:
This is related to the experimental changes of
commit: a0acd3aebbf5b389981c04f448bf296e7c0a4863
which removes the memobjs from events.

This went unnoticed since lttng is not built by default, maybe it is time to update the test environment?


franz:
There used to be one (before Github CI), so it should not be a problem. The userspace libraries for lttng are in the ubuntu repos

PR #1547, opened 2024-08-30
pthread: Add workgroup function builtin kernels

jansol:
We needed a bunch of builtin kernels on the pthread device for internal testing and initially went with an similar approach as was introduced in #1521 . This however caused deadlocks in our use case (multiple interdependent command queues) since the builtin kernels block the pthread scheduler. To fix this we decided to make use of the existing infrastructure for dynamically loading LLVM-generated workgroup functions by moving the builtin kernels to separate dynamic libraries and adding the ability to manually specify the library name to load into the driver's dynamic library cache for kernels, and formulating the BiK functions as workgroup functions.

This is a bit of a departure from #1521 and I'm not entirely sure how well it works with the recent DBK work or if this is even a direction that we want to take with builtin kernels. The interesting bits are the dlhandle cache stuff in common.c (which I am quite tempted to pull into its own files at this point) as well as the (relative lack of) changes to pthread_scheduler.c.

### DIFF
--- a/lib/CL/devices/basic/basic.c
+++ b/lib/CL/devices/basic/basic.c
@@ -490,7 +490,7 @@ static void
 basic_command_scheduler (pocl_basic_data_t *d)
 {
   _cl_command_node *node;
-  
+
   /* execute commands from ready list */
   while ((node = d->ready_list))
     {
@@ -517,7 +517,7 @@ pocl_basic_submit (_cl_command_node *node, cl_command_queue cq)
       if (!program->builtin_kernel_attributes)
         {
           node->command.run.device_data
-            = pocl_check_kernel_dlhandle_cache (node, CL_TRUE, CL_TRUE);
+            = pocl_check_kernel_dlhandle_cache (node, CL_TRUE, CL_TRUE, NULL);
         }
     }
 
@@ -592,7 +592,7 @@ pocl_basic_compile_kernel (_cl_command_node *cmd, cl_kernel kernel,
   char *saved_name = NULL;
   pocl_sanitize_builtin_kernel_name (kernel, &saved_name);
   if (cmd != NULL && cmd->type == CL_COMMAND_NDRANGE_KERNEL)
-    pocl_check_kernel_dlhandle_cache (cmd, CL_FALSE, specialize);
+    pocl_check_kernel_dlhandle_cache (cmd, CL_FALSE, specialize, NULL);
   pocl_restore_builtin_kernel_name (kernel, saved_name);
 }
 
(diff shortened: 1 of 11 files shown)


pjaaskel:
Please read the DBK and cl_tensor spec. We should add BiKs that are generic (not specific to a driver) there and utilize the APIs/infrastructures of it. If we need to extend the spec then we'll do so. The DBK extension is at a prototype spec.

Another question is that why bother with the dynamically loaded BiK implementations? What's the motivation for that added complexity instead of just compiling the implementations in (for now)?

jansol:
This predates the current DBK efforts. The motivation for dynamically loaded BiKs was to reuse the existing kernel scheduling code which ingests dynamically loaded functions (which are compiled/cached at runtime). And the motivation for reusing the same scheduling code as OpenCL C kernels was that implementing a separate code path for builtin kernels like (is done now in the DBK infra) was consistently causing deadlock situations in our test application.

Changing the scheduler code to accept arbitrary function pointers instead of always loading a dynamic library would have been an option but seemed like a lot more work (and more controversial) than simply moving the builtin kernels out into libraries at that time.

franz:
> OpenCL C kernels was that implementing a separate code path for builtin kernels like (is done now in the DBK infra)

There has been infrastructure to implement "regular" (non-DBK) builtin kernels (BiKs) from sources using OpenCL C sources for quite a while (both driver specific in `lib/CL/devices/<driver>/builtins.cl` and "cross driver" in `lib/CL/devices/common_builtin_kernels.cl`). Once those source-based BiKs are built, they should behave exactly like any other kernel (WRT scheduling). I should clarify this should still be the case after DBK PR was pulled in - *only* the DBKs have special scheduling in the CPU driver, the non-DBK BiKs are scheduled like before, perhaps the code is not very clear on this (or has a bug). IMO we should use this (OpenCL C based) approach for all builtin kernels that are possible to implement this way, as this reuses most of the existing driver code.

The current code however does not take into account the possibility of future DBKs which are source-based; it assumes all DBKs are implemented via external libraries, this is TBD. And yeah the scheduling can be tricky, not least because some DBK implementations can have their own scheduling, e.g. libxsmm has the possibility to use OpenMP (but it can also be used single-threaded).

But it's unclear to me what kind of deadlock did you experience, can you elaborate ? AFAICT the current code should block only a single driver thread while executing the DBK, similarly to how it blocks when executing using `pocl_exec_command()`.

pjaaskel:
Is this still relevant? I believe the develoment went elsewhere?

jansol:
Yeah this ended up not being the path that was chosen for DBKs

PR #1554, opened 2024-09-06
GH workflow updates, fixes to cl_offline_compiler.sh (SPIR-V CTS)

franz:
...by checking undefined symbols *after* we have linked in all of the functions from the builtin library. This is necessary, because the SPIR-V wrapper always creates all wrapper functions, even if the builtins which are wrapped don't exist in the builtin library (e.g. because of compiling with ENABLE_CONFORMANCE=1 which disables many extensions).

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -56,7 +56,8 @@ jobs:
             config: chipstar
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
 
       - name: CMake
         id: cmake
@@ -196,7 +197,8 @@ jobs:
             config: static
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
 
       - name: Install prerequisites and ocl-icd
         run: >
@@ -292,7 +294,8 @@ jobs:
         llvm: [18]
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
 
       - name: CMake
         id: cmake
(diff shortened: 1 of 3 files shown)


lib/llvmopencl/linker.cpp
```diff
@@ -422,70 +469,56 @@ int link(llvm::Module *Program, const llvm::Module *Lib, std::string &Log,
...
+  DeclaredFunctions.clear();
+  for (FI = Program->begin(), FE = Program->end(); FI != FE; FI++) {
+    if (FI->isDeclaration()) {
+      DB_PRINT("Post-link: %s is not defined\n", fi->getName().data());
+      DeclaredFunctions.insert(FI->getName());
+      continue;
+    }
+  }
 
+  bool found_all_undefined = true;
```
pjaaskel, line 494: Pls convert to C++ naming convention.

lib/llvmopencl/linker.cpp
```diff
@@ -422,70 +469,56 @@ int link(llvm::Module *Program, const llvm::Module *Lib, std::string &Log,
...
+    Function *F = Program->getFunction(FName);
+
+        if ((F == NULL) || (F->isDeclaration() &&
                             // A target might want to expose the C99 printf in
                             // case not supporting the OpenCL 1.2 printf.
-                            !f->getName().equals("printf") &&
-                            !f->getName().equals(pocl_sampler_handler) &&
-                            !f->getName().starts_with(llvm_intrins))) {
+                            !F->getName().equals("printf") &&
+                            !F->getName().equals(pocl_sampler_handler) &&
```
pjaaskel, line 509: Use == instead of equals() for string comparisons. equals() is gone in LLVM 19.

pjaaskel: (this will be fixed in my LLVM 19 PR so there might be a conflict)

pjaaskel (changes requested):

lib/llvmopencl/linker.cpp
franz: fixed

lib/CL/pocl_debug.h
```diff
@@ -169,6 +169,31 @@ POCL_EXPORT
 
         #define POCL_DEBUG_HEADER(FILTER, FILTER_TYPE) \
             pocl_debug_print_header (__func__, __LINE__, #FILTER, FILTER_TYPE);
+/**
+ * Dumps all commands and events of the context as a graphviz dot file.
+ *
+ * Currently works only with PoCL contexts.
+ *
+ * @param context the OpenCL context to dump.
```
pjaaskel, line 177: Pls change to \param in the next PR.

PR #1559, opened 2024-09-11
Kernel compiler refactorings, cleanups and improvements (LLVM 19-rc4 passes 'make check')

pjaaskel:
A lot of cleanups, optimizations, refactorings and misc. improvements to the kernel compiler passes. 

LLVM 19-rc4 passes 'make check' with this branch (also for CBS, unlike in main). 

I haven't tested other LLVMs yet. Let's see the extent of CI failures should we merge this in to 6.1 or not. I can also upstream the CBS failure fix (the two topmost commits should do) separately if it looks too bad .


pjaaskel:
@illuhad - I wonder if you have updates/fixes to CBS in AdaptiveCPP we could merge?

The current plan is to utilize CBS as a fallback for too complex CFG cases where WILoops would not produce nicely vectorizable parallel loops anyhow, and simplify/cleanup WILoops code base towards LLVM upstreaming. Thus CBS should get more use from now on if fallback cases appear (there haven't been many so far).

illuhad:
@pjaaskel I think there were some fixes that went in - @fodinabor, I think you have a better picture?

pjaaskel:
This is obsolete. Splitting the commits to smaller PRs and changing the approach for the context handling.

PR #1560, opened 2024-09-12
Jpeg dbk

RABijl:
Putting the commit out there for feedback while I work on the decompression dbk.

highlights:
* uses libjpeg-turbo api 3, so will require from source build of that
* builds with and without said lib, but you can not use said dbk
* only added to the pthread driver for now
* saves the configured jpeg handle as part of the kernel device data, so each device has its own instance
* handle gets freed when kernel is freed.
* test program with test data provided by @juliusikkala 
* added decompression dbk.
* dbk has been tested on basic and cpu
* add rst entry.

left to do:
* add configurable jpeg supsampling parameter attribute instead of defaulting to the most common one.
* remove internal memcpy


lib/CL/pocl_builtin_kernels.c
```diff
@@ -804,6 +820,11 @@ pocl_validate_dbk_attributes (BuiltinKernelId kernel_id,
         return GemmCB (Attrs->trans_a, Attrs->trans_b, &Attrs->a, &Attrs->b,
                        NULL, &Attrs->c, NULL, NULL);
       }
+    case POCL_CDBI_DBK_KHR_JPEG_DECODE:
+    case POCL_CDBI_DBK_KHR_JPEG_ENCODE:
+      {
```
pjaaskel, line 825: No need for brackets.

tests/runtime/test_dbk_jpeg.c
```diff
@@ -0,0 +1,132 @@
+//
```
pjaaskel, line 1: Use the header/license/copyr text format as in other files.

pjaaskel (changes requested):

include/CL/cl_exp_defined_builtin_kernels.h
```diff
@@ -90,6 +90,8 @@ typedef enum
   POCL_CDBI_GAUSSIAN3X3_P512 = 37,
   POCL_CDBI_DBK_KHR_GEMM = 38,
   POCL_CDBI_DBK_KHR_MATMUL = 39,
+  POCL_CDBI_DBK_KHR_JPEG_ENCODE = 40,
```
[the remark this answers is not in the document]
RABijl: oke, with old ones I assume gemm and matmul, so i'll do those as well.

linehill:
What is the format of the data coming out of the decode / going into the encode DBKs? E.g. Are the pixels RGB? Are they organized into planes? What's the order of the channels? Does the format depend on the JPEG format?

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,197 @@
...
+  status = tj3DecompressHeader (state->tj_handle, jpeg, *jpeg_size);
+  if (status != 0)
+    {
+      POCL_MSG_ERR ("tj3DecompressHeader failed: %s (fatal? %d)\n",
+                    tj3GetErrorStr (state->tj_handle),
+                    tj3GetErrorCode (state->tj_handle));
+      return CL_DEVICE_NOT_AVAILABLE;
+    }
+  status
+    = tj3Decompress8 (state->tj_handle, jpeg, *jpeg_size, image, 0, TJCS_RGB);
```
linehill, line 161: Is it checked somewhere that the allocation size of the `image` is large enough for this call?

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,197 @@
...
+  jpeg_encode_state_t *state = kernel->data[dev_i];
+
+  int status;
+  status = tj3DecompressHeader (state->tj_handle, jpeg, *jpeg_size);
+  if (status != 0)
+    {
+      POCL_MSG_ERR ("tj3DecompressHeader failed: %s (fatal? %d)\n",
+                    tj3GetErrorStr (state->tj_handle),
+                    tj3GetErrorCode (state->tj_handle));
+      return CL_DEVICE_NOT_AVAILABLE;
```
linehill, line 158: Seems inappropriate error code. OTOH, I don't know what would be better. Perhaps CL_OUT_OF_RESOURCES?

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
RABijl: That would be the responsibility of the user similar how kernel functions do not do out of bounds checking.

RABijl:
> What is the format of the data coming out of the decode / going into the encode DBKs? E.g. Are the pixels RGB? Are they organized into planes? What's the order of the channels? Does the format depend on the JPEG format?

Good point, i'll some documentation on it. But to answer your question, it's following the common layout of images you see in other frameworks say OpenCV (only in RGB this time), point (0,0) is top left, the pixels are 3 chars after each other in RGB order, to get to the next row, you need to offset the index by width*3. The output is an image encoded into a jpeg that contains the metadata to decompress it. The decoding kernel should not have any problem with decoding a jpeg not encoded by a dbk.

the test input is such a rgb image and you can have a look with image magick using `display -size 640x480 -depth 8 RGB:input.rgb`

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
linehill: Right, because it's non-trivial to do bounds checking on arbitrary user defined kernels. Here it should be trivial, I think. I thinks it's better to do bounds checking here rather than seqfault or cause other undefined behavior.

RABijl: How should this be conveyed to the user? errors return codes of this function will not be checked, so i'm thinking we print an error message and abort before any damage is done as this is clearly an error by the programmer. Another option would be to make a check when the kernel arg is set and catch the error early.

RABijl: now that I think about it, a third option would be to add the cl_mem objs to the attribute list and do the check during validation.

linehill: * option 1: LGTM if abort only occurs in the Debug mode. For release modes, the error message is enough, I think.
* option 2: This would require that the JPEG  header is parsed at `clSetKernelArg` call time to get the decoded size but the JPEG buffer might uninitialized at that point.
* option 3: The DBK would lose flexibility (can't switch buffers). Also, has the issue of (2).

include/CL/cl_exp_defined_builtin_kernels.h
```diff
@@ -90,7 +90,9 @@ typedef enum
   POCL_CDBI_GAUSSIAN3X3_P512 = 37,
   POCL_CDBI_DBK_EXP_GEMM = 38,
   POCL_CDBI_DBK_EXP_MATMUL = 39,
+  /* see portablecl.org for details on usage */
```
pjaaskel, line 93: Please add a user manual reference instead of a web site domain.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
RABijl: I've implemented option 1 with some bounds checking on both dbks that will abort on a debug build.

include/CL/cl_exp_defined_builtin_kernels.h
RABijl: changed ref to section name and file where it is described.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,217 @@
...
+      POCL_MSG_ERR ("tj3DecompressHeader failed: %s (fatal? %d)\n",
+                    tj3GetErrorStr (state->tj_handle),
+                    tj3GetErrorCode (state->tj_handle));
+      return CL_OUT_OF_RESOURCES;
+    }
+
+  int height = tj3Get (state->tj_handle, TJPARAM_JPEGHEIGHT);
+  int width = tj3Get (state->tj_handle, TJPARAM_JPEGWIDTH);
+  cl_mem jpeg_mem = *(cl_mem *)(arguments[2].value);
+  if (jpeg_mem->size < width * height * tjPixelSize[TJCS_RGB])
```
linehill, line 172:
```suggestion
  size_t height = tj3Get (state->tj_handle, TJPARAM_JPEGHEIGHT);
  size_t width = tj3Get (state->tj_handle, TJPARAM_JPEGWIDTH);
  cl_mem jpeg_mem = *(cl_mem *)(arguments[2].value);
  if (jpeg_mem->size < width * height * tjPixelSize[TJCS_RGB])
```
To avoid undefined behavior in `width * height * tjPixelSize` expression when the `width` and `height` variables are sufficiently large (e.g. `width == height == 65535`).

RABijl: it should be go fine since jpeg_mem->size is of size_t, but this does make it more clear what is happening

pjaaskel:
Can you squash this to a single commit?

doc/sphinx/source/features.rst
```diff
@@ -10,3 +10,4 @@ Supported OpenCL features, device drivers
    conformance
    extensions
    drivers
+   jpeg_dbk
```
pjaaskel, line 13: I don't think a single DBK warrants a whole section of its own in the doc. Please move this under the section which discusses DBKs/BiKs.

doc/sphinx/source/jpeg_dbk.rst
```diff
@@ -0,0 +1,25 @@
...
+refer to the `build instructions <https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/BUILDING.md>`_
+on how to build it. After running `make install`, the library will be installed in `/opt` . CMake might have some
+trouble finding this library, therefore add the following line to your CMake arguments when building PoCL:
+
+    -Dlibjpeg-turbo_DIR=/opt/libjpeg-turbo/lib64/cmake/libjpeg-turbo/
+
+Usage
+_____
+
+The jpeg dbks work on buffers of RGB images, similar to other image manipulation frameworks. The characteristics of the
```
pjaaskel, line 21: Check caps JPEG, DBKs, etc.

include/CL/cl_exp_defined_builtin_kernels.h
```diff
@@ -88,8 +88,14 @@ typedef enum
   POCL_CDBI_MAGNITUDE_P512 = 35,
   POCL_CDBI_ORIENTED_NONMAX_P512 = 36,
   POCL_CDBI_GAUSSIAN3X3_P512 = 37,
-  POCL_CDBI_DBK_KHR_GEMM = 38,
-  POCL_CDBI_DBK_KHR_MATMUL = 39,
+  POCL_CDBI_DBK_EXP_GEMM = 38,
+  POCL_CDBI_DBK_EXP_MATMUL = 39,
+  /* see 'JPEG Defined Builtin Kernels:Usage' in
```
pjaaskel, line 93:
```suggestion
  /* See 'JPEG Defined Builtin Kernels:Usage' in jpeg_dbk.rst for details on usage. */
```
Please use full sentences and more of the allowed 80 chars per line.

lib/CL/dbk/dbk_khr_jpeg_shared.c
```diff
@@ -0,0 +1,104 @@
+/* dbk_khr_jpeg_shared.c - generic jpeg defined builtin kernel functions
+
+   Copyright (c) 2024 PoCL developers
```
pjaaskel, line 3: Please use the names of authors and their affil at the time of code writing.

doc/sphinx/source/features.rst
pjaaskel: Also add a mention of this new (experimental) feature to the next release notes .rst

lib/CL/dbk/dbk_khr_jpeg_shared.h
```diff
@@ -0,0 +1,38 @@
+/* dbk_khr_jpeg_shared.h - generic jpeg defined builtin kernel functions
+
+   Copyright (c) 2024 PoCL developers
```
pjaaskel, line 3: same here

lib/CL/dbk/dbk_khr_jpeg_shared.h
```diff
@@ -0,0 +1,38 @@
+/* dbk_khr_jpeg_shared.h - generic jpeg defined builtin kernel functions
```
pjaaskel, line 1:
```suggestion
/* pocl_dbk_khr_jpeg_shared.h - generic jpeg defined builtin kernel functions
```

lib/CL/dbk/dbk_khr_jpeg_shared.h
```diff
@@ -0,0 +1,38 @@
...
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+   IN THE SOFTWARE.
+*/
+
+#ifndef _DBK_KHR_JPEG_SHARED_H_
```
pjaaskel, line 24:
```suggestion
#ifndef POCL_DBK_KHR_JPEG_SHARED_H_
```

lib/CL/devices/basic/basic.c
```diff
@@ -960,3 +956,102 @@ pocl_basic_set_kernel_exec_info_ext (cl_device_id dev,
...
+        {
+          dbk_index = i;
+          return dbk_index;
+        }
+    }
+  return dbk_index;
+}
+
+/**
+ * CPU_DBK_TAG
```
pjaaskel, line 977: ?

lib/CL/devices/basic/basic.c
```diff
@@ -960,3 +956,102 @@ pocl_basic_set_kernel_exec_info_ext (cl_device_id dev,
...
+ */
+int
+pocl_basic_create_kernel (cl_device_id device,
+                          cl_program p,
+                          cl_kernel k,
+                          unsigned device_i)
+{
+
+  /* no dbks, nothing to do */
```
pjaaskel, line 986: Extract this functionality to a separate helper method so we can call it from other device drivers which want to add JPEG DBK support.

lib/CL/devices/basic/basic.c
```diff
@@ -960,3 +956,102 @@ pocl_basic_set_kernel_exec_info_ext (cl_device_id dev,
...
+      }
+#endif
+    default:
+      POCL_ABORT ("pthread create kernel called with unknown/unimplemented "
+                  "dbk kernel\n");
+    }
+}
+
+/**
+ * CPU_DBK_TAG
```
pjaaskel, line 1020: ?

lib/CL/devices/common_utils.c
```diff
@@ -112,7 +112,9 @@ align_ptr (char *p)
 #define FALLBACK_MAX_THREAD_COUNT 8
 
 /* Initializes CPU-specific device info default, that cannot / should
-   not be initialized in pocl_init_default_device_infos() */
+   not be initialized in pocl_init_default_device_infos()
+   CPU_DBK_TAG
```
pjaaskel, line 116: ?

lib/CL/devices/common_utils.c
```diff
@@ -616,20 +630,37 @@ pocl_cpu_validate_khr_gemm (cl_bool TransA,
 }
 #endif
 
+/**
+ * CPU_DBK_TAG
```
pjaaskel, line 634: What are these?

lib/CL/devices/common_utils.c
```diff
@@ -616,20 +630,37 @@ pocl_cpu_validate_khr_gemm (cl_bool TransA,
...
-  return pocl_validate_dbk_attributes (kernel_id, kernel_attributes,
-                                       pocl_cpu_validate_khr_gemm);
-#else
-  POCL_RETURN_ERROR_ON (1, CL_UNSUPPORTED_DBK,
-                        "The CPU driver must be compiled with libxsmm "
-                        "to support tensor DBKs\n");
+    case POCL_CDBI_DBK_EXP_GEMM:
+    case POCL_CDBI_DBK_EXP_MATMUL:
+      {
+        /* the following code checks for LIBXSMM specific requirements on
```
pjaaskel, line 647: Please use a full sentence.

lib/CL/devices/common_utils.c
```diff
@@ -640,17 +671,40 @@ pocl_cpu_probe ()
...
+#endif
+  /* TODO: is it necessary to return an error here or can it be caught earlier
+     on? */
+  POCL_RETURN_ERROR_ON (
+    1, CL_BUILD_PROGRAM_FAILURE,
+    "The CPU driver has not been compiled with support for DBKs\n");
+}
+
+void *
+pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id)
```
pjaaskel, line 696: Missing comment.

lib/CL/devices/common_utils.c
```diff
@@ -830,27 +884,12 @@ pocl_cpu_execute_gemm_anytype (char *Aptr,
...
-  return (void *)ptr;
-}
-
 int
-pocl_cpu_execute_dbk (cl_program program,
-                      cl_kernel kernel,
-                      pocl_kernel_metadata_t *meta,
-                      cl_uint dev_i,
-                      struct pocl_argument *arguments)
+pocl_xsmm_execute_dbk (cl_program program,
```
pjaaskel, line 888: Please extract DBK utils to a separate module. common_utils.c is bloated.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,217 @@
+/* dbk_khr_jpeg_cpu.c - jpeg defined builtin kernel functions for cpu devices
+
+   Copyright (c) 2024 Robin Bijl/PoCL developers
```
pjaaskel, line 3: Your affil is currently Tampere University.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,217 @@
...
+
+#ifdef HAVE_LIBJPEG_TURBO
+
+#include <turbojpeg.h>
+
+typedef struct
+{
+  tjhandle tj_handle;
+  unsigned char *tj_buffer;
+  /* used to store the buffer size when the compression
```
pjaaskel, line 35: full sentences

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,217 @@
...
+  size_t jt_buffer_size;
+} jpeg_encode_state_t;
+
+typedef struct
+{
+  tjhandle tj_handle;
+} jpeg_decode_state_t;
+
+int
+pocl_cpu_execute_dbk_khr_jpeg_encode (cl_program program,
```
pjaaskel, line 46: Should this be named after it being turbojpeg-specific instead of "pocl cpu"? We could integrate it with other device drivers as well (even though the will utilize CPU in the processing).

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
```diff
@@ -0,0 +1,217 @@
...
+      *status = CL_OUT_OF_RESOURCES;
+      return NULL;
+    }
+  /* todo set jpeg no_realloc */
+
+  return state;
+}
+
+int
+pocl_cpu_destroy_dbk_khr_jpeg_encode (void **kernel_data)
```
pjaaskel, line 131: Yeah. The whole module should be named after turbojpeg since this is turbojpeg impl. of the JPEG DBK. Let's not involve the device driver names here as we can integrate these also to custom devices etc.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
```diff
@@ -0,0 +1,95 @@
+/* dbk_khr_jpeg_cpu.h - jpeg defined builtin kernel functions for cpu devices
+
+   Copyright (c) 2024 PoCL developers
```
pjaaskel, line 3: ...

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
```diff
@@ -0,0 +1,95 @@
...
+   IN THE SOFTWARE.
+*/
+
+#ifndef _DBK_KHR_JPEG_CPU_H_
+#define _DBK_KHR_JPEG_CPU_H_
+
+#include "pocl_cl.h"
+
+/**
+ * @brief Do jpeg compression from a rgb image in the first argument and store
```
pjaaskel, line 30:
```suggestion
 * JPEG encode an RGB image in the first argument and store
```

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
```diff
@@ -0,0 +1,95 @@
...
+ * the result and size in the second two arguments.
+ * @param program [in] used to retrieve device data
```
pjaaskel, line 32: add an empty line between the param list

pjaaskel: and should we use \param or @param throughout? I think \param as that's what LLVM prefers.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
```diff
@@ -0,0 +1,95 @@
...
+ */
+POCL_EXPORT int
+pocl_cpu_execute_dbk_khr_jpeg_encode (cl_program program,
+                                      cl_kernel kernel,
+                                      pocl_kernel_metadata_t *meta,
+                                      cl_uint dev_i,
+                                      struct pocl_argument *arguments);
+
+/**
+ * @brief Create and init the jpeg state, including a turbo jpeg handle.
```
pjaaskel, line 46: Drop the @brief. LLVM doesn't like them and for a good reason, they are just extra as Doxygen can parse the brief without (we have used without brief since the beginning in TCE/OpenASIP).

tests/runtime/test_dbk_jpeg.c
```diff
@@ -0,0 +1,227 @@
+/* test_dbk_jpeg.c - test program for jpeg defined builtin kernels
+
+   Copyright (c) 2024 PoCL developers
```
pjaaskel, line 3: ....

lib/CL/devices/cpu_dbk/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+
+if(HAVE_LIBJPEG_TURBO)
```
pjaaskel, line 2: @franz when you have a chance (not urgent), can you add turbojpeg to one of the CIs so we can test this automatically?

pjaaskel (changes requested):
Bit of cleanup/refactoring needed, but getting there.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
linehill: The expression is evaluated as `int` before and then promoted to `size_t`for the comparison. See: https://godbolt.org/z/a9Wx13Pcz.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
linehill: `\param` is used the most in LLVM.

doc/sphinx/source/features.rst
RABijl: There is no current section to discussing just DBKs/BiKs , only the almaif touches a bit on it. I'll make a new section.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
RABijl: I think it should depend on the file type, `/**` is Javadoc style and uses @param. Javadoc style fits pretty well with c style comments. `///` is c++ style so /param fits there. https://www.doxygen.nl/manual/docblocks.html

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.c
RABijl: I was following the trend that these dbks are only implemented for cpu devices like basic and pthread, Just like common_utils.c is intended only for cpu devices. I'd imagine it'd give some confusion if non-cpu devices where using turbojpeg in the backend

lib/CL/devices/basic/basic.c
RABijl: For now it's specific to cpu dbks. I'd imagine that other devices could have a different method of creating their kernels.

lib/CL/devices/common_utils.c
RABijl: I added these as an easy way to find all the code sections that need to be edited when adding more cpu dbks.

lib/CL/devices/common_utils.c
RABijl: The formatting ended up putting tensors on a new line, but I can try to make the original comment more clear.

pjaaskel (changes requested):

doc/sphinx/source/extensions.rst
```diff
@@ -37,3 +37,16 @@ commands are prepared at command buffer recording
 time, reducing the overhead of dispatching the sequence
 and allowing drivers to optimize the scheduling of
 commands within a buffer.
+
+cl_khr_defined_builtin_kernels
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The purpose of this extension is to provide a standardized
```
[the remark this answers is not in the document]
RABijl: This chapter describes the different pocl extensions and there is a different chapter describing dbks, so I'm not sure if it makes sense to do that.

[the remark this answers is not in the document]
pjaaskel: OK, then this belongs to the DBK chapter as this _is_ the DBK (an extension making a set of BiKs well-defined).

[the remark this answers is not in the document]
RABijl: This section is just giving a brief overview of DBKs and links to the DBK chapter that mentions the same thing (although in different words). I believe it belongs here as well as this chapter is dedicated to giving an overview of all the extensions PoCL supports. If you disagree, I can always replace the text with a link to the DBK chapter.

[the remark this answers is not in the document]
pjaaskel: Yes, a link is better than repetition.

lib/CL/devices/cpu_dbk/dbk_khr_jpeg_cpu.h
pjaaskel: Yep, best use \param throughout since it's LLVM's preference and no point diverging from that for C.

PR #1563, opened 2024-09-20
proxy: update to new organization of mem_objs

RABijl:
Do note that i've only tested this on x86 with example0 and not yet on android

This commit updates the proxy device to get
its mem_objs from the node instead of the event,
as is now done in all other devices.
There is also a fix in one of the tests for
when pocl is renamed.


pjaaskel:
@franz I think proxy device is not tested by any CI right now, is it so?

franz:
@pjaaskel  yes that is correct

PR #1564, opened 2024-09-23
convert printf to use (modified) LLVM emitPrintfCall()

franz:


franz:
There are still two unsolved issues (TCE is broken and for Mac OS disabled failing `spirv/printf` test) but it's ready for review

lib/llvmopencl/EmitPrintf.cc
```diff
@@ -0,0 +1,679 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Utility function to lower a printf call into a series of device
+// library calls on the AMDGPU target.
+//
+// WARNING: This file knows about certain library functions. It recognizes them
+// by name, and hardwires knowledge of their semantics.
+//
```
pjaaskel, line 14: Please add a note that this is copied from LLVM and should be retained in sync over the versions with upstreaming planned. We likely have to exclude it from clang-format too if the LLVM formatting doesn't much ours. In fact we should fix our clang-format script in that case.

pjaaskel: We should keep it nicely diffable to AMDGPUEmitPrintf.cpp with the diff minimized over time.

lib/llvmopencl/linker.cpp
```diff
@@ -360,14 +329,70 @@ static void shared_copy(llvm::Module *program, const llvm::Module *lib,
     program->eraseNamedMetadata(DebugCU);
 }
 
+static void handleDeviceSidePrintf(
```
pjaaskel, line 332: Can you add a function comment that briefly describes how the handling is done?

pjaaskel (approved):
Looks good. Small things that can be done post-pull.

lib/llvmopencl/linker.cpp
```diff
@@ -329,21 +329,47 @@ static void shared_copy(llvm::Module *program, const llvm::Module *lib,
...
+  } else {
+    PFlags.StorePtrInsteadOfMD5 = false;
+    PFlags.AlwaysStoreFmtPtr = true;
+  }
+
+  PFlags.ArgPromotionCharShort = true;
+  if (ClDev->double_fp_config == 0) {
+    PFlags.ArgPromotionFloat = false;
+  }
+  if (ClDev->endian_little == CL_FALSE) {
```
pjaaskel, line 367: No need to use the CL macros, right? Just !

lib/CL/pocl_cl.h
```diff
@@ -987,6 +987,10 @@ typedef struct pocl_global_mem_t {
 
 #define NUM_OPENCL_IMAGE_TYPES 6
 
+/* typedef for intrinsics replacement function. TODO this is a hack,
```
pjaaskel, line 990: I wonder should we call this a temporary hack or just accept it as a new feature that could be useful for other targets for converting intrinsics to lib functions without updating the LLVM backend. Logically, these cross-AS-memcpys are for OpenCL and PoCL is an OpenCL implementation so it should provide the implementations (for the targets). I'd just make it an "official" device hook for now and document it properly.

franz: > accept it as a new feature that could be useful for other targets

It might be useful, but it's still a hack to fix deficiencies in the backend(s). And it is not maintenance free, because right now memcpy cannot be implemented in C source (because of the i1 argument) only in '.ll' and the format of that is LLVM version dependent.

> Logically, these cross-AS-memcpys are for OpenCL and PoCL is an OpenCL implementation so it should provide the implementations (for the targets).

From what i read, Clang and GCC both can emit calls to functions like memcpy independently of the source language, and it is the responsibility of the runtime to provide the implementations for memset, memcpy, memmove and memcmp. And if these functions are to be efficient, they need target specific knowledge. So IMO this does not belong into PoCL and should not be made "official".

pjaaskel: OK, let's generalize when there's another target needing this.

> From what i read, Clang and GCC both can emit calls to functions like memcpy independently of the source language, and it is the responsibility of the runtime to provide the implementations for memset, memcpy, memmove and memcmp. And if these functions are to be efficient, they need target specific knowledge. So IMO this does not belong into PoCL and should not be made "official".

For "target specific implementations" we have the target-specific built-in library override mechanism. Sure, we can push it to a separate builtin library, but it doesn't fit to libc due to libc being a C library (address space 0 memcpys etc. belong there), then the question is what is the correct library/place for these if not in an OpenCL (compiler) implementation of which multi-AS compiler is producing them? It's not an entirely clear answer to me. I wonder where the other targets put them or just expand in the backend to IR? CPUs just flatten out the AS so they can use the normal memcpy.

franz: > I wonder where the other targets put them or just expand in the backend to IR?

AFAICT, you can't put them anywhere. LLVM refuses to compile an .ll file which actually defines (not just declares) an intrinsic. So it seems to me the intention is that the backend is supposed to expand the intrinsic.

PR #1567, opened 2024-09-25
Initial onnx dbk implementation

RABijl:
@jansol's ONNX DBK rebased on top of the main.


doc/sphinx/source/dbk.rst
```diff
@@ -34,3 +34,63 @@ buffer are as follows: point (0,0) is the top left of the image and each pixel c
 padding. So to get a pixel value from a certain row, add an offset of row_index * width * 3 to the index. The fourcc
 code that comes closest to this is the BI_RGB 24bpp format. Check the `include/CL/cl_exp_defined_builtin_kernels.h` for
 info on DBK attributes and `tests/runtime/test_dbk_jpeg.c` for an example program making use of the JPEG DBKs.
+
+
+exp_onnx_inference
+__________________
+
+This experimental extension proposes adding a builtin kernel that ingests an
```
pjaaskel, line 42: The idea of the DBK extension is that it gets updated with new versions over time, without needing to add a new extension for each DBK. Let's see how well that works and not add a new extension but just a new DBK to the DBK extension. Thus the title here should be ONNX Inference DBK or such.

doc/sphinx/source/dbk.rst
```diff
@@ -34,3 +34,63 @@ buffer are as follows: point (0,0) is the top left of the image and each pixel c
...
+ONNX file and performs inference of the contained network.
```
pjaaskel, line 43:
```suggestion
ONNX file and performs inference with the contained network.
```

doc/sphinx/source/dbk.rst
```diff
@@ -34,3 +34,63 @@ buffer are as follows: point (0,0) is the top left of the image and each pixel c
...
+
+Building with ONNX kernel support
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to support this builtin kernel, PoCL needs to be built with ONNX Runtime
```
pjaaskel, line 48:
```suggestion
In order to support this builtin kernel through the CPU devices, PoCL needs to be built with ONNX Runtime
```

include/CL/cl_exp_defined_builtin_kernels.h
```diff
@@ -209,4 +210,23 @@ typedef struct
   cl_int quality;
 } cl_dbk_attributes_exp_jpeg_encode;
 
+/* Name: "exp_onnx_inference"
```
pjaaskel, line 213:
```suggestion
/* Name: "onnx_inference"
```
No need to have the 'exp' / khr / etx prefix if we go with the "all DBKs in the same extension" approach.

lib/CL/clSetKernelArg.c
```diff
@@ -155,7 +155,35 @@ pocl_verify_dbk_kernel_args (cl_mem buf,
     case POCL_CDBI_DBK_EXP_JPEG_ENCODE:
```
pjaaskel, line 155: Can we move the DBK handling to a DBK utility module? Otherwise it will bloat the clSetKernelArg.c.

lib/CL/clSetKernelArg.c
```diff
@@ -155,7 +155,35 @@ pocl_verify_dbk_kernel_args (cl_mem buf,
...
+        const cl_dbk_attributes_khr_onnx_inference *attrs
+            = (const cl_dbk_attributes_khr_onnx_inference *)
+                meta->builtin_kernel_attrs;
+        if (arg_index == 0
+            && buf->size < attrs->num_inputs * sizeof (uint64_t))
+          return CL_INVALID_ARG_VALUE;
+        size_t total_input_size = 0;
+        for (size_t i = 0; i < attrs->num_inputs; ++i)
+          {
+            // TODO: sum up the sizes of all input tensors
```
pjaaskel, line 169: Hm? Why it's not done? Unfinished implementation?

lib/CL/clSetKernelArg.c
```diff
@@ -155,7 +155,35 @@ pocl_verify_dbk_kernel_args (cl_mem buf,
...
+          }
+        if (arg_index == 1 && buf->size < total_input_size * sizeof (uint64_t))
+          return CL_INVALID_ARG_VALUE;
+        if (arg_index == 2
+            && buf->size < attrs->num_outputs * sizeof (uint64_t))
+          return CL_INVALID_ARG_VALUE;
+        size_t total_output_size = 0;
+        for (size_t i = 0; i < attrs->num_outputs; ++i)
+          {
+            // TODO: sum up the sizes of all output tensors
```
pjaaskel, line 179: ...

lib/CL/pocl_builtin_kernels.c
```diff
@@ -832,10 +840,31 @@ pocl_validate_dbk_attributes (BuiltinKernelId kernel_id,
     case POCL_CDBI_DBK_EXP_JPEG_DECODE:
     case POCL_CDBI_DBK_EXP_JPEG_ENCODE:
       return pocl_validate_khr_jpeg (kernel_id, kernel_attributes);
-    default:
+#ifdef HAVE_ONNXRT
+    case POCL_CDBI_DBK_EXP_ONNX_INFERENCE:
+      {
```
pjaaskel, line 845: Can we move the arg validation also to this file?  Also please don't use C++ comments in C.

lib/CL/pocl_onnxrt.c
```diff
@@ -0,0 +1,511 @@
...
+  *onnxrt = NULL;
+  free (ort);
+
+ERROR:
+  return CL_SUCCESS;
+}
+
+/** Run the model loaded into the given onnxrt session
+ *
+ * @param ort An ONNXRuntime inference session initialized with
```
pjaaskel, line 259:
```suggestion
 * \param ort An ONNXRuntime inference session initialized with
```

tools/scripts/generate_minimal_onnx.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
```
pjaaskel, line 1: Please add a license/copyright notice.

include/CL/cl_exp_defined_builtin_kernels.h
RABijl: That could be done in the future, but rn the pattern has been to set the name to the DBK string name and it could cause some confusion to deviate only on the onnx one.

lib/CL/clSetKernelArg.c
RABijl: That could be done after this pr, it might be easy to move to another existing file.

lib/CL/pocl_builtin_kernels.c
RABijl: should be possible, there is already a validate attributes function

include/CL/cl_exp_defined_builtin_kernels.h
```diff
@@ -209,4 +210,23 @@ typedef struct
...
+ * Attributes for constructing an inference session for ONNX format ML models.
+ */
+typedef struct _cl_dbk_attributes_exp_onnx_inference
+{
+  char *model_data;
```
linehill, line 218: All pointer, in the DBK attribute structures, should be `const` qualified. 
```suggestion
  const char *model_data;
```

RABijl: That would make things more cumbersome to use, I've opted instead to make the instances of the attributes const. Maybe in the future the program->builtin_kernel_attributes could be made const.

jansol: We should be able to treat all attribute fields as immutable at all times I think? The only exception would be of course creating and deleting the struct. But that can be tucked away in helper functions that use separate local variables to assemble the fields and only write the final data in the struct when they have everything ready. And for zeroing out the struct upon deletion we can sidestep constness entirely with memset.

RABijl: yes they should be fine treated as immutable, the original attributes passed over to `clCreateFunctionWithDefinedBuiltinKernels` are deep copied. It's just annoying for an end user to make sure all the strings and data is const (cast) when creating the attribute instance.

linehill: I'm doubtful that Khronos will accept structures with non-`const` pointers in them. See `cl_mutable_*` structures in the OpenCL API spec - all of the pointer members are `const` qualified. 

However, there is at least one exception to this pattern: `cl_dx9_surface_info_khr` structure.

RABijl: I see, what should be done with the double pointer cases? Should there be a set limit and use a const array, or something along the lines of `const char * const *` ?

linehill: I think `const TYPE **` form is enough. I don't see any `const TYPE * const *`s in the core API spec.

RABijl: also, I noticed that `cl_tensor_desc.layout` is of type `void *` shouldn't that also be a `const` then?

linehill: Yeah, that probably should be changed too. I'll do that when I get time to.

RABijl: oke the code has been changed to use const for pointers. I also found out that clReleaseProgram didn't free kernel attributes, so i've made a separate commit for that.

RABijl: @linehill for review, besides the struct, this is the file that has been changed with the `const` changes: https://github.com/pocl/pocl/pull/1567/files#diff-cebcbd0d759fb1fd119d11af3678805dc635025332ee1c2e0860a3fad5a99c28

tests/runtime/test_dbk_onnx_inference.c
```diff
@@ -0,0 +1,218 @@
...
+    = clCreateBuffer (context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR,
+                      sizeof (uint64_t) * 1, output_offsets, &error);
+  CHECK_CL_ERROR (error);
+
+  CHECK_CL_ERROR (
+    clSetKernelArg (kernel, 0, sizeof (cl_mem), &input_offsets_buf));
+  CHECK_CL_ERROR (clSetKernelArg (kernel, 1, sizeof (cl_mem), &all_inputs));
+  CHECK_CL_ERROR (
+    clSetKernelArg (kernel, 2, sizeof (cl_mem), &output_offsets_buf));
+  CHECK_CL_ERROR (clSetKernelArg (kernel, 3, sizeof (cl_mem), &all_outputs));
```
linehill, line 179: Since ONNX model takes inputs and outputs as tensors, shouldn't the ONNX kernel be interfaced with tensors too?
Can't we have the `clCreateProgramWithDefinedBuiltInKernels` to create a kernel with N + M kernel (tensor) parameters based on N inputs and M outputs in the DBK attribute structure?

jansol: That was my original idea, but the conclusion on that was that that would have too much of a risk of running into unfeasible numbers of kernel arguments. Trivially so, if we also want to override initializers at that level.

linehill: Is the concern here that the `CL_DEVICE_MAX_PARAMETER_SIZE` could be exceeded on some models?

I agree there potentially could be unfeasible number of kernel arguments if the initializers would be exposed as arguments too but I'm not sure about that with the I/O tensors alone as kernel arguments. Are there such models that have more than ten or even hundred I/O tensors?

doc/sphinx/source/dbk.rst
```diff
@@ -34,3 +34,58 @@ buffer are as follows: point (0,0) is the top left of the image and each pixel c
...
+
+Some popular Linux distributions don't provide prebuilt packages of ONNX Runtime
+so you may have to
+`build your own <https://onnxruntime.ai/docs/build/inferencing.html>` and tell
+CMake where to find it. Steps for building a suitable copy of ONNX Runtime (run
+outside of the PoCL source tree) are as follows:
+
+    git clone --recursive https://github.com/microsoft/onnxruntime/
+    cd onnxruntime
+    ./build.sh --config Release --build_shared_lib --parallel --compile_no_warning_as_error --skip_submodule_sync --skip_tests --cmake_extra_defines CMAKE_INSTALL_PREFIX=${HOME}/onnxruntime-bin
```
linehill, line 71: What does `--skip_submodule_sync` do?

jansol: It's part of the recommended flags from the build instructions on the onnxruntime website. `./build.sh --help` clarifies that:
```
  --skip_submodule_sync
                        Don't do a 'git submodule update'. Makes the Update phase faster.
```

Which makes sense after cloning the repo with `--recursive` right on the line above (as that clones and updates all submodules). Dropping the `--recursive` flag from the clone and the `--skip_submodule_sync` from the build flags is also an option.

tests/runtime/test_dbk_onnx_inference.c
jansol: No idea, I haven't looked at a lot of models. DeepFilterNet consists of 3 networks with the most verbose one having 9 inputs/outputs. I don't think `CL_DEVICE_MAX_PARAMETER_SIZE` would be the problem since the parameters are just `cl_mem` in the end.

lib/CL/devices/remote/communication.c
```diff
@@ -2362,13 +2362,26 @@ pocl_network_setup_peer_mesh ()
...
  */
 cl_int
-pocl_network_build_or_link_program (
-    remote_device_data_t *ddata, const void *payload, size_t payload_size,
-    int is_binary, int is_builtin, int is_spirv, uint32_t prog_id,
-    const char *options, char **kernel_meta_bytes, size_t *kernel_meta_size,
-    uint32_t *devices, uint32_t *platforms, size_t num_devices,
-    char **build_logs, char **binaries, size_t *binary_sizes,
-    size_t svm_region_offset, int compile_only, int link_only)
+pocl_network_build_or_link_program (remote_device_data_t *ddata,
```
pjaaskel, line 2365: Is this really the format clang-format prefers here for the long argument list?

lib/CL/devices/remote/communication.h
```diff
@@ -330,13 +330,26 @@ cl_int pocl_network_setup_metadata (char *buffer, size_t total_size,
...
                                     pocl_kernel_metadata_t **kernel_meta);
 
-cl_int pocl_network_build_or_link_program (
-    remote_device_data_t *ddata, const void *payload, size_t payload_size,
-    int is_binary, int is_builtin, int is_spirv, uint32_t prog_id,
-    const char *options, char **kernel_meta_bytes, size_t *kernel_meta_size,
-    uint32_t *devices, uint32_t *platforms, size_t num_devices,
-    char **build_log, char **binaries, size_t *binary_sizes,
-    size_t svm_region_offset, int compile_only, int link_only);
+cl_int pocl_network_build_or_link_program (remote_device_data_t *ddata,
```
pjaaskel, line 333: same

lib/CL/devices/remote/remote.c
```diff
@@ -1214,6 +1294,33 @@ pocl_remote_supports_binary (cl_device_id device, size_t length,
   return 0;
 }
 
+extern pocl_kernel_metadata_t pocl_BIDescriptors[BIKERNELS];
```
pjaaskel, line 1297: check the case of the variable name

poclu/misc.c
```diff
@@ -673,3 +674,15 @@ poclu_load_program (cl_context context, cl_device_id device,
                                       spirv, poclbin, explicit_binary,
                                       extra_build_opts, p);
 }
+
+cl_half
+poclu_float_to_cl_half (float value)
```
pjaaskel, line 679: This might be problematic. poclu was originally meant to be an OpenCL-implementation independent utility library for tests etc. thus it should not call pocl internal utilities directly as those are not made available in the produced libOpenCL.so/pocl, especially if one links via ICD. If we don't need the halfs in the tests, I think we can remove it from poclu and leave it to pocl_util or even duplicate the code for now via including .c. I cannot think of a better solution quickly.

pjaaskel (changes requested):

lib/CL/devices/remote/communication.c
RABijl: I ran `format-last-commit.sh` again just to be sure and that what it did.

poclu/misc.c
RABijl: I see I'll have to rebase commit out at some point. But in any case, poclu is built with the pocl_cl_half_util source file, so it does not have a dependency on pocl. And while I was searching for the file, I found a third [cl_half](https://github.com/pocl/pocl/blob/main/include/CL/cl_half.h) file from khronos, maybe that one should replace the misc and internal cl_half conversion functions.

RABijl:
I'm putting this on ready for review as it allows running onnx dbks with the remote device and it is already almost 2k lines changed. Any API changes could be added later.

pjaaskel (approved):
I believe we can merge this in and improve later as we get experiences of using it. Please rebase and retest.

RABijl:
oke it has been rebased on top of master and I tested both CPU and remote devices with the test program.

pjaaskel:
https://github.com/pocl/pocl/actions/runs/11503734600/job/32021663424?pr=1567#step:5:427

PR #1575, opened 2024-09-30
misc: fix opencl version un-/re-definition warnings

RABijl:
pocld is now built with the
CL_TARGET_OPENCL_VERSION and
CL_HPP_TARGET_OPENCL_VERSION definitions set.
For regression tests, the version definition
is unset and set the to the infile defined one.


pjaaskel:
I wonder why we force set that version in the tests. @franz do you recall?

franz:
@pjaaskel I think it was originally set in each test separately, then i added a -D definition on the command line that is (almost) project-wide. BTW we have warned in 6.0 release notes of deprecating building with pre-3.0 headers; so the better fix would be to remove all of the pre-3.0 stuff like these macros and CMake checks, then check that everything still builds with 3.0 headers, and force 3.0 project-wide.

RABijl:
Oke, I removed those checks, except in the khronos opencl headers, and pocl.h. The khronos headers might need to be updated, see this [issue](https://github.com/KhronosGroup/OpenCL-CLHPP/issues/108) that recently got fixed. I tested with tier1 and only the pyopencl ones failed, due to an import error for some reason. So now i'm putting it through the build bot and squashing the commits later.

jansol (approved):

franz:
The OpenASIP CI was failing because the CI had an outdated ocl-icd (pre 3.0), this should be fixed now. The other failures are unrelated to this PR (oneAPI should be fixed in main, chipstar probably failed b/c machine was overloaded today).

RABijl:
@franz I just realized that I should have squashed those two commits, is that still possible?

pjaaskel:
@RABijl nope, it's in already. Don't worry.

PR #1578, opened 2024-09-30
Fixes and polish

pjaaskel:


jansol (approved):

jansol:
Hmm, those changes shouldn't really upset CUDA and OpenASIP like that.

PR #1579, opened 2024-10-01
CPU: Fix missing isinf(half) and isinf(halfn)

### LINKED ISSUE #1562: undefined symbol: _Z9_cl_isinfDv8_Dh'
Encountered on OpenVINO running an inference an YOLO model.

Reproducer:
```c++
#define CL_HPP_ENABLE_EXCEPTIONS
#include "CL/opencl.hpp"
#include <iostream>
#include <string>

int main() try {
  unsigned PlatformIdx = 0;
  unsigned DeviceIdx = 0;

  std::vector<cl::Platform> Platforms;
  cl::Platform::get(&Platforms);
  cl::Platform Platform = Platforms.at(PlatformIdx);

  std::vector<cl::Device> Devices;
  Platform.getDevices(CL_DEVICE_TYPE_ALL, &Devices);
  cl::Device Device = Devices.at(DeviceIdx);

  auto Source = std::string(R"OCL(
#pragma OPENCL EXTENSION cl_khr_fp16 : enable
global half8 Src = (half8)(1.0h);
global short8 Dst;
kernel void k() { Dst = isinf(Src); }
)OCL");

  auto Program = cl::Program(Source);
  try {
    Program.build("-cl-std=CL3.0");
  } catch (cl::Error &Ex) {
    std::string Log = Program.getBuildInfo<CL_PROGRAM_BUILD_LOG>(Device);
    std::cout << "Error building a program. Build log:\n" << Log << std::endl;
    return 1;
  }

  cl::Kernel K(Program, "k");
  auto Q = cl::CommandQueue::getDefault();
  Q.enqueueNDRangeKernel(K, cl::NullRange, cl::NDRange(1));
  Q.finish();

  std::cout << "Passed!" << std::endl;
  return 0;
} catch (cl::Error &Ex) {
  std::cerr << "Caught OpenCL exception: " << Ex.what() << "\n";
  std::cerr << "Error code: " << Ex.err() << "\n";
  return 2;
} catch (std::exception &Ex) {
  std::cerr << "Caught STL exception: " << Ex.what() << "\n";
  return 2;
}
```
Produces:
```
dlopen("/home/linehill/.cache/pocl/kcache/AI/BHKAOJFKKLHMGBPCGPEPJEEBHAGDJEEOBFBBN/k/1-1-1-goffs0-smallgrid/k.so") failed with '/home/linehill/.cache/pocl/kcache/AI/BHKAOJFKKLHMGBPCGPEPJEEBHAGDJEEOBFBBN/k/1-1-1-goffs0-smallgrid/k.so: undefined symbol: _Z9_cl_isinfDv8_Dh'.
note: missing symbols in the kernel binary might be reported as 'file not found' errors.
```

linehill:
They were missing because `isinf()` definitions were pulled from `lib/kernel/libclc*` which doesn't have overloads for `half` scalar and vectors.

Resolves #1562.

### DIFF
--- a/lib/kernel/host/CMakeLists.txt
+++ b/lib/kernel/host/CMakeLists.txt
@@ -342,6 +342,10 @@ else()
   set(KERNEL_SOURCES ${SOURCES_GENERIC})
 endif()
 
+if(HOST_CPU_ENABLE_CL_KHR_FP16)
+  list(APPEND KERNEL_SOURCES isinf_half.cl)
+endif()
+
 if(HOST_DEVICE_CL_VERSION_MAJOR GREATER_EQUAL 2)
 if(MIPS)
   message(STATUS "OpenCL 2.0 atomics are currently broken on MIPS")
--- a/lib/kernel/host/CMakeLists.txt
+++ b/lib/kernel/host/CMakeLists.txt
@@ -289,7 +289,8 @@ if(HOST_CPU_ENABLE_CL_KHR_FP16)
     round.cl
     sin.cl
     sqrt.cl
-    trunc.cl)
+    trunc.cl
+    isinf_half.cl)
 endif()
 
 
@@ -340,10 +341,6 @@ if(ENABLE_SLEEF)
   set(KERNEL_SOURCES ${SOURCES_WITH_SLEEF})
 else()
   set(KERNEL_SOURCES ${SOURCES_GENERIC})
-endif()
-
-if(HOST_CPU_ENABLE_CL_KHR_FP16)
-  list(APPEND KERNEL_SOURCES isinf_half.cl)
 endif()
 
 if(HOST_DEVICE_CL_VERSION_MAJOR GREATER_EQUAL 2)
--- a/lib/kernel/isinf_half.cl
+++ b/lib/kernel/isinf_half.cl
@@ -1,0 +1,34 @@
+/* OpenCL built-in library: isinf()
+
+   Copyright (c) 2024 Henry Linjamäki / Intel Finland Oy
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to deal
+   in the Software without restriction, including without limitation the rights
+   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+   copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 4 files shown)


lib/kernel/host/CMakeLists.txt
```diff
@@ -342,6 +342,10 @@ else()
   set(KERNEL_SOURCES ${SOURCES_GENERIC})
 endif()
 
+if(HOST_CPU_ENABLE_CL_KHR_FP16)
```
franz, line 345: There is already an existing if condition on line 245, please use that one. Also, add the removal of libclc isinf in the if block.

lib/kernel/isinf_half.cl
```diff
@@ -0,0 +1,34 @@
...
+   THE SOFTWARE.
+*/
+
+#include "templates.h"
+
+int __attribute__((overloadable)) isinf(half a) {
+  return __builtin_isinf(a);
+}
+
+IMPLEMENT_BUILTIN_L_V(isinf, short2, half2, half, lo, hi)
```
franz, line 30: Is it necessary to add a new file only for halfs ? The existing macros in `lib/kernel/isinf.cl` should already automatically define the half variants of isinf when it's available, so i think we should just use that file (and fix it if necessary).

franz (changes requested):

lib/kernel/isinf_half.cl
linehill: Question is that is the libclc's isinf() implementation preferred over `lib/kernel/isinf.cl` or could/should the libclc's implementation be discarded?

linehill: What is the upstream? The libclc files hosted in PoCL doesn't look like they are from llvm-project/libclc at all.

franz: > look like they are from llvm-project/libclc

They are from libclc (sort of). IIRC files in `lib/kernel/libclc-pocl` are wrappers (for compiling the same sources with different vector width) and the sources in `lib/kernel/libclc` are originally from libclc adapted to PoCL. They are not upstreamable, PoCL is the upstream.

> Question is that is the libclc's isinf() implementation preferred

No, it's not preferred. The PoCL's libclc sources do not have half support, because IIRC half support was almost nonexistent in both hardware and software at the time this was adapted from libclc (2017 i think).

pjaaskel: OK, maybe it's so old fork that it has diverged heavily upstream.

lib/kernel/host/CMakeLists.txt
linehill: Relocated. libclc's isinf is preserved and the `isinf_half.cl` complements it by adding implementation for the half overload. Agreed it's non-ideal, it's just for quick fix until we decide how we go forward with the kernel builtins.

pjaaskel:
@franz if this looks ok to you, please ack the review and pull it in.

pjaaskel (approved):

PR #1599, opened 2024-10-07
Allow using LLVM Support lib for dynlibs

pjaaskel:
This helps portability to non-POSIX platforms.

### DIFF
--- a/config.h.in.cmake
+++ b/config.h.in.cmake
@@ -203,6 +203,7 @@
 
 #cmakedefine LLVM_BUILD_MODE_DEBUG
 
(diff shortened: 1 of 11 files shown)


config.h.in.cmake
```diff
@@ -203,6 +203,7 @@
 
 #cmakedefine LLVM_BUILD_MODE_DEBUG
 
+#cmakedefine USE_LLVM_FOR_DYNLIBS
```
franz, line 206: I think this is not necessary since this variable is only used in CMake. I'll remove it in the PR with the file/thread support.

franz (approved):
LGTM, one small nitpick

PR #1603, opened 2024-10-08
remove cl_half utils from poclu

RABijl:
@franz I thought about about using the `cl_half.h` library for pocl's internal functions, but it was not clear to me how the current implementation handles an under-/over-flow errors. Do you have any opinions?

`cl_half.h` provides these options:
```
typedef enum
{
  CL_HALF_RTE, // round to nearest even
  CL_HALF_RTZ, // round towards zero
  CL_HALF_RTP, // round towards positive infinity
  CL_HALF_RTN, // round towards negative infinity
} cl_half_rounding_mode;
```

### DIFF
--- a/lib/CL/pocl_cl_half_util.c
+++ b/lib/CL/pocl_cl_half_util.c
@@ -23,6 +23,12 @@
    IN THE SOFTWARE.
 */
 
+/**
+ * TODO: see if this library can be dropped for the include/CL/cl_half.h
+ *  library. The question is which rounding mode should be used in that
+ *  case.
+ */
+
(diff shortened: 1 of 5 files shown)


franz:
> not clear to me how the current implementation handles an under-/over-flow errors

Sorry I don't really remember, this was ages ago. But AFAICT it's only used in that one places, and the usage is just verifying the code, so it should be OK to remove.

RABijl:
Putting a note on if `pocl_cl_half_utils.c` should be removed and marking this pr ready for review.

franz (approved):
LGTM

PR #1604, opened 2024-10-09
Fix couldn't find libjpeg-turbo package built from sources

linehill:
... due to different package name (libturbojpeg).

Also, run the `test_dbk_jpeg` test.


franz (approved):
LGTM

PR #1607, opened 2024-10-09
Major WILoops cleanup via removing cond barrier region peeling and explicit unrolling

pjaaskel:
Handle cond barriers with CBS. Do not handle kernels with non-uniform predicates with WILoops (for now). These used to be handled by "peeling" the first iteration of the work-item loop to check if the barrier is taken or not, but it complicates the parallel region formation and control flow construction quite a bit without known benchmark benefits. The produced peeled loops are also not easily vectorizable, thus the performance won't be good anyhow for SIMD.

For VLIW/ILP the peeling might be useful though, but it still doesn't justify the complexity add.

Remove peeling as well as explicit unrolling code to simplify the WILoops handling a lot.

Also always add an clean entry node to the parallel region to add context restores to, which simplifies PR handling in various places.

### DIFF
--- a/doc/sphinx/source/notes_6_1.rst
+++ b/doc/sphinx/source/notes_6_1.rst
@@ -72,4 +72,7 @@ Deprecation/feature removal notices
    method was removed to clean up the kernel compiler. It did not
    anymore have any use cases that could not be covered by fully
    unrolling "loops".
-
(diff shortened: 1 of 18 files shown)


pjaaskel:
Improved this in a private branch. Will push a new PR later.

PR #1611, opened 2024-10-11
Proxy: add SPIR-V support

RABijl:
This commit adds the required infrastructure
needed to use clCreateProgramWithIl with a Proxy
device. It makes use of the SPIR-V parser for
metadata, so it can be used on platforms that do
not provide kernel arg metadata but do support
SPIR-V like the Android Mali GPU OpenCL libraries. There has been a conversion of proxy.c->proxy.cpp
in order to use the parser and the mapping of
the parsed spir-v to pocl_kernel_metadata_t
from the level0 device has been moved and is now
shared with the proxy.


lib/CL/devices/spirv_parser.hh
```diff
@@ -100,6 +101,26 @@ POCL_EXPORT
 void applyAtomicCmpXchgWorkaround(const int32_t *InStream, size_t NumWords,
                                   std::vector<uint8_t> &OutStream);
 
+/// Method to map OCLFuncInfo objects to PoCL's kernel metadata struct.
```
pjaaskel, line 104:
```suggestion
/// Maps OCLFuncInfo objects to PoCL's kernel metadata struct.
```

lib/CL/pocl_image_util.h
```diff
@@ -30,47 +30,60 @@
 #pragma GCC visibility push(hidden)
 #endif
 
-void origin_to_bytes (cl_mem mem, const size_t *origin, size_t *byte_offset);
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  void origin_to_bytes (cl_mem mem, const size_t *origin, size_t *byte_offset);
```
pjaaskel, line 38: This is an annoying flaw in the clang format. We should not indent due to the extern "C" nor namespace brackets. Please fix by hand unless you find a way to teach clang-format to not do this.

pjaaskel (changes requested):
A bit of polishing needed, but LGTM.

lib/CL/devices/proxy/pocl_proxy.cpp
```diff
@@ -1164,24 +1197,42 @@ pocl_proxy_build_binary (cl_program program, cl_uint device_i,
...
+      pocl_cache_create_program_cachedir(program, device_i, program->program_il,
+                                         program->program_il_size, spirv_path);
+    }
+  } else {
+    // TODO should binary be already loaded ?
+    assert(program->pocl_binaries[device_i]);
+    char program_bc_path[POCL_MAX_PATHNAME_LENGTH];
+    pocl_cache_program_bc_path(program_bc_path, program, device_i);
+
+    assert(pocl_exists(program_bc_path));
```
franz, line 1226: Preferably, do not use asserts for error checking. Return an error (with an error message) and let the caller deal with it.

lib/CL/devices/proxy/pocl_proxy.cpp
```diff
@@ -1277,55 +1328,119 @@ pocl_proxy_setup_metadata (cl_device_id device, cl_program program,
...
+  err = clGetDeviceInfo(d->device_id, CL_DEVICE_EXTENSIONS, 0, NULL,
+                        &extension_length);
+  if (err != CL_SUCCESS) {
     return 0;
+  }
+  char *extensions = (char *)malloc(extension_length);
+  err = clGetDeviceInfo(d->device_id, CL_DEVICE_EXTENSIONS, extension_length,
+                        extensions, NULL);
+  int extension_support = 1;
+  if (err != CL_SUCCESS || strstr(extensions, "cl_khr_il_program") == NULL)
```
franz, line 1437: Small nitpick: AFAIK the presence of this extension means the platform supports `clCreateProgramWithILKHR`; however, from OpenCL 2.1 this API is "core" meaning it should be always available as `clCreateProgramWithIL` (irrespectively if the platform reports support for `cl_khr_il_program` extension)

franz (changes requested):
overall ok, some changes needed

lib/CL/devices/level0/pocl-level0.cc
```diff
@@ -821,32 +821,8 @@ int pocl_level0_setup_metadata(cl_device_id Device, cl_program Program,
...
-    Meta->preferred_wg_multiple =
-        (size_t *)calloc(Program->num_devices, sizeof(size_t));
-    Meta->local_mem_size =
-        (cl_ulong *)calloc(Program->num_devices, sizeof(cl_ulong));
-    Meta->private_mem_size =
-        (cl_ulong *)calloc(Program->num_devices, sizeof(cl_ulong));
-    Meta->spill_mem_size =
-        (cl_ulong *)calloc(Program->num_devices, sizeof(cl_ulong));
+
+    SPIRVParser::mapToPoCLMetadata(I, Program->num_devices, &Program->kernel_meta[Idx]);
```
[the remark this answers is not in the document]
franz: To be clear: it makes sense to have a function that handles metadata setup from SPIRV shared by multiple PoCL drivers, but perhaps we should keep it separate from the parser itself.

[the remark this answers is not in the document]
RABijl: alright, I'll move it to a new `pocl_spirv_utils` file. I do not have a level0 device, so an eye will have to be kept to make sure that removing the function doesn't break anything.

lib/CL/devices/proxy/pocl_proxy.cpp
RABijl: I just checked with example0 and an Nvidia gpu which doesn't have the `cl_khr_il_program` extension and it does run spir-v

lib/CL/devices/proxy/pocl_proxy.cpp
RABijl: This specific piece of code already existed and just got moved around, but l now that i'm digging deeper into it, I'm not even sure if it makes sense, so i'll have to test this.

RABijl:
I've tested the binary functionality and have been able to get the binary created of the supplied spir-v code by the proxied program and then later use it with `clCreateProgramWithBinary`.

lib/CL/devices/proxy/pocl_proxy.cpp
```diff
@@ -1276,55 +1391,119 @@ pocl_proxy_setup_metadata (cl_device_id device, cl_program program,
...
+
+      err = clGetKernelInfo(kernels[index], CL_KERNEL_FUNCTION_NAME,
+                            POCL_MAX_PATHNAME_LENGTH, string_value, NULL);
+      assert(strcmp(p[index].name, string_value) == 0);
+
+      cl_uint prox_num_args;
+      err = clGetKernelInfo(kernels[index], CL_KERNEL_NUM_ARGS,
+                            sizeof(prox_num_args), &prox_num_args, NULL);
+      assert(err == CL_SUCCESS);
+      assert(p[index].num_args = prox_num_args);
```
franz, line 1456: This should be '==' probably ?

pjaaskel:
https://github.com/pocl/pocl/actions/runs/11517624776/job/32062685375?pr=1611#step:5:392 it seems to spit out a lot of warnings, pls check those too.

pjaaskel:
```
The following tests FAILED:
	217 - remote/clSetMemObjectDestructorCallback (Failed)
	223 - remote/test_svm (Failed)
```
I don't see how these changes could affect those tests.

pjaaskel (approved):

RABijl:
> ```
> The following tests FAILED:
> 	217 - remote/clSetMemObjectDestructorCallback (Failed)
> 	223 - remote/test_svm (Failed)
> ```
> 
> I don't see how these changes could affect those tests.

I saw the same ones failing for the llvm-18 pr so I don't think it's related to the code

PR #1614, opened 2024-10-14
ci: bump remote and RDMA to LLVM 18

jansol:
Let's keep them at least somewhat in sync with the main workflows to rule out unrelated causes of errors.


franz (approved):
LGTM

PR #1620, opened 2024-10-15
support building on MinGW part 1

franz:


lib/CL/pocl_util.c
```diff
@@ -24,12 +24,24 @@
...
+
 #include <errno.h>
 #include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
 
+#if defined(__FreeBSD__)
+#include <stdlib.h>
+#elif defined(__WIN32)
```
linehill, line 39: I think, MSVC does not define `__win32`.
```suggestion
#elif defined(_WIN32)
```

lib/CL/pocl_threads.cc
```diff
@@ -0,0 +1,206 @@
+/* OpenCL runtime library: utility functions for thread operations
```
pjaaskel, line 1: ...for C++-supported hosts.

lib/CL/pocl_threads_c.h
```diff
@@ -0,0 +1,154 @@
+/* OpenCL runtime library: utility functions for thread operations
```
pjaaskel, line 1: Implemented using pthreads. Maybe the file name should reflect it as well. It's not using the C standard but the POSIX threads standard, if we are strict here. pocl_threads_posix.h?

lib/CL/pocl_timing.cc
```diff
@@ -0,0 +1,56 @@
+/* OpenCL runtime library: time measurement utility functions
```
pjaaskel, line 1: ...implemented using C++11 (?).

pjaaskel (approved):
Good job. Just a couple of insights you can fix post pull. I suggest @linehill to quickly peruse this before we pull it in.

PR #1624, opened 2024-10-17
Produce some of the WI funcs by the kernel compiler

pjaaskel:
This makes the bitcode easier to read and analyze due to not having the switch...case cluttering the code.


franz (approved):
LGTM

PR #1626, opened 2024-10-21
Fix segfault when using pthread driver with OpenMP with 0 WG

Leixb:


lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -638,9 +638,12 @@ pthread_scheduler_get_work (thread_data *td)
             {
 #ifdef ENABLE_HOST_CPU_DEVICES_OPENMP
               run_cmd = pocl_pthread_prepare_kernel (cmd->device->data, cmd);
-              work_group_scheduler (run_cmd, td);
-              finalize_kernel_command (td, run_cmd);
-              run_cmd = NULL;
+              if (run_cmd)
```
pjaaskel, line 641:
```suggestion
              if (run_cmd != NULL)
```

I prefer explicit NULL pointer comparisons, but it's a matter of taste.

pjaaskel (approved):

lib/CL/devices/pthread/pthread_scheduler.c
Leixb: I kept it consistent with the if statement from `pthread_scheduler_get_work`:
https://github.com/pocl/pocl/blob/e5f604a40b0db81e020b46f925a4cd7331b6fa4e/lib/CL/devices/pthread/pthread_scheduler.c#L592-L597

PR #1637, opened 2024-10-28
Ban variable length arrays

linehill:
Ban VLAs because MSVC does not support them - not even for C code because C99 is not supported where the VLA is a mandatory feature. This restriction is attempted to be enforced with -Werror=vla.

VLAs are replaced with malloc, std::vector or alloca.

### DIFF
--- a/lib/CL/clCreateCommandBufferKHR.c
+++ b/lib/CL/clCreateCommandBufferKHR.c
@@ -153,7 +153,6 @@ POname (clCreateCommandBufferKHR) (
   return cmdbuf;
 
 ERROR:
-  POCL_MEM_FREE (seen_keys);
   if (cmdbuf)
     {
       POCL_MEM_FREE (cmdbuf->queues);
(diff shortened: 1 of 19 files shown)


lib/CL/clCreateCommandBufferKHR.c
```diff
@@ -150,6 +153,7 @@ POname (clCreateCommandBufferKHR) (
   return cmdbuf;
 
 ERROR:
+  POCL_MEM_FREE (seen_keys);
```
jansol, line 156: > Do not attempt to [free](https://linux.die.net/man/3/free)(3) space allocated by alloca()!

https://linux.die.net/man/3/alloca

franz:
Looking at the logs of failed CIs, there are still a bunch of places in drivers using VLAs.

pjaaskel:
https://github.com/pocl/pocl/actions/runs/11590550977/job/32268366371?pr=1637#step:8:1125 TBB uses VLAs.

pjaaskel (approved):

PR #1643, opened 2024-10-29
Grid launcher, a new bin finalization hook, cleanups

pjaaskel:
* Enable including a device-main in the bitcode library. Device mains can have the kernel command handling and launch the work-groups via grid launcher.
* Fix grid launcher generation. It rots because phsa is not tested anymore, but is useful for some targets.
* Allow overriding the final linkage step via a new finalize_binary() hook.
* Cleanups, documentation, minor internal convenience. pocl_init_default_device_infos() now lets setting triplet etc. beforehand and it doesn't overwrite with the 'host' defaults. This enables generating the built-in lib pathnames with the helper with more flexibility.

### DIFF
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -208,31 +208,40 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
 
   POCL_MSG_PRINT_INFO ("Linking final module\n");
 
-  /* Link through Clang driver interface which knows the correct toolchains
-     for all of its targets.  */
-  const char *cmd_line[64]
-    = { pocl_get_path ("CLANG", CLANG), "-o", tmp_module, tmp_objfile };
-  unsigned last_arg_idx = 4;
-  /* Immediate flush enabled results in "__printf_flush_buffer" symbol
-   * referenced it the built kernel.so, however that function exists only
-   * on the host side; therefore link to libpocl.so which provides it */
+  /* If the device has a custom linkage/binary generation step, call it
+     instead of the default Clang-driven linkage step. It's likely a
+     non-host target in that case. */
+  if (device->ops->finalize_binary != NULL)
+    {
+      error = device->ops->finalize_binary (tmp_module, tmp_objfile);
+    }
+  else
+    {
+      /* Link through Clang driver interface which knows the correct toolchains
+         for all of its targets.  */
+      const char *cmd_line[64]
+        = { pocl_get_path ("CLANG", CLANG), "-o", tmp_module, tmp_objfile };
+      unsigned last_arg_idx = 4;
+      /* Immediate flush enabled results in "__printf_flush_buffer" symbol
+       * referenced it the built kernel.so, however that function exists only
+       * on the host side; therefore link to libpocl.so which provides it */
 #ifdef ENABLE_PRINTF_IMMEDIATE_FLUSH
 #ifdef HAVE_DLFCN_H
-  const char *fname = pocl_dynlib_pathname ((void *)pocl_cache_tempname);
-  assert (fname != NULL);
-  cmd_line[last_arg_idx++] = fname;
+      const char *fname = pocl_dynlib_pathname ((void *)pocl_cache_tempname);
+      assert (fname != NULL);
+      cmd_line[last_arg_idx++] = fname;
 #else
 #error ENABLE_PRINTF_IMMEDIATE_FLUSH requires HAVE_DLFCN_H
 #endif
 #endif
-  const char **last_arg = &cmd_line[last_arg_idx];
-  const char **device_ld_arg = device->final_linkage_flags;
-  while ((*last_arg++ = *device_ld_arg++))
-    {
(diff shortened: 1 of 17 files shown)


lib/CL/devices/common.c
```diff
@@ -1693,16 +1718,20 @@ pocl_init_default_device_infos (cl_device_id dev,
...
 
 #else /* No compiler, no CPU info */
   dev->kernellib_name = NULL;
   dev->kernellib_fallback_name = NULL;
-  dev->kernellib_subdir = "host";
-  dev->llvm_cpu = NULL;
+  if (dev->kernellib_subdir != NULL)
+    dev->kernellib_subdir = "host";
+  if (dev->llvm_cpu == NULL)
+    dev->llvm_cpu = NULL;
```
jansol, line 1731: This looks a bit odd

pjaaskel:
I couldn't repro the conformance_*_basic_micro_image failure with LLVM 17 nor LLVM 19 locally.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -492,7 +492,8 @@ static void addStage1PassesToPipeline(cl_device_id Dev,
   addPass(Passes, "mem2reg");
   addAnalysis(Passes, "domtree");
   addAnalysis(Passes, "workitem-handler-chooser");
-  if (Dev->spmd != CL_FALSE) {
+
+  if (!Dev->spmd) {
```
pjaaskel, line 496: Flipped logic. Maybe this causes something...

lib/CL/devices/common.c
```diff
@@ -1446,6 +1455,18 @@ static const cl_image_format supported_image_formats[] = {
...
 
+/** Initializes device info defaults.
+ *
+ * It setups the most generic defaults applicable to most devices,
+ * which the device drivers can overwrite with more specific values.
+ * pocl_cpu_init_common() should be called instead for CPU-devices.
+ * Some of the parameters such as llvm_target_triplet and llvm_cpu
+ * can be set also beforehand, in which case the default host values
+ * are not used and they are used to initialize the rest of the data.
+ *
```
franz, line 1466: Seems unnecessarily complicated....

The CMake variables like OCL_KERNEL_TARGET, OCL_KERNEL_TARGET_CPU and LLC_TRIPLE are *always* related to the host CPU. Is there any other other device where they makes sense ? So perhaps it would be better 1) move the setup of target specific variables (llvm_cpu, llvm_target_triplet etc) into the driver-specific init function (`pocl_cpu_init_common` etc) 2) keep only the kernellib setup in the `pocl_init_default_device_infos` since that can be shared.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -1605,11 +1606,16 @@ int pocl_llvm_codegen(cl_device_id Device, cl_program program, void *Modp,
   pocl_cache_write_kernel_asmfile(AsmFileName, AsmStr.c_str(), AsmStr.size());
   pocl_cache_tempname(ObjFileName, OBJ_EXT, nullptr);
 
+  std::string MCPU = Device->llvm_cpu != nullptr
+                         ? (std::string("-mcpu=") + Device->llvm_cpu)
```
franz, line 1610: Is this always -mcpu on all architectures ? at least the code in `cmake/LLVM.cmake` suggests that it's different on some architectures.  Perhaps use the variable CLANG_MARCH_FLAG ?

lib/CL/pocl_util.h
```diff
@@ -327,6 +327,22 @@ POCL_EXPORT
...
+ *
+ * 1) system() was causing issues (gpu lockups) with HSA when compiling code
+ * (via compile_parallel_bc_to_brig) with OpenCL 2.0 atomics (like CalcPie
+ * from AMD SDK). The reason of lockups is unknown so far.
+ *
+ * 2) system() uses fork() which copies page table maps, and runs out of AS
+ * when pocl has already allocated huge buffers in memory. This happened
+ * in llvm_codegen()
+ *
+ * 3) vfork() does not copy pagetables.
```
franz, line 341: I think this whole justification is ancient and can be deleted.

lib/llvmopencl/Workgroup.cc
```diff
@@ -49,13 +49,15 @@ POP_COMPILER_DIAGS
...
 #include "CanonicalizeBarriers.h"
+#include "DebugHelpers.h"
 #include "KernelCompilerUtils.h"
 #include "LLVMUtils.h"
 #include "ProgramScopeVariables.h"
 #include "VariableUniformityAnalysis.h"
 #include "Workgroup.h"
 #include "WorkitemHandlerChooser.h"
 
+#include "pocl_file_util.h"
```
franz, line 60: Why the include ? i don't see any file function in the diff

pjaaskel:
Some flakiness in the CIs for some reason, rerunning makes them go away. Only ARM64 untested. Trying once more.

franz (changes requested):
some minor suggestions

franz (approved):

PR #1648, opened 2024-11-01
misc: fix dangling pointer in printf_buffer

RABijl:
also includes fixing uint to int conversion
warning and some other clang tidy warnings.

### DIFF
--- a/lib/CL/devices/printf_buffer.c
+++ b/lib/CL/devices/printf_buffer.c
@@ -30,7 +30,6 @@
 #include "printf_base.h"
 
 #include <assert.h>
-#include <stdarg.h>
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


pjaaskel:
Flakiness occurs here with the LLVM 17 bot: `conformance_main_images_cl_fill_images`. Some image tests failed before me as well sporadically.

lib/CL/devices/printf_buffer.c
```diff
@@ -377,7 +371,7 @@ __pocl_printf_format_full (param_t *p, char *buffer, uint32_t buffer_size)
                             "printf error: vector-length is zero\n");
                           return -1;
                         }
-                      if (vector_length > (UINT32_MAX - 9) / 10)
+                      if (vector_length < 0)
```
franz, line 374: Signed integer overflow is undefined behaviour. In this case though, we can just change the condition to: `vector_length > 16` - there are no valid larger vector lengths (and AFAIK no extensions with larger vectors).

franz (changes requested):
small nitpick

lib/CL/devices/printf_buffer.c
RABijl: fixed.

franz (approved):

PR #1651, opened 2024-11-04
Fix few file handling issues

linehill:
* Fix writing a binary to a file in *text mode* garbled the contents due to special handling of some characters on Windows.

* Fix kernel sources got truncated due to `ftell()` returning a value that does not match to the actual file size (on Windows) if the file is opened in text mode ("If the stream is open in text mode, the value returned by this function is unspecified and is only meaningful as the input to fseek()").

* Fix documentation sugggested use of `ftell()` on a file opened in text mode.

### DIFF
--- a/poclu/misc.c
+++ b/poclu/misc.c
@@ -251,7 +251,7 @@ poclu_read_file (const char *filename)
   return res;
 }
 
-static int
+int
 poclu_write_file_in_mode (const char *filename, char *content, size_t size,
                           const char *mode)
 {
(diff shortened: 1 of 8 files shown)


poclu/misc.c
```diff
@@ -272,6 +273,18 @@ poclu_write_file (const char *filename, char *content, size_t size)
   return 0;
 }
 
+int
+poclu_write_file (const char *filename, char *content, size_t size)
```
franz, line 277: This name (and function) is unspecific, while the other two are quite specific WRT mode.

The `poclu_write_file_in_mode` is useful if we e.g. want to append, but keeping this function ("write with platform-specific binary/text handling") is IMO pointless and confusing. I would just keep `poclu_write_file_in_mode` and `poclu_write_binfile`, delete the `poclu_write_file`.

franz (changes requested):
LGTM but one suggestion

poclu/misc.c
linehill: I improved the documentation a bit on the `poclu_write_file()`. Not sure if we can delete the function, at least without deprecating it first, as it is an user facing function, I think. `poclu_write_file_in_mode()` can be made public when there is demand for it.

franz: `user facing function` -> the whole poclu library is for internal use only, mainly to avoid code duplication in tests. There are no guarantees on the API and it is not installed by `make install` (you can check there is no cmake `install` in poclu/CMakeLists.txt), so you can go ahead and delete the function.

franz (approved):

PR #1652, opened 2024-11-04
Use interprocess mutex for llvm_codegen() on Windows

linehill:
... to avoid race condition on `pocl_rename()` calls by multiple processes and threads targeting same files. File renaming and moving does not seems to be atomic operation on windows as `pocl_rename()` calls within the `llvm_codegen()` occationally fail while running internal tests parallel with kernel cache cleared.

Add an API for name interprocess mutex and implement it for Windows. Use it avoid the race condition.

### DIFF
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -39,6 +39,7 @@
 #include <utlist.h>
 
 #ifdef _WIN32
+#include "pocl_ipc_mutex.h"
 #include "vccompat.hpp"
 #endif
 
@@ -124,6 +125,19 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
   char final_binary_path[POCL_MAX_PATHNAME_LENGTH];
   pocl_cache_final_binary_path (final_binary_path, program, device_i, kernel,
                                 command, specialize);
+
+#ifdef _WIN32
+  // Avoid race condition of multiple processes accessing same files.
+  char MtxName[MAX_PATH];
+  strncpy (MtxName, program->build_hash[device_i], MAX_PATH);
+  MtxName[MAX_PATH - 1] = '\0';
+  pocl_ipc_mutex_t ipc_mtx;
+  error = pocl_ipc_mutex_create_and_lock (MtxName, &ipc_mtx);
+  assert (!error && "IPC mutex lock failure!");
+  // For release builds, continue despite not having the mutex - maybe we are
+  // in luck not having a race condition.
+  int have_locked_mutex = !error;
+#endif
 
   if (pocl_exists (final_binary_path))
     goto FINISH;
@@ -251,6 +265,7 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
 
   /* rename temporary kernel.so */
   error = pocl_rename (tmp_module, final_binary_path);
+
   if (error)
     {
       POCL_MSG_PRINT_LLVM (
@@ -279,6 +294,10 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
     }
 
 FINISH:
+#ifdef _WIN32
+  if (have_locked_mutex)
+    pocl_ipc_mutex_unlock_and_release (&ipc_mtx);
+#endif
   pocl_destroy_llvm_module (llvm_module, kernel->context);
   POCL_MEM_FREE (objfile);
   POCL_MEASURE_FINISH (llvm_codegen);
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -127,8 +127,8 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
                                 command, specialize);
 
 #ifdef _WIN32
-  // Avoid race condition of multiple processes accessing same files.
-  char MtxName[MAX_PATH];
+  /* Avoid race condition of multiple processes accessing same files. */
+  char MtxName[MAX_PATH];  /* MAX_PATH is defined by Windows API.  */
   strncpy (MtxName, program->build_hash[device_i], MAX_PATH);
   MtxName[MAX_PATH - 1] = '\0';
   pocl_ipc_mutex_t ipc_mtx;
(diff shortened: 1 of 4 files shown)


lib/CL/devices/common.c
```diff
@@ -125,6 +126,19 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
   pocl_cache_final_binary_path (final_binary_path, program, device_i, kernel,
                                 command, specialize);
 
+#ifdef _WIN32
+  // Avoid race condition of multiple processes accessing same files.
```
pjaaskel, line 130: This is a C file, pls follow the C style guide.

lib/CL/devices/common.c
```diff
@@ -125,6 +126,19 @@ llvm_codegen (char *output, unsigned device_i, cl_kernel kernel,
...
+  char MtxName[MAX_PATH];
```
pjaaskel, line 131: We have POCL_MAX_PATHNAME_LENGTH, is it the same (or at least as large as) MAX_PATH?

lib/CL/pocl_ipc_mutex.h
```diff
@@ -0,0 +1,77 @@
...
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+   IN THE SOFTWARE.
+*/
+/* This header hosts API for named interprocess mutex that does not
```
pjaaskel, line 23:
```suggestion
/* \file pocl_ip_mutex.h named interprocess mutex that does not
```

lib/CL/pocl_ipc_mutex.h
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * The '�pc_mtx' must be a valid object created by pocl_ipc_mutex_create().
+ *
+ * If mutex is locked successfully, zero is returned. Otherwise, on an error,
+ * non-zero is returned.  */
+int pocl_ipc_mutex_lock (pocl_ipc_mutex_t ipc_mtx);
+
+/** Combined variant of pocl_ipc_mutex_create() and pocl_ipc_mutex_lock().  */
+int pocl_ipc_mutex_create_and_lock (const char *name,
+                                      pocl_ipc_mutex_t *ipc_mtx);
```
pjaaskel, line 61: Check the style with formatter scripts. I think the C style wants the return values in their own line, etc.

pjaaskel (changes requested):

lib/CL/devices/common.c
linehill: MAX_PATH is 260.

lib/CL/pocl_ipc_mutex.h
```diff
@@ -1,6 +1,6 @@
 /* pocl_ipc_mutex.h: Named interprocess mutex
 
-   Copyright (c) 2024 Henry Linjam�ki / Intel Finland Oy
+   Copyright (c) 2024 Henry Linjamäki / Intel Finland Oy
```
pjaaskel, line 3: UTF8 instead?

lib/CL/pocl_ipc_mutex.h
```diff
@@ -50,7 +50,7 @@ int pocl_ipc_mutex_create (const char *name, pocl_ipc_mutex_t *ipc_mtx);
  * has locked the mutex. The mutex is unlocked in case the process
  * holding the lock terminates without unlocking first.
  *
- * The '�pc_mtx' must be a valid object created by pocl_ipc_mutex_create().
+ * The 'ípc_mtx' must be a valid object created by pocl_ipc_mutex_create().
```
pjaaskel: Hm encoding broken due to some incapable editor?

pjaaskel (changes requested):

lib/CL/pocl_ipc_mutex.h
linehill: Fixed.

> I think the C style wants the return values in their own line, etc.

I ran `git clang-format-17 --binary /usr/bin/clang-format-17  --style=file:./tools/scripts/style.GNU main -- *.c *.h` but it didn't want to do this.

lib/CL/pocl_ipc_mutex.h
linehill: Fixed.

lib/CL/pocl_ipc_mutex.h
linehill: Fixed.

lib/CL/devices/common.c
linehill: Fixed.

lib/CL/pocl_ipc_mutex.h
pjaaskel: Also the second line seems badly indented. Can you run the clang-format-branch.sh or similar script?

linehill: Fixed manually. 

> Can you run the clang-format-branch.sh or similar script?

No improvement over the `git clang-format ...`.

lib/CL/pocl_ipc_mutex_win.cc
```diff
@@ -0,0 +1,111 @@
...
+
+#include "pocl_debug.h"
+#include "pocl_ipc_mutex.h"
+
+#include <windows.h>
+
+#include <cassert>
+#include <string>
+
+extern "C" int pocl_ipc_mutex_create(const char *Name,
```
pjaaskel, line 33: extern "C" is enough to add to the declaration. No need to add them for the definitions too.

https://stackoverflow.com/questions/1380829/is-extern-c-only-required-on-the-function-declaration

linehill: Addressed.

PR #1653, opened 2024-11-04
Fix off-by-one end iterator

linehill:
This error was mostly harmless as it added an extra array element (C string) which was nullptr. However, it caused a crash while attempting to print the element for debugging.


pjaaskel (approved):

franz:
The timeout is unrelated to this PR, i think i've seen it before a few times, the test does too much work for the CPU to finish in time.

PR #1659, opened 2024-11-05
misc: fix warnings with -Wall and -Wextra

RABijl:
A WIP fixing warnings that `-Wall -Wextra` show so that in the future it would be possible to compile locally and see warnings introduced by new code.
As a side note, the CI could probably also use the `-Wno-unused-function` flag.

Some things being fixed:
* signed compare warnings
* set but unused variables
* bracket warnings

### DIFF
--- a/lib/CL/devices/printf_base.c
+++ b/lib/CL/devices/printf_base.c
@@ -24,8 +24,7 @@
 
 #include "printf_base.h"
 
-#include <stddef.h>
-#include <stdint.h>
+#include <limits.h>
 #include <stdio.h>
 
 void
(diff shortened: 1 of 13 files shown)


pjaaskel (approved):

pjaaskel:
Is this a draft on purpose? Please rebase and let's pull your fixes in otherwise you need to keep rebasing (if your intention is to add more warning fixes over time).

RABijl:
@pjaaskel, yes I was planning to add more fixes over time so that it would be one commit instead of a number of them spread around. I've now rebased it and added a couple more changes, including a file with some macros to tell the compiler some warnings are oke.

lib/CL/devices/printf_base.c
```diff
@@ -428,12 +427,12 @@ unsigned
 __pocl_printf_puts_ljust (param_t *p,
                           const char *string,
                           unsigned width,
-                          int max_width)
+                          int inp_max_width)
```
linehill, line 430: Not sure what "inp" means. Does it mean "input"? Should it be named as "input_max_width"?

RABijl: I've changed the internal unsigned `max_width` to `max_width_u`. This should clear up the confusion.

PR #1663, opened 2024-11-06
Misc. fixes and cleanups

pjaaskel:
* Fix POCL_CONTEXT_COPY64TO32: There was an error due to the truncating conversion.
* Create the parallel.so temp to the parallel.bc dir to enable a potential customized finalization step to create multiple files next to it. It used to be created to the root of the temp.
* Still more argbuffer WG function fixing. Also make the output a bit more readable by naming the argument loads a bit better.
* pocl_run_command_capture_output to drain the pipe.

### DIFF
--- a/lib/CL/devices/bufalloc.h
+++ b/lib/CL/devices/bufalloc.h
@@ -1,7 +1,7 @@
 /* OpenCL runtime/device driver library: custom buffer allocator
 
    Copyright (c) 2011 Tampere University of Technology
-                 2023 Pekka Jääskeläinen / Intel Finland Oy
+                 2023-2024 Pekka Jääskeläinen / Intel Finland Oy
 
    Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


lib/CL/devices/bufalloc.h
```diff
@@ -167,14 +188,17 @@ void pocl_free_chunk (chunk_info_t *chunk);
...
- * @param size Size of the address space.
  */
 POCL_EXPORT
 void pocl_init_mem_region (
     memory_region_t *region, memory_address_t start, size_t size);
 
+/**
+ * Creates a reference to a part of a chunk.
+ *
+ * @todo Register to the parent also so it can free the
```
jansol, line 199:
```suggestion
 * \todo Register to the parent also so it can free the
```

pjaaskel: Will fix this in a future PR.

franz (approved):
LGTM

PR #1665, opened 2024-11-07
Fix PoCL on MSVC toolchain

linehill:
100% internal test pass on "cpu" device and 89% on "basic" device on Windows 11 Enterprise and MS Visual Studio 2022 Professional and Community edition. BUILD_SHARED_LIBS=ON is not supported yet. A CI will be added in a follow up PR.

* Added build instructions.

* Bump minimum CMake version to 3.15 for supporting CMAKE_MSVC_RUNTIME_LIBRARY option.

* Fix path handling in CMake files.

* Fix some clumped strings in CMakeFiles (`"${FOO} ${BAR}"` --> `"${FOO}" "${BAR}"`) that caused compilation errors.

* For MSVC set C language standard to C11 (it's earliest supported C standard in MSVC anyway). Few parts in the code base relies on it.

* Set couple flags to enforce conformant C and C++ behavior on MSVC (but not all of them).

* Suppress `min()` & `max()` macros coming from windows.h. They cause breakage in C++ code. E.g. on `std::numeric_limits<T>::max()`.

* Remove `restrict` --> `__restrict` alias. It breaks compilation on LLVM headers by transforming `__declspec(restrict)` into an invalid code.

* Fix printf specifiers. Replace `%lu` with `%zu`.

* Fix pointer arithmetics on void* pointer.

* Fix extra `;` tokens (e.g. `;;` or `struct foo {; int some_member; ...`). MSVC C compiler does not tolerate them.

* Fix called `free()` for aligned allocations.

* Remove dlopen() & friends in vccompat.hpp. They are redundant now.

* Guard the vccompat.hpp include with _WIN32.

* Compile C sources as C for MSVC (C is supported since VS2019).

* Promote variables used atomically to 64-bits as atomic ops are expecting 64-bit integers.

* Adjust mangling of OpenCL/PoCL built-ins.

* Remove unused includes.

* Suppress "Creating library *.lib and object *.exp" messages coming from MSVC linker. This caused failures on some internal tests.

* Fix/adjust internal tests for MSVC:

  * test_flatten_barrier_subs.cpp: Remove check on input data which is randomly generated (does not affect verification).

  * test_inifinite_loop.cpp: use `quick_exit(0)` instead of `execlp("true", "true, NULL")`. Windows command-line does not recognize "true" command.

  * test_clSetEventCallback.c: Use mutexes from <threads.h> (C11).


CMakeLists.txt
```diff
@@ -24,7 +24,7 @@
 #
 #=============================================================================
 
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
pjaaskel, line 27: What is the CMake shipped with the oldest still freely supported Ubuntu LTS or Debian Stable?

pjaaskel (approved):
Looks great, well done!

CMakeLists.txt
linehill: * For Ubuntu (20.04 if considering up to standard LTS): 3.16.
* For Debian (Bullseye if considering LTS): 3.18.

So it should be fine?

linehill:
Rebased.

PR #1668, opened 2024-11-08
More small tweaks 

pjaaskel:
Misc. changes:
* pocl_init_default_device_infos: assume little endian by default. If BE, redefine after the call.
* printf-buffer: Support for 32b pointer printouts
* Split printf tests to basic/fp64/int64
* Allow running the workgroup tests for non-basic devices
* Fix MAX_EXTENDED_ALIGNMENT to 128
* Allow using the stack size based WG tuning for other targets than CPU-type


tests/workgroup/run_kernel.c
```diff
@@ -66,6 +67,10 @@ main (int argc, char **argv)
   size_t local_work_size[3];
   char kernel_path[2048];
 
+  /* Use the single thread cpu-minimal by default to get guaranteed work-group
+     execution order for stable global id printouts, unless overriden from
+     the environment. */
+  setenv ("POCL_DEVICES", "cpu-minimal", 0);
```
franz, line 73: I don't think putting environment configuration into test code is a good practice. I propose a different solution: if the tests needs stable execution order, then instead of messing with POCL_DEVICES it should check for number of compute-units and return a skip code if it's larger than one. Then it can work with any device that has 1 CU.

tests/kernel/test_printf.cl
```diff
@@ -169,5 +168,6 @@ kernel void test_printf()
 
   printf("|%c|%4c|%-4c|\n", 'a', 'b', 'c');
   printf("|%s|%4s|%-4s|%4s|%.4s|\n", "aa", "bb", "cc", "dddddddddd", "eeeeee");
+  // Print 32b pointers so they come out the same on 64-bit and 32-bit targets.
```
franz, line 171: C comment

franz (changes requested):
minor stuff

tests/workgroup/run_kernel.c
pjaaskel: The problem is how to enforce such a device without forcing it always. Also, even with 1CU the WG execution order is not guaranteed. Not sure what's best here.

pjaaskel:
[Now it was `remote/test_buffer_migration`failing.]

franz (approved):

PR #1674, opened 2024-11-14
Update to match current tensor & DBK spec draft

linehill:
Update tensor and DBK work to match cl_exp_tensor v0.2.0 and cl_exp_defined_builtin_kernels v0.3.1. The changes are mostly just renames and reformatting.

### DIFF
--- a/examples/accel/firmware.c
+++ b/examples/accel/firmware.c
@@ -56,7 +56,7 @@
 
 // NOTE: This enum contains highly experimental built-in kernel IDs, that are
 // subject to change in future PoCL releases without any deprecation period.
-enum BuiltinKernelId : uint16_t
+enum cl_dbk_id_exp : uint16_t
 {
   // CD = custom device, BI = built-in
   // 1D array byte copy, get_global_size(0) defines the size of data to copy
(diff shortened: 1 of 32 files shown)


tools/scripts/format-branch.sh
```diff
@@ -23,12 +23,23 @@ trap 'rm -f $PATCHY' EXIT
 
 git diff main -U0 --no-color >"$PATCHY"
 
-"$RELPATH"/clang-format-diff.py -regex '.*(\.h$|\.c$|\.cl$)' -i -p1 -style=file:"$RELPATH/style.GNU" <"$PATCHY"
+if [ -z "${CLANG_FORMAT_BIN}" ]; then
+  CLANG_FORMAT_BIN=clang-format
+  for i in {25..14}; do
```
franz, line 28: 1) LLVM 14 does not work with our scripts anymore (it can't parse the `style.GNU`), 15 seems to parse it fine
2) could you add this to the other format scripts in `tools/scripts` as well ?

franz (commented):
LGTM, one minor nitpick

tools/scripts/format-branch.sh
linehill: Oops, didn't meant push formatter script changes in this PR.

linehill: Reverted but will be reintroduced in a separate PR.

> LLVM 14 does not work with our scripts anymore (it can't parse the style.GNU)

Seems to be the case.

> could you add this to the other format scripts in tools/scripts as well ?

I'm working on this one.

pjaaskel (approved):

PR #1675, opened 2024-11-14
remote: make the remote faster by going slower

RABijl:
for large files, use the slow queue as the fast
queue can't handle large amounts of data (like onnx blobs).

### DIFF
--- a/lib/CL/devices/remote/communication.c
+++ b/lib/CL/devices/remote/communication.c
@@ -2547,7 +2547,10 @@ pocl_network_build_or_link_program (remote_device_data_t *ddata,
 
   POCL_MSG_PRINT_REMOTE ("Compile/Build/LinkProgram %p\n", netcmd);
 
-  SEND_REQ_FAST;
+  /* The build command contains the source payload which can be orders
+   * of magnitudes bigger than other commands, so send this command on the slow
+   * queue which is better equipped for large payloads.*/
+  SEND_REQ_SLOW;
 
   wait_on_netcmd (netcmd);
 


jansol:
More detailed explanation: the "low-latency" connection asks the network stack to reduce its send and receive buffers to only 4 kilobytes in order to avoid stalls due to buffering. Most PoCL-Remote commands are in the order of 100 bytes and the largest fixed-size command (DeviceInfoReply) is IIRC just a bit shy of 4 kilobytes so this works out quite nicely most of the time. However BuildProgram now takes arbitrary amounts of extra data (easily 100 kB for large OpenCL C kernels, several megabytes or even gigabytes for ONNX blobs for the new DBK API). As a result, this singular command repeatedly fills up the buffers of the TCP stack, which (presumably) triggers an automatic stop-and-wait resend request (ARQ) from the server.

As a result we can observe poll() on the server taking several milliseconds between consecutive reads for a single command, even though nothing of that sort happens when uploading content to huge buffers - PoCL-R easily sustains at least 50Gbps during buffer uploads.

Swapping BuildProgram over to use the "slow" bulk transfer socket avoids this problem and transfers are fast again with large source or binary blobs. Given the nature of this command (and it being synchronous) this seems like a much better tradeoff than the latency optimizations of the fast socket.

jansol (approved):

pjaaskel (changes requested):
Good spot. But can you add (some of) that explanation as a comment next to the modified lines?

lib/CL/devices/remote/communication.c
```diff
@@ -2547,7 +2547,10 @@ pocl_network_build_or_link_program (remote_device_data_t *ddata,
 
   POCL_MSG_PRINT_REMOTE ("Compile/Build/LinkProgram %p\n", netcmd);
 
-  SEND_REQ_FAST;
+  /* The build command contains the source payload which can be orders
+   * of magnitudes bigger than other commands, so send this commands on the slow
```
pjaaskel, line 2551:
```suggestion
   * of magnitudes bigger than other commands, so send this command on the slow
```

pjaaskel (approved):
(The nit can be fixed in the next PR.)

lib/CL/devices/remote/communication.c
RABijl: fixed

PR #1676, opened 2024-11-14
Fix svm_check_get_pointer checking items with NULL vm_ptr

Leixb:
Previously, if `item->vm_ptr` was `NULL` and `item->size > svm_ptr`, we would match with the wrong buffer.


linehill (approved):
LGTM! Thanks.

PR #1678, opened 2024-11-15
Yet more misc. fixes

pjaaskel:
* Reduce the iteration count in for_with_divergent_return test
 * Refactor copy (rect) default implementations: Split to a *memcpy version which takes in a target pointer so it can be called from drivers with a host backing device store pointer or similar use case.
 * tests/kernel/common.cl: enable int64 tests only if cl_khr_int64
 * argbuffer wrapper: Always align args to MAX_EXTENDED_ALIGNMENT to make passing structs on argbuffer feasible.
 * cmake/LLVM: Allow overriding the LLVM_HOST_TARGET
 * Remove test_subbuffers_multidevice: The test runner scripts should setup the device list and run test_subbuffers which tests all available devices.
 * Fix the broken run_cpu_tests script

### DIFF
--- a/cmake/LLVM.cmake
+++ b/cmake/LLVM.cmake
@@ -132,7 +132,9 @@ endif()
 
 string(REPLACE "${LLVM_PREFIX}" "${LLVM_PREFIX_CMAKE}" LLVM_OBJ_ROOT "${LLVM_OBJ_ROOT}")
 run_llvm_config(LLVM_ALL_TARGETS --targets-built)
-run_llvm_config(LLVM_HOST_TARGET --host-target)
+if (NOT DEFINED LLVM_HOST_TARGET)
+  run_llvm_config(LLVM_HOST_TARGET --host-target)
+endif()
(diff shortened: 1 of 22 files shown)


tests/kernel/common.cl
```diff
@@ -8,7 +8,7 @@
 #endif
 
 #undef __IF_INT64
-#if defined(__opencl_c_int64) || defined(cl_khr_int64)
```
franz, line 11: I'm missing an explanation for this change. AFAICT the `__opencl_c_int64` is an actual language feature in the official OpenCL 3.0 Specification, and `cl_khr_int64` is a made-up "extension" used in past by PoCL to signal & check 64bit integer support. So... this seems going in the wrong direction, and also no mention why is this required. Could you add an explanatory comment.

franz (changes requested):
one minor nitpick

franz:
The chipStar test failure is most likely another test relying on atexit cleanup handlers (didn't fail until now); i'll disable it.

The OpenASIP failure seems genuine though.

PR #1680, opened 2024-11-18
misc: cpp related CI warning fixes

RABijl:
types of fixes include:
* sign compare fixes
* using range operators in for loops
* init clearifications
* warnings related to use of move
* removing const from return types

note: this does not address all warnings


franz (approved):
LGTM

PR #1684, opened 2024-11-20
DBK fixes + enable in CI

franz:


lib/CL/devices/devices.c
```diff
@@ -372,21 +372,17 @@ pocl_get_device_type_count(cl_device_type device_type)
...
 
-  cl_device_type device_type_tmp = device_type;
-  if (device_type_tmp == CL_DEVICE_TYPE_ALL)
-    {
-      device_type_tmp = ~CL_DEVICE_TYPE_CUSTOM;
-    }
+  if (device_type == CL_DEVICE_TYPE_DEFAULT)
+  {
+    return pocl_devices ? 1 : 0;
+  }
```
linehill, line 378: Misindentation?

CMakeLists.txt
```diff
@@ -1163,26 +1164,38 @@ function(print_target_properties target)
...
-    message("Found ONNXRuntime headers at ${onnxruntime_INCLUDE_DIRS} for configs ${onnxruntime_IMPORTED_CONFIGURATIONS}")
+    message(STATUS "ONNXRuntime version ${onnxruntime_VERSION} found, with headers at ${onnxruntime_INCLUDE_DIRS} for configs ${onnxruntime_IMPORTED_CONFIGURATIONS}")
   else()
     message(STATUS "ONNXRuntime not found, disabled")
   endif()
 
-  pkg_search_module(LIBJPEG_TURBO IMPORTED_TARGET
-          libjpeg-turbo>=3.0.0 libturbojpeg>=3.0.0)
-  if(NOT LIBJPEG_TURBO_FOUND)
+  find_package(LIBJPEG_TURBO 3.0.0 QUIET CONFIG NAMES libjpeg-turbo)
```
linehill, line 1180: Should the alternate name (libturbojpeg) still be considered?

linehill (approved):
Couple nits and one suggestion. Otherwise, LGTM.

CMakeLists.txt
jansol: My source build (for 3.x, since ubuntu 24.04 only packages version 2.1.5) calls the cmake config `libjpeg-turbo` and provides two pkgconfig files `libjpeg.pc` and `libturbojpeg.pc`. This seems to match the ubuntu package as well - minus `libjpeg.pc` which is not provided there. As far as I can tell, `find_package` only looks for the cmake configuration so I don't think there is anything to gain from the alternate name.

PR #1687, opened 2024-11-22
Ci warning fixes

RABijl:
More fixes, with these most warnings should be gone and hopefully legitimate ones should show up now (like the implicit fallthrough pr fixes). 
The most interesting warnings were:
* -Wredundant-move in common.cc
* -Wpessimizing-move in pocld.cc


tests/runtime/test_command_buffer_multi_device.c
```diff
@@ -120,7 +120,7 @@ main (int _argc, char **_argv)
             devices[j], CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR,
             sizeof (cl_device_id) * num_sync_devices, sync_devices, NULL));
 
-          for (unsigned i = 0; i < num_devices; ++i)
+          for (i = 0; i < num_devices; ++i)
```
linehill, line 123: By glance, this suggests that `i` could be used outside the loop but it's not the case so it looks a bit confusing to me. Perhaps, all loops should be changed from `for(i = ...)` to `for (<type> i = ...)`. 

```suggestion
          for (unsigned i = 0; i < num_devices; ++i)
```

lib/CL/clCreateCommandBufferKHR.c
```diff
@@ -64,6 +64,7 @@ POname (clCreateCommandBufferKHR) (
   cl_uint num_properties = 0;
   cl_command_buffer_properties_khr *seen_keys = NULL;
   cl_bool property_mutable = 0;
+
```
linehill, line 67: Noisy change to me. Does not seem to add value.
```suggestion

```

pocld/virtual_cl_context.cc
```diff
@@ -779,7 +779,7 @@ void VirtualCLContext::BuildOrLinkProgram(Request *req, Reply *rep,
 
       std::vector<unsigned char> binary(buf, buf + bin_size);
       buf += bin_size;
-      assert((buf - buffer) <= (size_t)m.payload_size);
+      assert((size_t)(buf - buffer) <= (size_t)m.payload_size);
```
linehill, line 782: FWIW, maybe use `static_cast` here instead of C-style cast.

RABijl: The whole file uses C-style casts, I did not want to go against the current pattern.

tests/runtime/test_command_buffer_multi_device.c
RABijl: The function uses the pattern of declaring `i` at the top, which you will often see in older C code or code with goto statements. putting the unsigned back will result in a warning that `i` is being shadowed.

linehill: IIUC, the top-level `i`can be removed (after all the loops are transformed to to form in the above).

pocld/virtual_cl_context.cc
linehill: Fair enough.

tests/runtime/test_command_buffer_multi_device.c
RABijl: I've implemented the suggestiong

linehill (approved):
LGTM. Thanks.

PR #1689, opened 2024-11-22
CI: add -Wno-unused-function flag

RABijl:


jansol (approved):

pjaaskel:
Did it warn about anything worth fixing instead? Which functions it warned about?

franz:
First two (`check_for_circular_dep` and `pocl_dump_migration_infos`) are called, but the calls are ifdef-ed out, because they're useful for debugging only. I'd keep the functions but maybe change both the functions and their calls to be behind `#ifdef DEBUGGING_MIGRATIONS` or some such.

`pocl_regen_spirv_binary` is called but the call is behind `#ifdef ENABLE_SPIRV`. This is still required, fixing the warning just needs moving the whole function body behind `#ifdef ENABLE_SPIRV`.

`pocl_cpu_get_memsize` can be removed, and also `pocl_size_ceil2_64` from Workgroup.cc can be removed.

PR #1690, opened 2024-11-22
build fixes for android reference app

Yashvardhan-A:
Some changes to fix building the android reference app with latest pocl main.
 
When PROXY_USE_LIBOPENCL_STUB is enabled then the libopen_stub included in the proxy device is used and we do not need to link to libopencl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1044,7 +1044,9 @@ if(WIN32 AND (NOT MINGW))
 
 elseif(UNIX)
 
-  if (CMAKE_CROSSCOMPILING AND (NOT ENABLE_HOST_CPU_DEVICES) AND (NOT ENABLE_HSA))
+  # if proxy uses opencl-stub then dlfcn is required.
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1044,7 +1044,9 @@ if(WIN32 AND (NOT MINGW))
 
 elseif(UNIX)
 
-  if (CMAKE_CROSSCOMPILING AND (NOT ENABLE_HOST_CPU_DEVICES) AND (NOT ENABLE_HSA))
+  # proxy on android uses opencl-stub that requires dlfcn.
+  if (CMAKE_CROSSCOMPILING AND (NOT ENABLE_HOST_CPU_DEVICES) AND (NOT ENABLE_HSA)
+        AND (NOT (ENABLE_PROXY_DEVICE AND ANDROID)))
```
RABijl, line 1049: The proxy only dlopens when `PROXY_USE_LIBOPENCL_STUB` is set and not limited to just Android, so I'd check for that instead.

RABijl (approved):

jansol (approved):

PR #1693, opened 2024-11-25
misc: remove unused function warnings

RABijl:
This commit removes functions not needed anymore
and puts debugging functions behind ifdefs.


franz (approved):
LGTM

PR #1694, opened 2024-11-25
Android dbk cmake additions

RABijl:
two additions to cmake in order to use the onnx and jpeg dbks


CMakeLists.txt
```diff
@@ -532,6 +532,14 @@ endif()
 
 find_package(PkgConfig MODULE)
 
+# Useful for when the CMAKE_PREFIX_PATH is set by a toolchain file but you need
+# to add another path for pkg-config to look in. For example, when building
+# PoCL with a from source libjpeg-turbo on Android.
+if(POCL_EXTRA_PREFIX_PATH)
```
franz, line 538: "with a from source" is a bit confusing :) also, there already exists an env variable for extra paths for pkg-config: `PKG_CONFIG_PATH`. Isn't that env var enough ?

RABijl: I'd have to try how `PKG_CONFIG_PATH` plays with a toolchain file.

RABijl: I see that the cmakelists.txt has been reverted to using find_package, which allows for setting path to look for as a parameter, so this is not needed anymore. On a related note, Arch does not ship a static version of libjpeg-turbo and I am not sure why PoCL would try to link to a static version.

franz: The official releases (from https://github.com/libjpeg-turbo/libjpeg-turbo/releases) contain both static & shared libs in them, at least the AMD64 deb packages. I've tried to match PoCL's static/shared mode with dependencies (just like with libxsmm), but if it's a problem we can change it.

RABijl: oke I've added a snippet to fallback to the shared lib if the static one is not available.

PR #1697, opened 2024-11-27
dbk fixes

RABijl:
A number of small fixes. And while looking at the the create DBK code, I was wondering why it only returns an invalid value if none of the devices support the set of requested DBKs. I would imagine that would lead to unexpected errors when the program is created just fine, but an error occurs when trying to run the DBK on an unsupported device.

### DIFF
--- a/lib/CL/clCreateProgramWithDefinedBuiltInKernels.c
+++ b/lib/CL/clCreateProgramWithDefinedBuiltInKernels.c
@@ -110,7 +110,11 @@ POname (clCreateProgramWithDefinedBuiltInKernelsEXP) (
       unsigned num_supported_kernels = 0;
       cl_device_id dev = device_list[i];
       if (dev->ops->supports_dbk == NULL)
-        continue;
+        {
+          if (device_support)
+            device_support[i] = CL_DBK_UNSUPPORTED_EXP;
+          continue;
+        }
       for (j = 0; j < num_kernels; ++j)
         {
           errcode = dev->ops->supports_dbk (dev, kernel_ids[j],
--- a/lib/CL/dbk/pocl_dbk_khr_onnxrt_shared.c
+++ b/lib/CL/dbk/pocl_dbk_khr_onnxrt_shared.c
@@ -127,6 +127,12 @@ pocl_copy_onnx_inference_dbk_attributes (
                   data_len);
         }
     }
+  else
+    {
+      attrs->initializer_names = NULL;
+      attrs->initializer_data = NULL;
+      attrs->initializer_tensor_descs = NULL;
+    }
 
   return attrs;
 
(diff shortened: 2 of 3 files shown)


jansol (approved):
LGTM. One note but it's not a blocker.

franz:
> I was wondering why it only returns an invalid value if none of the devices support the set of requested DBKs
> but an error occurs when trying to run the DBK

You will not be able to run it, because you will get an error value when you try to build the program (via `clBuildProgram`). You can create a `cl_program` with all devices in the context, then decide to build it only for a subset of them.

This follows the logic of similar OpenCL APIs like `clCreateProgramWithBuiltInKernels`. The spec on `clCreateProgramWithBuiltInKernels` says about return values:

`CL_INVALID_VALUE if kernel_names is NULL or kernel_names contains a kernel name that is not
supported by any of the devices in device_list`

linehill:
> The spec on clCreateProgramWithBuiltInKernels says about return values: ...

The `clCreateProgramWithDefinedBuiltInKernels()` in the [latest spec](https://htmlpreview.github.io/?https://github.com/linehill/OpenCL-Docs/blob/defined-built-in-kernels/extensions/cl_exp_defined_builtin_kernels.html) returns different error code for unsupported DBKs. I'll copy the relevant parts here:

```
In case of a failure to
create program for a device, one of the following errors code is set
in _device_errcode_ret_ list for the respective device:

* *CL_DBK_UNSUPPORTED_EXP* if the device does not support one of the
   built-in kernels listed in _kernel_ids_.
```

```
* *CL_DBK_UNSUPPORTED_EXP* if _device_errcode_ret_ is NULL and any
   device in _device_list_ does not support a defined built-in kernel.

* *CL_DBK_UNSUPPORTED_EXP* if _device_errcode_ret_ is non-NULL and all
   devices in _device_list_ does not support a defined built-in kernel.
```

RABijl:
> You can create a cl_program with all devices in the context, then decide to build it only for a subset of them.
I see, so that is the place where the programmer has to figure out which devices are not supported.

franz:
@linehill  Right, it has a bit more complicated rules for return value, but if i'm not mistaken, with non-NULL `device_errcode_ret`  the function's return value will match the behavior of `clCreateProgramWithBuiltInKernels` (= error only if none of the devices support some DBK).

linehill:
> ... but if i'm not mistaken, with non-NULL device_errcode_ret the function's return value will match the behavior of clCreateProgramWithBuiltInKernels (= error only if none of the devices support some DBK).

Currently it doesn't but we can improve it if it's desired. Right now the `clCreateProgramWithDefinedBuiltInKernels()` is specified to return `CL_INVALID_UNSUPPORTED_EXP` is none of the devices support a DBK whether the `device_errcode_ret` is NULL or not. Maybe worth of a mention that there is also `CL_INVALID_DBK_ID_EXP` error code for unrecognized DBK ID values.

PR #1698, opened 2024-11-27
Build fix with ENABLE_POCLCC=0

Yashvardhan-A:
cl_offline_compiler.sh.in.cmake includes the target 'spirv_fix_atomic_compare_exchange', which is not added if ENABLE_POCLCC=0.

I am not sure if this is the correct fix for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2171,9 +2171,11 @@ configure_file("cl_offline_compiler.sh.in.cmake" "cl_offline_compiler.sh.config.
 rename_if_different(
   "${CMAKE_BINARY_DIR}/cl_offline_compiler.sh.config.new"
   "${CMAKE_BINARY_DIR}/cl_offline_compiler.sh.config" 1)
-file(GENERATE
-  OUTPUT "${CMAKE_BINARY_DIR}/cl_offline_compiler.sh"
-  INPUT "${CMAKE_BINARY_DIR}/cl_offline_compiler.sh.config")
+if(ENABLE_SPIRV)
+  file(GENERATE
+    OUTPUT "${CMAKE_BINARY_DIR}/cl_offline_compiler.sh"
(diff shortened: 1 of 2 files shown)


franz:
Hi, this is unfortunately not the correct fix, because the cl_offline_compiler.sh is not related to poclcc.  The correct fix is to move the `if(ENABLE_POCLCC) ... endif()` from main `CMakeLists` into `bin/CMakeLists.txt`, around the cmake commands related to poclcc binary. All of the cl_offline_compiler related cmake lines should be if'ed on `ENABLE_SPIRV`.

Yashvardhan-A:
```
# poclcc bin
if(ENABLE_POCLCC)
    add_subdirectory("bin")
endif()
```

Okay makes sense. Also, then this condition in the main `CMakeLists` should include `ENABLE_SPIRV` right? That way the bin folder is included for SPIRV and POCLCC and the conditions in `bin/CMakeLists.txt` should take care of the rest.

CMakeLists.txt
```diff
@@ -2290,7 +2292,7 @@ set(POCLU_LINK_OPTIONS poclu ${OPENCL_LIBS} ${LIBMATH})
 message(STATUS "POCLU LINK OPTS: ${POCLU_LINK_OPTIONS}")
 
 # poclcc bin
-if(ENABLE_POCLCC)
+if(ENABLE_POCLCC AND ENABLE_SPIRV)
```
franz, line 2295: shouldn't this be OR instead of AND ?

franz (changes requested):
one problem, otherwise LGTM

CMakeLists.txt
Yashvardhan-A: Ah yes, my bad

franz (approved):

PR #1708, opened 2024-12-04
Be a bit more explicit about DBK build dependencies

jansol:
The version requirements for onnxruntime and libxsmm in CMakeLists.txt were wrong. Additionally ubuntu 24.04 finally packages libxsmm and that version was accepted by the build scripts... but then fails at build time with little indication of what the problem actually is. This should help avoid losing much time to confusion in the future.

@franz if you have libxsmm set up on CI somewhere this probably breaks since upstream still declares its version as 0.17.0 (latest release tag) even for git builds.


linehill (approved):
LGTM.

franz:
@jansol The CI for DBKs is actually using Github-hosted runners. The `Install XSMM & ONNX & libjpeg-turbo` step in  `.github/workflows/build_linux_gh.yml` needs to be edited to fix the version. Probably enough to add `echo 'release-1.18' >version.txt` before the last `make`.

PR #1712, opened 2024-12-06
Misc fixes

franz:


lib/llvmopencl/Workgroup.cc
```diff
@@ -413,6 +407,77 @@ bool WorkgroupImpl::runOnModule(Module &M, llvm::FunctionAnalysisManager &FAM) {
...
+    // these should be safe to enable
+    F.addFnAttr(Attribute::NoBuiltin);
+    F.addFnAttr(Attribute::NoFree);
+    // recursion forbidden by OpenCL
+    F.addFnAttr(Attribute::NoRecurse);
+    // no exceptions or return by callback
+    F.addFnAttr(Attribute::NoCallback);
+    F.addFnAttr(Attribute::NoUnwind);
+#if 0
+    // TBD: find out if infinite recursion is UB in OpenCL.
```
linehill, line 475: Call recursion is not supported in OpenCL C [OpenCL C 6.11.k].

franz: Oops, wrong comment. It should be "infinite loop"

linehill: There doesn't seem to be exceptions in OpenCL C for infinite loops so C99 probably applies here which means they are defined - aka. a kernel would hang if a work-item enters such loop.

lib/llvmopencl/FixBitshiftUB.cc
```diff
@@ -0,0 +1,134 @@
...
+// on x86-64 CPU, shifting a scalar by a value N will result in actual shift by
+// log2(N) least significant bits of N. IOW, (int32 << 33) is the same as
+// (int32_t << 1). Shifting with AVX registers however behaves differently;
+// shifting by too many bits results in zero.
+//
+// According to the C standard, this is fine; shifting by >type-width bits
+// is undefined behavior, so both ways are acceptable. But in OpenCL C, it is
+// not undefined behavior ; OpenCL C requires *all* shifts follow the "log2(N)
+// least significant bits of N" rule.
+// https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/shiftOperators.html
```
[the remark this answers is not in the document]
franz: You're rigth, this is in fact an issue in the SYCL CTS. I've removed the commit.

lib/llvmopencl/Workgroup.cc
franz: That's my understanding as well. The 'willreturn" attribute breaks with infinite loops (LLVM replaces them with unreachable inst), so i've fixed the comment.

tests/regression/test_infinite_loop.cpp
```diff
@@ -120,6 +123,8 @@ main(void)
 
         std::cout << "OK" << std::endl;
 
+        std::this_thread::sleep_for(std::chrono::milliseconds(1200));
+
```
linehill, line 127: Not quite sure what is the goal of this but it's no biggie. Is it for avoiding crash by exiting too fast?

lib/CL/devices/common_utils.c
```diff
@@ -342,6 +342,24 @@ pocl_cpu_init_common (cl_device_id device)
...
+          | CL_DEVICE_LOCAL_FP_ATOMIC_MIN_MAX_EXT;
+      device->features
+        = HOST_DEVICE_FEATURES_30 " __opencl_c_ext_fp32_global_atomic_add "
+                                  "__opencl_c_ext_fp64_global_atomic_add"
+                                  " __opencl_c_ext_fp32_local_atomic_add "
+                                  "__opencl_c_ext_fp64_local_atomic_add"
+                                  " __opencl_c_ext_fp32_global_atomic_min_max "
+                                  "__opencl_c_ext_fp64_global_atomic_min_max"
+                                  " __opencl_c_ext_fp32_local_atomic_min_max "
+                                  "__opencl_c_ext_fp64_local_atomic_min_max";
```
linehill, line 360: Could you format the literals consistently - e.g. have  a single space always on the left side?

linehill (approved):
Left a suggestion for literal formatting. Otherwise, LGTM.

tests/regression/test_infinite_loop.cpp
franz: It's to make the test more reliable. I've noticed while tuning function attributes, that the test would crash on the Apple CI but not x86-64 CI. The reason of crash was that kernel function contained unreachable inst, but the the x86-64 managed to exit (with success) before the kernel function was ran by the CPU driver, while Apple did actually run the kernel and crashed.

lib/CL/devices/common_utils.c
franz: fixed in next PR

PR #1717, opened 2024-12-11
remote: handle EINTR error

RABijl:
In the case of an interrupt, try reading the data
again. This may not be the most considerate thing
to do, but it will read th data.

I encountered this case on Android where establishing the connection can take a while.


jansol (approved):

pjaaskel (approved):

pjaaskel:
CodeQL fails to build the code, but must be not be because of this PR.

PR #1718, opened 2024-12-11
dbk: fix circular dependency

RABijl:
I noticed a circular dependency that would occasionally cause problems.

### DIFF
--- a/lib/CL/pocl_mem_management.c
+++ b/lib/CL/pocl_mem_management.c
@@ -1090,3 +1090,18 @@ pocl_find_raw_ptr_with_dev_ptr (cl_context context, const void *dev_ptr)
   POCL_UNLOCK_OBJ (context);
   return item;
 }
+
+void *
+pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id)
+{
+  if (arg->value == NULL)
+    return NULL;
+
+  if (arg->is_raw_ptr)
+    return *(void **)arg->value;
+
+  cl_mem mem = *(cl_mem *)(arg->value);
+  char *ptr = (char *)(mem->device_ptrs[global_mem_id].mem_ptr);
+  ptr += arg->offset;
+  return (void *)ptr;
+}
--- a/lib/CL/pocl_mem_management.h
+++ b/lib/CL/pocl_mem_management.h
@@ -25,6 +25,11 @@
 #include "pocl_cl.h"
 #include "utlist.h"
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 #ifdef __GNUC__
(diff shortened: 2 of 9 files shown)


pjaaskel (changes requested):

linehill:
Is this much effort needed to break the circular dependency (of header includes)? Can't the issue be solved with forward declarations? What is the error message?

RABijl:
@linehill there were some linking errors that would occasionally show up when disabling the ocl_icd that I attributed to the common utils being both using and being used by the nms_dbk and the way I built the library. I've now force pushed a simplified way of doing it and so far no issues on my x86 and android builds.

lib/CL/pocl_mem_management.c
```diff
@@ -1090,3 +1090,24 @@ pocl_find_raw_ptr_with_dev_ptr (cl_context context, const void *dev_ptr)
   POCL_UNLOCK_OBJ (context);
   return item;
 }
+
+/**
```
pjaaskel, line 1094: I think this comment belongs to the header since it's a "public function".

RABijl: fixed

lib/CL/pocl_mem_management.h
```diff
@@ -111,6 +116,19 @@ pocl_convert_to_subbuffer_migrations (pocl_buffer_migration_info *buffer_usage,
...
 #endif
 
+/**
+ * Get the device memory pointer of the supplied pocl argument.
+ *
+ * \param global_mem_id [in] This is needed to get the device specific pointer.
+ * \return NULL if arg->value is NULL and otherwise the requested pointer.
+ */
+POCL_EXPORT
+void *pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id);
```
pjaaskel, line 126: I think the return value should be in its own line, but clang-format gets confused by POCL_EXPORT. Please fix that in the next PR.

pjaaskel (approved):

lib/CL/pocl_mem_management.h
RABijl: most of the uses of pocl_export I see look like this.

pjaaskel: Yeah, but our style guide is "GNU style for the C sources" and we wan tho converge towards the style guide whenever we commit. Looking at [the guide](https://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting), the correct one would be maybe to have `POCL_EXPORT void` in the same line?

RABijl: I come to the same conclusion reading the guide, I've updated it.

PR #1459, opened 2024-05-08
pocl_util: fix numberous deadlocks on disconnect

RABijl:
non exhaustive list:
* fix event being marked complete after being marked failed
* fix deadlock between event and memobj happening in pocl_update_event_finished and pocl_create_migration_commands
* fix deadlock of event trying to remove it self from wait_list of event trying notifying that it has failed.
* catch error when a event is created on a command queue that is not available

I'll add some comments to explain parts of the commit

PR #1719, opened 2024-12-12
event: add retain event during event sync

RABijl:
This is for now a wip and is intended for people who are interested in things related to events. This is a second attempt at #1459 .

Since the waiting event is added to the notify_list of the notifier_event, a reference should be kept
to prevent the notifier event from notifying an
event that doesn't exist anymore. With the recent
changes to the lock_events_inorder, this also
greatly simplifies the broadcast code.

### DIFF
--- a/include/pocl.h
+++ b/include/pocl.h
@@ -506,6 +506,17 @@ typedef union
 typedef struct _pocl_buffer_migration_info
   pocl_buffer_migration_info;
 
+/**
+ * Enum that represents different possible states of a command node.
+ */
+typedef enum
+{
+  POCL_COMMAND_FAILED = -1,
+  /* Default unless set to ready. */
+  POCL_COMMAND_NOT_READY = 0,
+  POCL_COMMAND_READY = 1,
+} pocl_command_state;
+
 /* One item in a command queue or a command buffer. */
 typedef struct _cl_command_node _cl_command_node;
 struct _cl_command_node
@@ -538,7 +549,10 @@ struct _cl_command_node
   cl_device_id device;
   /* The index of the targeted device in the **program** device list. */
   unsigned program_device_i;
-  cl_int ready;
+
+  /* Used to indicate if a node is ready to be executed or failed. */
+  pocl_command_state state;
+
   /* true if the node belongs to a cl_command_buffer_khr */
   cl_int buffered;
 
--- a/lib/CL/devices/remote/remote.c
+++ b/lib/CL/devices/remote/remote.c
@@ -1536,7 +1536,7 @@ pocl_remote_submit (_cl_command_node *node, cl_command_queue cq)
   POCL_INIT_COND (e_d->event_cond);
   e->data = (void *)e_d;
 
-  node->ready = 1;
+  node->state = POCL_COMMAND_READY;
   if (pocl_command_is_ready (node->sync.event.event))
     {
       pocl_update_event_submitted (node->sync.event.event);
@@ -1599,7 +1599,7 @@ pocl_remote_notify (cl_device_id device, cl_event event, cl_event finished)
       return;
     }
 
-  if (!node->ready)
+  if (node->state != POCL_COMMAND_READY)
     {
       POCL_MSG_PRINT_EVENTS (
           "remote: command related to the notified event %lu not ready\n",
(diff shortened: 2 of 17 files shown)


include/pocl.h
```diff
@@ -506,6 +506,17 @@ typedef union
 typedef struct _pocl_buffer_migration_info
   pocl_buffer_migration_info;
 
+/**
+ * Enum that represents different possible states of a command node.
+ */
+typedef enum
+{
+  POCL_COMMAND_FAILED = -1,
```
pjaaskel, line 514: Can you add some examples when this state could be invoked. Maybe we can use it for implementing #1619 also?

RABijl: I'm currently using it in the remote driver to mark a command node as failed when the device is not available. The remote driver does some bookkeeping regarding inflight commands and I'm using this to handle things a bit differently when it fails.
It shares a lot of purpose with the status of the command node's sync event, but since it is possible for the node to use something else than events, I've resorted to this instead.

RABijl: regarding the issue you mentioned, it could work. I'd have to check the code to be sure, but I believe that an event failing will cascade that to the rest of the events. 

And as it so happened, I recently did an out of bounds read in a kernel on the Nvidia OpenCL library and it would not crash, but instead return a CL_OUT_OF_RESOURCES when trying to read the output buffer. It was a bit confusing to figure why it was happening, but it does communicate to the host something is wrong.

lib/CL/devices/remote/remote.c
```diff
@@ -1536,7 +1536,7 @@ pocl_remote_submit (_cl_command_node *node, cl_command_queue cq)
   POCL_INIT_COND (e_d->event_cond);
   e->data = (void *)e_d;
 
-  node->ready = 1;
+  node->state = POCL_COMMAND_READY;
   if (pocl_command_is_ready (node->sync.event.event))
```
jansol, line 1540: I think this should not be called at all here, just submit the command unconditionally to the remote.

lib/CL/pocl_util.c
```diff
@@ -364,8 +364,20 @@ pocl_create_event_sync (cl_event notifier_event, cl_event waiting_event)
         }
     }
 
-  if (notifier_event->status == CL_COMPLETE)
-    goto FINISH;
+  /* If the notifier event is already complete (or failed),
+     don't create an event sync. This is fine since if the wait
+     event has no notifier events and gets submitted, it can start
```
franz, line 369: In case the notifier event fails, wouldn't this (not creating the sync) break the cascading of the failure to the dependent (waiting) events ? AFAICT the cascading is done through broadcast -> notify -> update_failed -> broadcast .. and that only works if there is an event sync ?

lib/CL/pocl_util.c
```diff
@@ -2125,12 +2103,15 @@ pocl_update_event_finished (cl_int status, const char *func, unsigned line,
...
+                          const char *func,
                           unsigned line,
                           cl_event event,
                           const char *msg)
 {
-  pocl_update_event_finished (CL_FAILED, func, line, event, msg);
+  /* Should only be used with error statuses. */
+  assert (status < 0);
+  pocl_update_event_finished (status, func, line, event, msg);
 }
```
franz, line 2116: you can also remove `pocl_update_event_failed_locked()`

lib/CL/pocl_util.c
franz: the PR overall LGTM (minus this unresolved issue)

RABijl: you're right that this would not create a cascading scenario. I went with the scenario that the failed notifier event already called broadcast before the sync is made, in that case you do not want to create the sync as the wait event will never be notified. My first thought to fix this cascading would be to fail it at the API level, but `clEnqueueNDRangeKernel` does not mention anything specifically about failed events, just invalid events. In any case, I think that the proper way of doing this is to fail the creation commands that are calling create_event_sync as whatever this waiting event belongs to is bound to fail.

franz: > My first thought to fix this cascading would be to fail it at the API level
> fail the creation commands that are calling create_event_sync

OK, i need to take another look, but I don't see how this is possible to fix at API (enqueue time), or event creation commands - because failure can occur at execution time, which is asynchronous to the enqueue time. Let's consider this: The user enqueues 100 events to a queue. The driver starts immediately executing them after first enqueue, but the execution is slow (much slower than enqueue). The user calls clFinish() after enqueueing all 100 events, but at the same time the driver is executing only the 10th event. Then the 20th event fails for some reason. How is the failure dealt with ? All of the events (and their syncs) have already been created, right ?

franz:
Created a new PR #1850 which has this PR + some additional fixes, continuing there.

PR #1721, opened 2024-12-12
Fix compiler expansion of certain WIFuncs and other small things

pjaaskel:
* Now lets the linker skip the functions if it detects all calls can be handled by the compiler, to avoid dealing with them in the inliner passes (Flatten*).
* dumpCFG: Highlight printfs and their format strings
* Add the LLVM IR checks to the 'internal' suite
* Code polishing etc. [NFC]


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -724,9 +724,21 @@ llvm::Type *SizeT(llvm::Module *M) {
...
+bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(const llvm::Function &F) {
+    Instruction::const_use_iterator UI = F.use_begin(), UE = F.use_end();
+    for (; UI != UE; ++UI) {
+      llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
+      if (Call == nullptr)
+        continue;
+      if (!isCompilerExpandableWIFunctionCall(*Call))
+        return false;
+    }
+    return true;
```
linehill, line 736: Indentation looks off here.

linehill (approved):
A nit. Otherwise, LGTM.

pjaaskel:
Thanks. Will reindent the block in the next PR.

PR #1726, opened 2024-12-14
cl_ext_buffer_device_address + some other fixes

franz:


lib/CL/devices/common_utils.c
```diff
@@ -265,6 +265,9 @@ pocl_cpu_init_common (cl_device_id device)
   device->llvm_cpu = OCL_KERNEL_TARGET_CPU;
   if (device->llvm_cpu == NULL)
     device->llvm_cpu = pocl_get_llvm_cpu_name ();
+
+  if (strstr (OCL_KERNEL_TARGET, "x86") != NULL)
+    device->run_safe_division_pass = CL_TRUE;
```
pjaaskel, line 270: Maybe 'safe' is not the best description for it? Better could be _noexcept_ or _notrap_ similar?

I think the non-trapping divisions are actually less safe ones in this sense.

lib/CL/pocl_cl.h
```diff
@@ -1126,6 +1126,11 @@ struct _cl_device_id {
   /* The program scope variable pass takes program-scope variables and replaces
      them by references into a buffer, and creates an initializer kernel. */
   cl_bool run_program_scope_variables_pass;
+  /* Some architectures (x86) have trapping div-by-zero and no way to
+     disable this behavior. OpenCL explicitly forbids raising exceptions on
+     division for any values that trigger UB. Therefore we must check arguments
```
pjaaskel, line 1131: undefined behavior?

lib/llvmopencl/SafeDivision.cc
```diff
@@ -0,0 +1,184 @@
...
+#include <set>
+#include <string>
+
+#include "pocl_llvm_api.h"
+
+// #define DEBUG_SAFE_DIVISION
+
+#define PASS_NAME "safe-division"
+#define PASS_CLASS pocl::SafeDivision
+#define PASS_DESC "Handle integer division&remainder UB in kernel code"
```
pjaaskel, line 51: Or: "Change division and remainder instructions to non-trapping ones"

lib/llvmopencl/SafeDivision.cc
```diff
@@ -0,0 +1,184 @@
...
+#include <iostream>
+#include <set>
+#include <string>
+
+#include "pocl_llvm_api.h"
+
+// #define DEBUG_SAFE_DIVISION
+
+#define PASS_NAME "safe-division"
+#define PASS_CLASS pocl::SafeDivision
```
pjaaskel, line 50: NonTrappingDivision ?

lib/llvmopencl/SafeDivision.cc
```diff
@@ -0,0 +1,184 @@
...
+
+  bool Changed = false;
+  for (Function::iterator I = F.begin(), E = F.end(); I != E; ++I) {
+    for (BasicBlock::iterator BI = I->begin(), BE = I->end(); BI != BE;) {
+      BinaryOperator *BinI = dyn_cast<BinaryOperator>(BI++);
+      if (BinI == nullptr)
+        continue;
+      if (!BinI->isIntDivRem())
+        continue;
+      // TODO do we need to handle case of division by zero constant ?
```
pjaaskel, line 135: I think we might have to and we might have to deal with them on -O0 output since optimizers might change those to unreachable or similar?

lib/llvmopencl/SafeDivision.cc
```diff
@@ -0,0 +1,184 @@
...
+      if (dyn_cast<Constant>(BinI->getOperand(1)))
+        continue;
+#ifdef DEBUG_SAFE_DIVISION
+      std::cerr << "Fixing division/remainder in Inst: \n";
+      BinI->dump();
+#endif
+      Value *Dividend = BinI->getOperand(0);
+      Value *Divisor = BinI->getOperand(1);
+      // unlikely case but its possible optimization haven't run yet
```
pjaaskel, line 144: Yes, we have to be prepared to be ran on unoptimized Clang FE output.

lib/llvmopencl/SafeDivision.h
```diff
@@ -0,0 +1,46 @@
...
+#include "config.h"
+
+#include <llvm/IR/Module.h>
+#include <llvm/IR/PassManager.h>
+#include <llvm/Pass.h>
+#include <llvm/Passes/PassBuilder.h>
+
+namespace pocl {
+
+class SafeDivision : public llvm::PassInfoMixin<SafeDivision> {
```
pjaaskel, line 36: NonTrappingDivision or something more descriptive?

lib/CL/devices/devices.c
```diff
@@ -689,10 +682,6 @@ pocl_init_devices ()
 #endif
 
 #ifdef ENABLE_SIGNAL_HANDLERS
-  if (pocl_get_bool_option ("POCL_SIGFPE_HANDLER", 1))
-    {
-      pocl_install_sigfpe_handler ();
```
pjaaskel, line 694: Do we really want to remove the old solution? As far as I see, there's quite a bit of injected code which affects the performance. Perhaps for Linux the signal handler-trapping one is faster and enables better vectorization?

pjaaskel (changes requested):
Perhaps send the notrap div as a separate PR and consider my suggestions there. BDA looks good, but there might be some changes for 1.0.0 as the discussion was not finished after all.

lib/CL/devices/common_utils.c
franz: > non-trapping divisions are actually less safe ones in this sense.

Can you explain what you mean, how are they less safe ?

lib/CL/devices/devices.c
pjaaskel: I doubt div/mod-by-zero etc. occurs in performance critical cases, it's like a bug ~always, the exception should not happen thus no overhead in the common case. If the vectorization performance is not improved in x86, then the scalar performance should be as you add those comparisons there. How much in real world benchmarks, likely not much, but if it's easy to keep both alternatives, why not?

BTW did you check what is the case with Arm/RISC-V: Do they have excepting mod/div and do they have vector ISA for integer div?

lib/CL/devices/common_utils.c
pjaaskel: Because I think div-by-zero is usually a application bug and this pass prevents the bug from causing exceptions. Yep, in OpenCL case there's no observable difference as exceptions are not allowed, but I cannot understand why it would be 'safer' either.

lib/CL/devices/devices.c
franz: > but if it's easy to keep both alternatives, why not?

downside is maintenance. But ok, i'll restore it.

> did you check what is the case with Arm/RISC-V

ARM has a control register bit to enable/disable trap on div-by-zero. I couldn't find what is the behavior of INT_MIN divided by -1.

RISC-V [does not trap](https://five-embeddev.com/riscv-user-isa-manual/Priv-v1.12/m.html#division-operations) and it seems there is no option to enable trap.

lib/llvmopencl/SafeDivision.cc
franz: yep, will remove the comment

lib/llvmopencl/SafeDivision.cc
franz: Good point, i'll add the code to handle it.

lib/llvmopencl/SafeDivision.cc
franz: I think that would be a misleading description. We're not changing instructions; also it implies that UB results in trap - which is only true on some architectures. However this pass is architecture-independent. So IMO the current description is more correct - it turns UB corner cases into defined behavior. It might even make sense to always enable it - it would lead to identical results of div/rem math across all architectures.

lib/CL/devices/common_utils.c
franz: It is safer in the sense that it has defined behavior for corner cases. IOW, you can rely on the behavior being the same, on all architectures. "noexcept" or "notrap" would imply that UB results in a trap, which isn't always true. But i'll change the name to something else.

lib/CL/devices/devices.c
franz: restored the old handler

pjaaskel (approved):

PR #1729, opened 2024-12-16
Cleanups [NFC]

pjaaskel:
Also dropped the crazy INT_MAX iteration count in for_divergent_return.cl to 100 to reduce the runtime of the test case without affecting the tested aspect.

### DIFF
--- a/lib/llvmopencl/LLVMUtils.cc
+++ b/lib/llvmopencl/LLVMUtils.cc
@@ -725,15 +725,15 @@ llvm::Type *SizeT(llvm::Module *M) {
 }
 
 bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(const llvm::Function &F) {
-    Instruction::const_use_iterator UI = F.use_begin(), UE = F.use_end();
-    for (; UI != UE; ++UI) {
-      llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
-      if (Call == nullptr)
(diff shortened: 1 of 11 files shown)


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -725,15 +725,15 @@ llvm::Type *SizeT(llvm::Module *M) {
...
-      llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
-      if (Call == nullptr)
-        continue;
-      if (!isCompilerExpandableWIFunctionCall(*Call))
-        return false;
-    }
-    return true;
+  Instruction::const_use_iterator UI = F.use_begin(), UE = F.use_end();
+  for (; UI != UE; ++UI) {
+    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
```
linehill, line 730: Could be simplified with something like this:
```suggestion
  for (const auto *U : F.uses()) {
    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(U->getUser());
```

linehill (approved):
A suggestion. Otherwise, LGTM.

lib/llvmopencl/LLVMUtils.cc
pjaaskel: Thanks, will change this in a future PR.

PR #1736, opened 2024-12-18
WIloops: context data rematerialization

### LINKED ISSUE #1525: Kernel execution fails with SIGBUS (Misaligned address error)

pjaaskel:
WILoops: Private (temp) variable rematerialization

Attempt to rematerialize work-item private temp variables that are used across multiple parallel regions instead of always using context save/restore via stack. Rematerialization here means recomputing the value for each user.

Since this feature reduces context data needs a lot, we can now enforce outer loop vectorization more aggressively even when the
original kernel doesn't use barriers.

Should fix #1525 (and checks off one more item in my 10+ year old bucket list).


lib/llvmopencl/WorkitemLoops.cc
```diff
@@ -893,6 +891,101 @@ llvm::AllocaInst *WorkitemLoopsImpl::getContextArray(llvm::Instruction *Inst,
...
+      std::cerr << "##### cannot remat a call" << std::endl;
+      Call->dump();
+#endif
+      if (CanDoIt != nullptr)
+        *CanDoIt = false;
+      return nullptr;
+    }
+  } else if (isa<Constant>(Def) || isa<Argument>(Def)) {
+    if (CanDoIt != nullptr)
+      return nullptr;
```
linehill, line 951: Should this be:
```suggestion
    if (CanDoIt != nullptr)
      *CanDoIt = true;
```
?

lib/llvmopencl/WorkitemLoops.cc
```diff
@@ -979,14 +1089,70 @@ void WorkitemLoopsImpl::addContextSaveRestore(llvm::Instruction *Def) {
       assert(IncomingBB != NULL);
       ContextRestoreLocation = IncomingBB->getTerminator();
     }
-    llvm::Value *LoadedValue =
-        addContextRestore(UserI, Alloca, Def->getType(), PaddingAdded,
-                          ContextRestoreLocation, isa<AllocaInst>(Def));
-    UserI->replaceUsesOfWith(Def, LoadedValue);
+
+    llvm::Value *LocalValue = nullptr;
```
linehill, line 1093: Would it be more clear if this variable were called `RematerializedValue`?

lib/llvmopencl/WorkitemLoops.cc
```diff
@@ -893,6 +891,101 @@ llvm::AllocaInst *WorkitemLoopsImpl::getContextArray(llvm::Instruction *Inst,
...
+  llvm::Instruction *Inst = dyn_cast<Instruction>(Def);
+  if (Inst == nullptr) {
+#ifdef DEBUG_WORK_ITEM_LOOPS
+    std::cerr << "##### cannot remat (unsupported value type)" << std::endl;
+    Def->dump();
+#endif
+    if (CanDoIt != nullptr)
+      *CanDoIt = false;
+    return nullptr;
+  }
```
linehill, line 973: Probably should return nullptr if the instruction may write to memory or have side-effects.

lib/llvmopencl/WorkitemLoops.cc
pjaaskel: No, I cannot set it to true due to recursion. It should be initialized to true and then if any of the recursively done checks sets it to false, it should not be flipped back.

lib/llvmopencl/WorkitemLoops.cc
pjaaskel: Originally the idea was to use it for both remat and normal context restored value, but yep, I'll refactor it to make this spot more readable.

lib/llvmopencl/WorkitemLoops.cc
linehill: I see. The surrounding code made me think that there was an error here.

linehill (approved):
LGTM.

PR #1740, opened 2024-12-19
Fix wrong computation of dynamic WG size

pjaaskel:
This caused the failure of test_program_from_binary_with_local with specialization off and outer loop vectorization on.


tneval (approved):
Confirmed that this resolves the issue.

PR #1744, opened 2024-12-20
A fix for self-referential macros of OCL builtins

### LINKED ISSUE #1735: Self-referential macro with an OpenCL builtin
A reduced case of a use of a [self-referential macro](https://www.math.utah.edu/docs/info/cpp_1.html#SEC25) from llama.cpp:
```
#define sub_group_reduce_add sub_group_reduce_add
kernel void test(global int *d) { *d = sub_group_reduce_add(*d); }
```
This compiles on Intel OpenCL (GPU) and PoCL-Level0 but not on PoCL-CPU:
```
error: /home/linehill/.cache/pocl/kcache/tempfile_TnTaja.cl:2:40 <Spelling=/home/linehill/.cache/pocl/kcache/tempfile_TnTaja.cl:1:30>: use of undeclared identifier 'sub_group_reduce_add'
warning: /home/linehill/.cache/pocl/kcache/tempfile_TnTaja.cl:1:9: 'sub_group_reduce_add' macro redefined
warning: /home/linehill/.cache/pocl/kcache/tempfile_TnTaja.cl:1:9: 'sub_group_reduce_add' macro redefined
```
For some reason the error occurs with `sub_group_reduce_add`. E.g. The following compiles:
```
#define foo foo
int foo(int x) { return x; }
kernel void test(global int *d) { *d = foo(*d); }
```

linehill:
Fixes #1735 by including the OpenCL built-in rename header in a separate preprocessing step.

### DIFF
--- a/cmake/bitcode_rules.cmake
+++ b/cmake/bitcode_rules.cmake
@@ -91,10 +91,12 @@ function(compile_cl_to_bc FILENAME SUBDIR BC_FILE_LIST EXTRA_CONFIG)
     endif()
 
     set(DEPENDLIST
+          "${CMAKE_SOURCE_DIR}/include/_builtin_renames.h"
           "${CMAKE_SOURCE_DIR}/include/_kernel.h"
           "${CMAKE_SOURCE_DIR}/include/_kernel_c.h"
           "${CMAKE_SOURCE_DIR}/include/pocl_types.h")
     set(INCLUDELIST
+        "-include" "${CMAKE_SOURCE_DIR}/include/_builtin_renames.h"
         "-include" "${CMAKE_SOURCE_DIR}/include/_kernel.h"
         "-include" "${CMAKE_SOURCE_DIR}/include/_enable_all_exts.h")
 
--- a/include/_kernel.h
+++ b/include/_kernel.h
@@ -51,8 +51,6 @@
 
 #include "_enable_all_exts.h"
 
-#include "_builtin_renames.h"
-
 /* Define some feature test macros to help write generic code. These are used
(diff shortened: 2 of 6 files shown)


pjaaskel (approved):
This is fine, but I keep thinking if there'd be a cleaner solution to the original problem than the rename macros that might cause some other troubles later.

pjaaskel:
The failure seems relevant:
```
error: /srv/github/actions-runner/_work/_temp/GH_POCL_CACHE/tempfile_mvQ4TT.cl:38:16: use of undeclared identifier 'UNROLL_FACTOR'
```
Some -D not propagated?

franz:
> if there'd be a cleaner solution to the original problem

I was thinking about it too, but i haven't figured out one. I've seen someone suggesting a Clang option to specify mangling, but i think it never went anywhere.

> rename macros that might cause some other troubles later.

TBH I'm not sure if this is a problem on PoCL's side. It depends, if OpenCL specifies that it's mandatory to declare builtins as functions. If not, our macros may be fine. E.g. if you redefine errno in your C program, you can hardly complain to glibc maintainers that your program stopped working.

linehill:
> Some -D not propagated?

No, it's not that. The problem here is that macros are not expanded in (some of the) pragmas. In the failing test the `UNROLL_FACTOR` doesn't get expanded in an unroll pragma in the first preprocess step.

Related LLVM issues: [[1](https://github.com/llvm/llvm-project/issues/112152)][[2](https://github.com/llvm/llvm-project/issues/65992)].

linehill:
Closing, I'll consider another solution later.

PR #1751, opened 2025-01-04
fix clGetMemAllocInfoINTEL queries for SVM pointers

bashbaug:
When passed an SVM pointer, clGetMemAllocInfoINTEL should return UNKNOWN as the USM type, and default placeholder values for other queries.

This is needed to differentiate between SVM pointers, which should be freed using clSVMFree, and USM pointers, which should be freed using clMemFreeINTEL or clMemBlockingFreeINTEL.


pjaaskel (approved):
LGTM.

PR #1777, opened 2025-01-23
Vectorize builtins

franz:


README.md
```diff
@@ -55,6 +55,8 @@ can be used from the build directory (without installing it system-wide).
 
 ### Support Matrix legend:
 
+:large_blue_diamond: Tested in CI extensively, passed the official upstream OpenCL-CTS
```
pjaaskel, line 58: Perhaps make it more emphasized that the port has been given the official conformance status?

doc/sphinx/source/conformance.rst
```diff
@@ -113,146 +103,11 @@ Known issues related to CTS
...
-Conformance tests results (precision of builtin math library functions)
+Conformance tests results
 -----------------------------------------------------------------------
 
-Note that it's impossible to test double precision on the entire range,
-therefore the results may vary.
-
-x86-64 CPU with AVX2+FMA, LLVM 4.0, tested on Nov 1, 2017
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+PoCL has submitted for OpenCL conformance in December 2024, and has been
```
pjaaskel, line 110: Why remove the title? It's important info related to which configuration and driver has been stamped? Also it would be good tag the exact commit which was approved in the repo.

lib/llvmopencl/Workgroup.cc
```diff
@@ -445,6 +445,7 @@ bool WorkgroupImpl::runOnModule(Module &M, llvm::FunctionAnalysisManager &FAM) {
     F.removeFnAttr(Attribute::StrictFP);
     F.removeFnAttr("denormal-fp-math");
     F.removeFnAttr("denormal-fp-math-f32");
+    F.removeFnAttr("no-builtins");
```
pjaaskel, line 448: Pls add a comment why this is removed here so we won't forget.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -225,18 +226,9 @@ llvm::Error PoCLModulePassManager::build(std::string PoclPipeline,
   PassBuilder &PB = *PassB.get();
 
 #if 0
```
pjaaskel, line 230: Did you notice the block is still disabled?

pjaaskel (commented):
Looks promising how easily and with little updates to the compilation flow we should (finally) get this feature done. Will check closer after finished (and let's talk offline).

lib/CL/pocl_llvm_wg.cc
franz: Yes, it's supposed to be. I've tried putting the createTLII(... vectorlibrary) here before i realized it only works in codegen.

doc/sphinx/source/conformance.rst
franz: This is ancient information from 2017.  The valid current information about the setup is in the submission webpage, but i can copy some of it here.

lib/llvmopencl/Workgroup.cc
franz: done

examples/vecbuiltin/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+#   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+#   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+#   THE SOFTWARE.
+#
+#=============================================================================
+
+if(NOT ENABLE_HOST_CPU_VECTORIZE_BUILTINS)
+return()
+endif()
```
linehill, line 28:
```suggestion
if(NOT ENABLE_HOST_CPU_VECTORIZE_BUILTINS)
  return()
endif()
```

lib/CL/pocl_llvm_wg.cc
```diff
@@ -1543,10 +1537,31 @@ void pocl_llvm_free_llvm_irs(cl_program program, unsigned device_i) {
...
+                                 driver::VectorLibrary::LIBMVEC);
+#endif
+#ifdef ENABLE_HOST_CPU_VECTORIZE_SLEEF
+                                     driver::VectorLibrary::SLEEF);
+#endif
+#ifdef ENABLE_HOST_CPU_VECTORIZE_SVML
+                                     driver::VectorLibrary::SVML);
+#endif
+                                     TLIPass = new TargetLibraryInfoWrapperPass(
+                                         *TLII);
```
linehill, line 1558: Nit. Indentation.
```suggestion
#ifdef ENABLE_HOST_CPU_VECTORIZE_LIBMVEC
                                 driver::VectorLibrary::LIBMVEC);
#endif
#ifdef ENABLE_HOST_CPU_VECTORIZE_SLEEF
                                 driver::VectorLibrary::SLEEF);
#endif
#ifdef ENABLE_HOST_CPU_VECTORIZE_SVML
                                 driver::VectorLibrary::SVML);
#endif
    TLIPass = new TargetLibraryInfoWrapperPass(*TLII);
```

lib/llvmopencl/Workgroup.cc
```diff
@@ -464,7 +466,7 @@ bool WorkgroupImpl::runOnModule(Module &M, llvm::FunctionAnalysisManager &FAM) {
       F.removeFnAttr("frame-pointer");
 
     // these should be safe to enable
-    F.addFnAttr(Attribute::NoBuiltin);
+    //F.addFnAttr(Attribute::NoBuiltin);
```
linehill, line 469: Perhaps the line should be removed?

linehill (approved):
Couple style nits. Otherwise, LGTM.

pjaaskel (changes requested):
Please add FileCheck LLVM IR tests for the intrinsics that should now vectorize, starting from simple scalar invocations of the builtin from OpenCL C. Perhaps under the workgroups test or the kernel tests.

doc/sphinx/source/conformance.rst
```diff
@@ -113,146 +103,16 @@ Known issues related to CTS
...
-             tanD           4.00                      -0x1.2f000000003edp+333
-            tanh            1.18                      -0x1.ca742ap-1
-            tanhD           1.19                      0x1.f400000000395p-1
-           tanpi            4.21                      -0x1.f99d16p-3
-           tanpiD           4.09                      0x1.f6000000001d3p-3
-           trunc            0.00                      0x0p+0
-           truncD           0.00                      0x0p+0
-====================   =========================   ===========================================================
+For the sumbission, CTS was ran with the following configuration:
+OS: Ubuntu 24.04.1 LTS
```
pjaaskel, line 115: Does this text render as you expect, or get lumped to a single paragraph (in the HTML)?

cmake/LLVM.cmake
```diff
@@ -783,3 +783,45 @@ execute_process(COMMAND "${CLANG}" "--print-resource-dir" OUTPUT_VARIABLE RESOUR
 string(STRIP "${RESOURCE_DIR}" RESOURCE_DIR)
 set(CLANG_RESOURCE_DIR "${RESOURCE_DIR}" CACHE INTERNAL "Clang resource dir")
 
+##########################################################
+
+if(LLVM_VERSION_MAJOR GREATER 17)
+  # these are experimental options and may change in future
```
pjaaskel, line 789: What prevents switching on by default?

examples/vecbuiltin/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+#=============================================================================
+#   CMake build system files
+#
+#   Copyright (c) 2024 Michal Babej / Intel Finland oy
```
pjaaskel, line 4: 2025

examples/vecbuiltin/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+add_executable("vecbuiltin" vecbuiltin.c vecbuiltin_exec.c vecbuiltin.cl)
+target_link_libraries("vecbuiltin" ${POCLU_LINK_OPTIONS})
+add_symlink_to_built_opencl_dynlib("vecbuiltin")
+
+# the llvm-filecheck only works with LLVM 19
+if(LLVM_VERSION VERSION_GREATER_EQUAL 19.0)
+  add_test_pocl(NAME "examples/vecbuiltin" COMMAND "vecbuiltin" LLVM_FILECHECK "vecbuiltin.fc" WORKITEM_HANDLER "loopvec")
+else()
+  add_test_pocl(NAME "examples/vecbuiltin" COMMAND "vecbuiltin" WORKITEM_HANDLER "loopvec")
```
pjaaskel, line 40: Maybe this test is a bit pointless if FC doesn't work. The check for filecheck is inside the add_test_pocl() ?

examples/vecbuiltin/vecbuiltin.c
```diff
@@ -0,0 +1,125 @@
+/* vecbuiltin - simple invocation of a builtins on global buffer of values
+
+   Copyright (c) 2024 Michal Babej / Intel Finland Oy
```
pjaaskel, line 3: 2025

examples/vecbuiltin/vecbuiltin_exec.c
```diff
@@ -0,0 +1,84 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  int
```
pjaaskel, line 12: check indent

pjaaskel: do we need extern "C" is this for Windows?

pjaaskel (changes requested):
Small things.

doc/sphinx/source/conformance.rst
franz: Good catch, I'll change it to unordered list.

examples/vecbuiltin/CMakeLists.txt
franz: Yeah probably pointless, i'll remove the else branch.

The "working" filecheck is not the problem; the problem is that different versions of LLVM vectorize different set of builtins. The current `example/vecbuiltin` doesn't get vectorized at all by older LLVMs, so inevitably filecheck reports a failure. So....  i'd have to prepare a separate version of the test for each LLVM version, or have a single version with so little builtins that it's (hopefully) vectorized by all LLVMs. I didn't see either option as preferable, so instead i've limited the filecheck test to LLVM 19.

pjaaskel: OK. Makes sense. We should not put any effort to the older LLVMs vectorization-wise.

cmake/LLVM.cmake
franz: i'll have to add libmvec detection, otherwise nothing

examples/vecbuiltin/vecbuiltin_exec.c
franz: I think it's not strictly required, i can remove it.

cmake/LLVM.cmake
pjaaskel: OK please do. It's always better for rot-avoidance if we can enable by default, especially this useful feature.

pjaaskel:
Is there a CI enabling the support and testing this?

pjaaskel (approved):

PR #1784, opened 2025-02-03
LLVM 20 fixes

franz:


lib/CL/devices/common_utils.h
```diff
@@ -55,6 +55,7 @@ struct kernel_run_command
   kernel_run_command *prev;
   kernel_run_command *next;
   unsigned long ref_count;
+  size_t unreachable;
```
linehill, line 58: Perhaps comment what this member means/is for. By the looks of the other changes, it seems to count the times WIs reached unreachable parts of code - is that correct?

lib/llvmopencl/FixUnreachable.cc
```diff
@@ -0,0 +1,146 @@
...
+  std::cerr << "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ BEFORE fixUnreachable: \n";
+  M->dump();
+  std::cerr << "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ \n";
+#endif
+
+  SmallVector<Instruction *, 8> PendingUnreachableInst;
+  SmallVector<BasicBlock *, 8> PendingDeletableBBs;
+  for (Function::iterator I = F.begin(), E = F.end(); I != E; ++I) {
+    for (BasicBlock::iterator BI = I->begin(), BE = I->end(); BI != BE;) {
+      if (auto UI = dyn_cast<UnreachableInst>(BI++)) {
```
linehill, line 66: You could just only check the last instruction BB since `UnreachableInst`s are block terminators.

lib/llvmopencl/FixUnreachable.cc
```diff
@@ -0,0 +1,146 @@
...
+
+  for (auto BB : PendingDeletableBBs) {
+    BB->eraseFromParent();
+  }
+
+  // Find basic block with return instruction
+  BasicBlock *RetBB = nullptr;
+  for (Function::iterator I = F.begin(), E = F.end(); I != E; ++I) {
+    BasicBlock *BB = &*I;
+    if (BB->getTerminator() == nullptr)
```
linehill, line 94: This should only happen if the IR is ill-formed, I think. All BBs should end with a terminator instruction.

lib/llvmopencl/FixUnreachable.cc
pjaaskel: BB->getTerminator() should return that.

lib/llvmopencl/Barrier.h
```diff
@@ -117,12 +135,14 @@ namespace pocl {
 
     // Returns true in case the given basic block starts with a barrier,
     // that is, contains a branch instruction after possible PHI nodes.
+#if 0
```
pjaaskel, line 138: If this function is not needed, just remove it.

lib/llvmopencl/Barrier.h
```diff
@@ -85,6 +86,23 @@ namespace pocl {
         (llvm::CallInst::Create(F, "", InsertBefore));
     }
 
+#else
+    static Barrier *create(InstListType::iterator InsertBefore) {
```
pjaaskel, line 90: Do you need to duplicate the whole function? Could typedef work here?

lib/llvmopencl/ImplicitConditionalBarriers.cc
```diff
@@ -171,7 +171,11 @@ ImplicitConditionalBarriers::run(llvm::Function &F,
     // TODO: investigate. It might related to the alloca-converted
     // PHIs. It has a loop that is autoconverted to a b-loop and the
     // conditional barrier is inserted after the loop short cut check.
+#if LLVM_MAJOR < 20
     Barrier::create(Pos->getFirstNonPHI());
+#else
+    Barrier::create(Pos->getFirstInsertionPt());
```
pjaaskel, line 177: Is this correct? It might insert before PHI?

lib/llvmopencl/ImplicitLoopBarriers.cc
```diff
@@ -116,8 +116,13 @@ static bool addInnerLoopBarrier(llvm::Loop &L,
 
     /* Add a barrier both to the beginning of the entry and to the very end
        to nicely isolate the parallel region. */
+#if LLVM_MAJOR < 20
     Barrier::create(brexit->getTerminator());
     Barrier::create(loopEntry->getFirstNonPHI());
+#else
+    Barrier::create(brexit->getTerminator()->getIterator());
+    Barrier::create(loopEntry->getFirstInsertionPt());
```
pjaaskel, line 124: Same here (Phi).

lib/llvmopencl/Kernel.cc
```diff
@@ -276,7 +280,12 @@ Kernel::getParallelRegions(llvm::LoopInfo &LI) {
 void Kernel::addLocalSizeInitCode(size_t LocalSizeX, size_t LocalSizeY,
                                   size_t LocalSizeZ) {
 
+#if LLVM_MAJOR < 20
   IRBuilder<> Builder(getEntryBlock().getFirstNonPHI());
+#else
+  IRBuilder<> Builder{getEntryBlock().getContext()};
+  Builder.SetInsertPoint(getEntryBlock().getFirstInsertionPt());
```
pjaaskel, line 287: And here.

lib/llvmopencl/ImplicitLoopBarriers.cc
```diff
@@ -116,8 +116,13 @@ static bool addInnerLoopBarrier(llvm::Loop &L,
 
     /* Add a barrier both to the beginning of the entry and to the very end
        to nicely isolate the parallel region. */
+#if LLVM_MAJOR < 20
     Barrier::create(brexit->getTerminator());
     Barrier::create(loopEntry->getFirstNonPHI());
+#else
+    Barrier::create(brexit->getTerminator()->getIterator());
```
pjaaskel, line 123: Looking at these. Why not move the getIterator() call inside Barrier::create and avoid a lot of these ifdefs?

include/pocl_context.h
```diff
@@ -43,6 +43,7 @@ struct pocl_context32 {
   uint printf_buffer_capacity;
   uint global_var_buffer;
   uint work_dim;
+  uint unreachable;
```
pjaaskel, line 46: Please add a comment as it's not obvious to outsiders what this is and how it is used.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -508,6 +508,8 @@ static void addStage1PassesToPipeline(cl_device_id Dev,
   // this must be done AFTER inlining, see note above
   addPass(Passes, "automatic-locals", PassType::Module);
 
+  addPass(Passes, "fix-unreachable");
```
pjaaskel, line 511: Let's avoid adding new small passes as I've just got rid of those in my private branch. Can you convert this to a helper function called from WILoops or similar? See the new DeSPMD/WILoops in my branch.

lib/llvmopencl/FixUnreachable.cc
```diff
@@ -0,0 +1,146 @@
...
+#define PASS_CLASS pocl::FixUnreachable
+#define PASS_DESC "Fix unreachable instruction uses to defined behavior"
+
+// #define DEBUG_FIX_UNREACHABLE
+
+namespace pocl {
+
+using namespace llvm;
+
+static bool fixUnreachable(Function &F) {
```
pjaaskel, line 52: "Fix" is not obvious here, please add a comment what, why and how is "fixed".

lib/llvmopencl/FixUnreachable.cc
```diff
@@ -0,0 +1,146 @@
...
+
+  Module *M = F.getParent();
+
+#ifdef DEBUG_FIX_UNREACHABLE
+  std::cerr << "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ BEFORE fixUnreachable: \n";
```
pjaaskel, line 57: Perhaps try to mimic the formatting the previous pass printouts. We should go towards a unified one with the printouts. Preferably to how dbgs() are done in LLVM since this code will be submitted to LLVM eventually.

pjaaskel (changes requested):

lib/llvmopencl/ImplicitConditionalBarriers.cc
linehill: It's fine: "Returns an iterator to the first instruction in this block that is suitable for inserting a non-PHI instruction."

lib/CL/pocl_llvm_wg.cc
franz: > Can you convert this to a helper function called from WILoops or similar?

AFAICT that's not possible. Different parts of this "handling" of unreachables need to be done in different points of the pipeline. The first part (where functions with unreachable are marked inlined) must be done before AlwaysInline pass, the second part ("fixUnreachable" - where all unreachable instructions are removed) must be run before ImplicitConditionalBarriers; so AFAICT no, this cannot be done from WILoops.

All of the work of this pass is done in a single function "fixUnreachable", so in your private branch, should be easy to convert this pass to a function call in the right place.

lib/llvmopencl/ImplicitConditionalBarriers.cc
franz: .... "In particular, it skips all PHIs and LandingPad instructions"

It seems (to me) more correct (in theory) to use `getFirstInsertionPt()`, but since OpenCL does not support exceptions, there shouldn't be any LandingPad instructions, so it should be identical to `getFirstNonPhi()`.

lib/llvmopencl/FixUnreachable.cc
franz: fixed

lib/llvmopencl/Barrier.h
franz: minimized the diff

lib/llvmopencl/FixUnreachable.cc
franz: fixed

lib/llvmopencl/FixUnreachable.cc
franz: converted to assert()

lib/CL/devices/common_utils.h
franz: renamed the flag & added a comment

lib/llvmopencl/FixUnreachable.cc
franz: renamed & added comment on the Pass in the file's header

include/pocl_context.h
franz: renamed the flag & added a comment

lib/CL/devices/pthread/pthread_scheduler.c
```diff
@@ -336,12 +337,12 @@ work_group_scheduler (kernel_run_command *k,
   pocl_free_kernel_arg_array_with_locals ((void **)arguments,
                                           (void **)arguments2, k);
 
-  return 1;
+  POCL_ATOMIC_ADD (k->execution_failed, execution_failed);
```
linehill, line 340: Could this addition overflow and, by a very unlikely chance, turn an execution failure to be recorded as successful one?

lib/llvmopencl/UnreachablesToReturns.cc
```diff
@@ -0,0 +1,140 @@
...
+      LLVM_DEBUG(dbgs() << "UNREACHABLE found: replacing Inst in "
+                        << F.getName().str() << "\n");
+#endif
+      // this can happen when inlining functions which have unreachable Inst
+      // we end up with a BB with 0 predecessors and a single unreachable
+      if (BB.hasNPredecessors(0)) {
+        PendingDeletableBBs.push_back(&BB);
+      } else {
+        PendingUnreachableInst.push_back(UI);
+      }
```
linehill, line 79: LLVM style nit:
```suggestion
      if (BB.hasNPredecessors(0))
        PendingDeletableBBs.push_back(&BB);
      else
        PendingUnreachableInst.push_back(UI);
```

lib/llvmopencl/UnreachablesToReturns.cc
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  for (auto BB : PendingDeletableBBs) {
+    BB->eraseFromParent();
+  }
```
linehill, line 85: Style nit:
```suggestion
  for (auto BB : PendingDeletableBBs)
    BB->eraseFromParent();
```

lib/llvmopencl/UnreachablesToReturns.cc
```diff
@@ -0,0 +1,140 @@
...
+    Builder.CreateStore(ConstOne, UnreachGV);
+    if (RetBB)
+      Builder.CreateBr(RetBB);
+    else
+      Builder.CreateRetVoid();
+  }
+
+  for (auto UI : PendingUnreachableInst) {
+    UI->eraseFromParent();
+  }
```
linehill, line 121: Style nit:
```suggestion
  for (auto UI : PendingUnreachableInst)
    UI->eraseFromParent();
```

lib/llvmopencl/Workgroup.cc
```diff
@@ -1020,6 +1035,57 @@ void WorkgroupImpl::privatizeGlobals(
...
+    for (BasicBlock::iterator ii = i->begin(), ee = i->end(), Next = ii;
+         ii != ee; ii = Next) {
+      Next = std::next(ii);
+      for (size_t j = 0; j < GlobalHandleNames.size(); ++j) {
+        if (PrivatePointers[j].first == nullptr) {
+          continue;
+        }
+        if (!isa<llvm::StoreInst>(ii)) {
+          continue;
+        }
```
linehill, line 1055: Style nit:
```suggestion
        if (PrivatePointers[j].first == nullptr)
          continue;
        if (!isa<llvm::StoreInst>(ii))
          continue;
```

pjaaskel:
I have a clang-tidy script in my private branch. I'll upstream it and we can add this file to its list of files to process.

pjaaskel:
Can you also add an LLVM 20 CI with this PR?

franz:
> I have a clang-tidy script in my private branch. I'll upstream it and we can add this file to its list of files to process.

I did run the `tools/scripts/format-last-commit.sh` which runs `clang-format-diff`. Is clang-tidy better ? IIRC it takes forever to run, but if it's better, it would be useful.

lib/CL/devices/pthread/pthread_scheduler.c
franz: Yes in theory it could overflow. I'll change it to atomic_or

lib/llvmopencl/SubCFGFormation.cc
```diff
@@ -267,10 +283,17 @@ getLocalSizeValues(llvm::Function &F, llvm::ArrayRef<unsigned long> LocalSizes,
 
       if (I->getParent() != &F.getEntryBlock()) {
         // must be in entry block. move.
+// TODO for some reason, moveAfter(InsertionPoint) exists in the header, but
```
pjaaskel, line 286: Leftover?

lib/llvmopencl/UnreachablesToReturns.cc
```diff
@@ -0,0 +1,140 @@
...
+  Module *M = F.getParent();
+
+  SmallVector<Instruction *, 8> PendingUnreachableInst;
+  SmallVector<BasicBlock *, 8> PendingDeletableBBs;
+  for (Function::iterator I = F.begin(), E = F.end(); I != E; ++I) {
+    BasicBlock &BB = *I;
+    assert(BB.getTerminator());
+    if (auto UI = dyn_cast<UnreachableInst>(BB.getTerminator())) {
+#ifdef DEBUG_CONVERT_UNREACHABLE
+      LLVM_DEBUG(dbgs() << "UNREACHABLE found: replacing Inst in "
```
pjaaskel, line 70: Started to think about this mechanism. Is it possible to enable debug outputs only from this pass easily without needing to dump all LLVM debugs which is quite a flood? Do we have a mechanism for that?

lib/llvmopencl/SubCFGFormation.cc
franz: "leftover" as in, an actual remaining unsolved problem. For now it works fine with the old code, because passing of Instruction instead of InsertionPoint is only deprecated (not removed), but they will likely remove it in future LLVM version.

lib/llvmopencl/UnreachablesToReturns.cc
franz: There is a LLVM option "-debug-only". See `lib/CL/pocl_llvm_utils.cc` lines 541-545 or so.

lib/llvmopencl/ImplicitLoopBarriers.cc
franz: fixed the ifdefs

pjaaskel (approved):
looks great

PR #1790, opened 2025-02-06
Better logic for the ID built-in expansion decision

pjaaskel:
* Convert the built-ins to declarations after first linking them in to handle all cases similarly.  Previously, if a built-in called an ID built-in, it was linked in as a defined one and the logic of excluding them didn't work.
* Also add get_num_groups() to the expanded built-ins to avoid yet one more source of messy switch..cases.

### DIFF
--- a/lib/llvmopencl/LLVMUtils.cc
+++ b/lib/llvmopencl/LLVMUtils.cc
@@ -726,10 +726,16 @@ llvm::Type *SizeT(llvm::Module *M) {
   return IntegerType::get(M->getContext(), AddressBits);
 }
 
-bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(const llvm::Function &F) {
-  Instruction::const_use_iterator UI = F.use_begin(), UE = F.use_end();
-  for (; UI != UE; ++UI) {
-    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
+bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(
+    const llvm::Function &F) {
+
+  if (F.getName() != GID_BUILTIN_NAME && F.getName() != GS_BUILTIN_NAME &&
+      F.getName() != GROUP_ID_BUILTIN_NAME && F.getName() != LID_BUILTIN_NAME &&
+      F.getName() != LS_BUILTIN_NAME && F.getName() != NGROUPS_BUILTIN_NAME)
+    return false;
+
+  for (const auto &U : F.uses()) {
+    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(U.getUser());
     if (Call == nullptr)
       continue;
     if (!isCompilerExpandableWIFunctionCall(*Call))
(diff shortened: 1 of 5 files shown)


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -726,10 +726,16 @@ llvm::Type *SizeT(llvm::Module *M) {
...
+bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(
+    const llvm::Function &F) {
+
+  if (F.getName() != GID_BUILTIN_NAME && F.getName() != GS_BUILTIN_NAME &&
+      F.getName() != GROUP_ID_BUILTIN_NAME && F.getName() != LID_BUILTIN_NAME &&
+      F.getName() != LS_BUILTIN_NAME && F.getName() != NGROUPS_BUILTIN_NAME)
+    return false;
+
+  for (const auto &U : F.uses()) {
+    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(U.getUser());
```
linehill, line 738: My two cents. The type is spelled in the dyn_cast:
```suggestion
    auto *Call = dyn_cast<llvm::CallInst>(U.getUser());
```

lib/llvmopencl/LLVMUtils.cc
```diff
@@ -726,10 +726,16 @@ llvm::Type *SizeT(llvm::Module *M) {
...
-  Instruction::const_use_iterator UI = F.use_begin(), UE = F.use_end();
-  for (; UI != UE; ++UI) {
-    llvm::CallInst *Call = dyn_cast<llvm::CallInst>(UI->getUser());
+bool isWorkitemFunctionWithOnlyCompilerExpandableCalls(
+    const llvm::Function &F) {
+
+  if (F.getName() != GID_BUILTIN_NAME && F.getName() != GS_BUILTIN_NAME &&
+      F.getName() != GROUP_ID_BUILTIN_NAME && F.getName() != LID_BUILTIN_NAME &&
+      F.getName() != LS_BUILTIN_NAME && F.getName() != NGROUPS_BUILTIN_NAME)
+    return false;
```
linehill, line 735: My two cents:
```suggestion
  if (!CompilerExpandableCalls.count(F.getName()))
    return false;
```
Where the `CompilerExpandableCalls` is a static object of a set.

linehill (approved):
Couple suggestions. Othwerwise, LGTM.

lib/llvmopencl/LLVMUtils.cc
pjaaskel: Yep, I think using auto should not be automatic (for readability), but mainly help with more complex types. But more of a matter of taste, I guess.

lib/llvmopencl/LLVMUtils.cc
pjaaskel: Yep, not in my taste :)

pjaaskel:
LLVM 18 tier1 bot fails:
```
 get_linear_ids...
PTHREAD ERROR in pocl_update_event_finished():2119: Invalid argument (22)
```
I cannot repro locally with LLVM 19. Building LLVM 18.

PR #1791, opened 2025-02-07
Misc fixes

anbe42:
add a test for a llvm segfault
distinguish X32 from I386


CMakeLists.txt
```diff
@@ -434,7 +434,9 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "riscv")
   endif()
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(i.86|AMD64|x86_64|amd64)")
   set(X86 1)
-  if(HOST_DEVICE_ADDRESS_BITS MATCHES "32")
+  if(CMAKE_LIBRARY_ARCHITECTURE STREQUAL "x86_64-linux-gnux32")
+    set(X32 1)
```
linehill, line 438: What is the effect of setting this variable? I'm not seeing or finding anything that could act on this variable.

anbe42: `X32` itself is not used anywhere (and `X86_64` does not seem to be used anywhere either), but the effect is `I386` no longer getting set in this case, and therefore some i386 specific code paths no longer being used on x32

tests/regression/test_issue_1608.cpp
```diff
@@ -0,0 +1,63 @@
...
+    for (int n = 0; n < 2; ++n)
+      C[m] = mad(B[m], f, C[m]);
+}
+__kernel void krnl(int g) {
+  fn2(g, 1);
+}
+)RAW";
+
+int main(int argc, char *argv[]) {
+  int err = EXIT_SUCCESS;
```
linehill, line 36: `cl_int`for portability (`cl_int` might not be equivalent to `int`) and some OpenCL error for avoiding false positives.
```suggestion
  cl_int err = CL_INVALID_VALUE;
```

CMakeLists.txt
anbe42: this change is sufficient to make PoCL  build on x32, but there are still some failing tests, but that is something I want to investigate later (after switching the Debian package to 6.1 and llvm-19)

tests/regression/test_issue_1608.cpp
anbe42: there is `int err = program.getInfo<>(CL_PROGRAM_BINARIES, &binaries);` in a few tests that you might want to fix, too

pjaaskel:
LLVM 19 - chipStar:

```
The following tests FAILED:
	521 - Unit_hipHostRegister_Negative - int (Subprocess aborted)
	522 - Unit_hipHostRegister_Negative - float (Subprocess aborted)
	523 - Unit_hipHostRegister_Negative - double (Subprocess aborted)
Errors while running CTest
```
I cannot see how the failure could be due to this PR. Flakiness?

franz:
@pjaaskel  nope, this an actual problem somewhere (but not in this PR). Also reproduces in my PR #1784.

pjaaskel:
@anbe42 can you rebase this now so we get the CIs happy?

tests/regression/CMakeLists.txt
```diff
@@ -84,6 +85,8 @@ add_test_pocl(NAME "regression/test_issue_1435" COMMAND "test_issue_1435")
 
 add_test_pocl(NAME "regression/test_issue_1525" COMMAND "test_issue_1525" LLVM_FILECHECK "test_issue_1525.fc" WORKITEM_HANDLER "loopvec")
 
+add_test_pocl(NAME "regression/test_issue_1608" COMMAND "test_issue_1608")
```
franz, line 88: Does this test now pass with all supported LLVM versions with the CPU device ?

Because the CI machines for CPU-device only run tests labeled with "cpu" label (which is a subset of all tests). This test does not have any label, hence it won't be picked up by any CI as it is now.

franz (changes requested):

tests/regression/CMakeLists.txt
franz: label can be added later.

franz (approved):

PR #1795, opened 2025-02-12
minor fixes + LevelZero command buffers

franz:


lib/CL/clEnqueueCommandBufferKHR.c
```diff
@@ -115,14 +115,39 @@ POname (clEnqueueCommandBufferKHR) (cl_uint num_queues,
...
+      /* TODO: figure out a way to handle migrations effectively. We can't
+       * change the cmd buffer once it's finished recording, as changing would
+       * also require changing the hardware resource (e.g. L0 command list),
+       * which would negate the point of using cmdbufs in the 1st place. */
+      _cl_command_node *node = NULL;
+      cl_event final_ev;
+      errcode = pocl_create_command (
+        &node, used_queues[0], CL_COMMAND_COMMAND_BUFFER_KHR, &final_ev,
+        num_events_in_wait_list, event_wait_list, NULL);
+      // pocl_deep_copy_migration_info_list (cmd->migr_infos, 0));
```
pjaaskel, line 132: Is this omitted by accident? Can be removed or at least commented to a C comment?

lib/CL/clUpdateMutableCommandsKHR.c
```diff
@@ -1,6 +1,6 @@
-/* OpenCL runtime library: clCommandBarrierWithWaitListKHR()
+/* OpenCL runtime library: clUpdateMutableCommandsKHR()
 
-   Copyright (c) 2022 Jan Solanti / Tampere University
+   Copyright (c) 2024 Michal Babej / Intel Finland Oy
```
pjaaskel, line 3: 2025 already, yeah, it's scifi.

lib/CL/devices/level0/level0-driver.cc
```diff
@@ -2932,6 +2932,7 @@ Level0Device::Level0Device(Level0Driver *Drv, ze_device_handle_t DeviceH,
   if (ClDev->atomic_memory_capabilities & CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES)
     OpenCL30Features.append(" __opencl_c_atomic_scope_all_devices");
 
+#ifndef ENABLE_CONFORMANCE
```
pjaaskel, line 2962: What's missing from conformance with CBufs? Could be documented somewhere?

lib/CL/devices/level0/pocl-level0.cc
```diff
@@ -702,7 +702,8 @@ int pocl_level0_supports_binary(cl_device_id Device, size_t Length,
 
   if (Device->compiler_available == CL_TRUE &&
       Device->linker_available == CL_TRUE && Device->num_ils_with_version > 0 &&
-      pocl_bitcode_is_spirv_execmodel_kernel(Binary, Length) != 0) {
+      pocl_bitcode_is_spirv_execmodel_kernel(Binary, Length,
+                                             Device->address_bits) != 0) {
```
pjaaskel, line 706: Do we need != 0 here?

lib/CL/pocl_util.c
```diff
@@ -2285,33 +2294,62 @@ bitcode_is_spirv_execmodel (const char *bitcode, size_t size, uint32_t type)
...
   uint32_t value, instruction;
+  /* every opcapab is followed by one value */
   instruction = htole32 (bc32[location++]);
   value = htole32 (bc32[location++]);
   while (instruction == OpCapab && location < (size / 4))
     {
-      if (value == type)
-        return 1;
+      if (value == exec_model)
+        {
```
pjaaskel, line 2307: Are the curly brackets needed?

lib/CL/devices/level0/level0-driver.cc
```diff
@@ -684,28 +701,38 @@ void *Level0Queue::createCommandBuffer(cl_command_buffer_khr CmdBuf) {
...
 }
 
 void Level0Queue::freeCommandBuffer(void *CmdBufPtr) {
   assert(CmdBufPtr);
   Level0CmdBufferData *CmdBufData = (Level0CmdBufferData *)CmdBufPtr;
-  while (!CmdBufData->Events.empty()) {
-    auto E = CmdBufData->Events.front();
-    CmdBufData->Events.pop();
-    AvailableDeviceEvents.push(E);
+  {
```
pjaaskel, line 727: Do we need the separate code block here for the lock guard? Could it be function scope?

poclu/poclu.h
```diff
@@ -369,6 +369,14 @@ int poclu_load_program_multidev (cl_platform_id platform,
  */
 POCLU_API cl_int POCLU_CALL poclu_show_program_build_log (cl_program program);
 
+/**
+ * \brief check if a device supports a particular version of SPIR-V
+ * @param il [in] a string like "SPIR-V_1.2" of the requested version
```
pjaaskel, line 374: We try to converge to the LLVM doxygen conventions. Pls use \param (or \p) and \return in new comments. No need for the \brief tag, but an empty comment line after it instead.

tests/kernel/kernel.c
```diff
@@ -59,8 +59,16 @@ int call_test(const char *name)
...
 
+  char kernel_name[1024];
   for (cl_uint i = 0; i < num_kernels; i++)
     {
+      size_t name_len = 0;
+      err = clGetKernelInfo (kernels[i], CL_KERNEL_FUNCTION_NAME, 1024,
+                             kernel_name, &name_len);
+      CHECK_OPENCL_ERROR_IN ("clGetKernelInfo failed");
+      kernel_name[name_len] = 0;
+      if (name_len && strstr (kernel_name, "DONT_RUN"))
```
pjaaskel, line 70: Not sure of this commit. Why not use volatile instead?

pjaaskel (approved):
Small things. Can be fixed in the next PR if there are no other changes to come for this PR.

lib/CL/clUpdateMutableCommandsKHR.c
franz: Yes, but the file content was created in Oct-Nov of 2024. So maybe 2024,2025 is the correct one ? It's unclear to me.

lib/CL/devices/level0/level0-driver.cc
franz: I think it can be enabled, nothing should be missing

lib/CL/devices/level0/pocl-level0.cc
franz: no we don't

lib/CL/pocl_util.c
franz: nope, will fix

lib/CL/devices/level0/level0-driver.cc
franz: No, it can't be function scope. The delete operator deletes the CmdBuf object ; we can't be holding a lock that's inside the object, because the lock would be unlocked at function exit (after delete), creating a write-after-free.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -1549,7 +1549,7 @@ static TargetLibraryInfoImpl *initPassManagerForCodeGen(legacy::PassManager &PM,
   TargetLibraryInfoImpl *TLII = nullptr;
 
 #ifdef ENABLE_HOST_CPU_VECTORIZE_BUILTINS
-  if (DevTriple.isX86()) {
+  if (Device->type == CL_DEVICE_TYPE_CPU) {
```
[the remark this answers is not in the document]
franz: I don't see how it would be better; have you seen the larger context of this code ? This particular piece of code applies *only* to CodeGen; the vectorization should be attempted much earlier (in the default LLVM optimization passes we use), via Loop/SLPVectorizer. To my understanding, this code should not in any way inhibit or enable vectorization; this only instructs the CPU target backend to lower already vectorized builtin calls to particular library calls. So AFAICT it does not make sense to change the condition here to `!Device->spmd`, because here we're creating a TLII with one of three libraries, and all three are CPU only (AFAIK).

lib/CL/devices/level0/pocl-level0.cc
```diff
@@ -1901,3 +1915,28 @@ cl_int pocl_level0_set_kernel_exec_info_ext(
...
 }
+
+cl_int
+pocl_level0_create_finalized_command_buffer(cl_device_id Dev,
+                                            cl_command_buffer_khr CmdBuf) {
+  Level0Device *Device = (Level0Device *)Dev->data;
+  CmdBuf->data[Dev->dev_id] = nullptr;
+  void *CmdBufData = Device->createCmdBuf(CmdBuf);
+  POCL_RETURN_ERROR_ON(
+      (CmdBufData == nullptr), CL_INVALID_VALUE,
```
linehill, line 1926: Is this the error code to be returned from `clFinalizeCommandBufferKHR` call? `CL_INVALID_VALUE` is not listed as one of the possible error codes. Perhaps the error code should be `CL_OUT_OF_RESOURCES`?

franz: fixed

lib/CL/clEnqueueCommandBufferKHR.c
franz: Converted to `#if 0`

lib/CL/clUpdateMutableCommandsKHR.c
pjaaskel: Yep,  or a range 2024-2025

lib/CL/pocl_llvm_wg.cc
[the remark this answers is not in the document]
pjaaskel: "CPU only" :  what do you count as "CPU" here? The host CPU only? For non-host CPUs/accelerators we would/will port a selected LLVM-supported vecbiltin library to the whatever SIMD-supporting target which then implements it with its intrinsics/asm or lowers vector datatypes (however the builtin lib in question is implemented). Just like a cross-compiler and the rest of the toolchain is ported to a target.

tests/kernel/kernel.c
franz: fixed

PR #1816, opened 2025-02-28
Small kernel compiler and test infra updates

pjaaskel:
* Fix dirty address space casts: Convert address space casts through integer casts to proper address space cast instructions.
* Add a new directory for kcompiler unit tests which use LLVM IR output inspection only.
* UVA: Consider all kinds of constant data uniform
* run_kernel: allow explicit output buffers for easier output data generation and inspection without forcing "inplace"-semantic.


lib/llvmopencl/Workgroup.cc
```diff
@@ -195,6 +195,52 @@ class WorkgroupImpl {
...
+static bool cleanupAddressSpaceCasts(Function &F) {
+  bool Changed = false;
+  for (auto &BB : F) {
+    for (BasicBlock::iterator BI = BB.begin(), BE = BB.end(); BI != BE;) {
+
+      // The casts look like this:
+      // %55 = ptrtoint ptr addrspace(1) %7 to i64
+      // %56 = inttoptr i64 %55 to ptr addrspace(4)
+
+      PtrToIntInst *PtrToInt = dyn_cast_or_null<PtrToIntInst>(&*BI);
```
linehill, line 215:
```suggestion
      PtrToIntInst *PtrToInt = dyn_cast<PtrToIntInst>(&*BI);
```
Unless the cast may receive `nullptr` values which probably is not the case.

lib/llvmopencl/Workgroup.cc
```diff
@@ -195,6 +195,52 @@ class WorkgroupImpl {
...
+      if (PtrToInt != nullptr &&
+          (PtrToInt->getPointerAddressSpace() == SPIR_ADDRESS_SPACE_GLOBAL ||
+           PtrToInt->getPointerAddressSpace() == SPIR_ADDRESS_SPACE_GENERIC) &&
+          PtrToInt->hasOneUse()) {
+        IntToPtrInst *User =
+            dyn_cast_or_null<IntToPtrInst>(*PtrToInt->users().begin());
```
linehill, line 221:
```suggestion
        IntToPtrInst *User = dyn_cast<IntToPtrInst>(*PtrToInt->users().begin());
```

tools/scripts/run-and-check-llvm-ir
```diff
@@ -38,16 +38,21 @@ CHECK_FILE=$3
...
 
+export POCL_KERNEL_CACHE=1
 export POCL_CACHE_DIR=$(mktemp -d /tmp/pocl-lit-test-XXXXXXXX)
-
 export POCL_LEAVE_KERNEL_COMPILER_TEMP_FILES=1
 
-$APP_CMD
+$APP_CMD || exit 1
+
+BC_FILES=`find $POCL_CACHE_DIR -name parallel.bc`
```
linehill, line 47: For consistency (there is already `$()` form in use in the above) and modernization:
```suggestion
BC_FILES=$(find $POCL_CACHE_DIR -name parallel.bc)
```

tests/workgroup/run_kernel.c
```diff
@@ -110,30 +110,55 @@ main (int argc, char **argv)
 
   size_t grid_size
       = global_work_size[0] * global_work_size[1] * global_work_size[2];
-  if (num_args == 1)
+  if (num_args > 0)
     {
```
linehill, line 114: Does this branch accept arbitrary amount of arguments or just up to two? I see there is an assert in the else branch for checking there is up to two arguments but no check for 2+ arguments in this branch.

lib/llvmopencl/Workgroup.cc
```diff
@@ -195,6 +195,52 @@ class WorkgroupImpl {
...
+        if (User == nullptr ||
+            (User->getAddressSpace() != SPIR_ADDRESS_SPACE_GLOBAL &&
+             User->getAddressSpace() != SPIR_ADDRESS_SPACE_GENERIC))
+          break;
+        llvm::IRBuilder<> Builder(F.getContext());
```
linehill, line 226: Repeated IRBuilder construction - hoist it out of the loop?

lib/llvmopencl/VariableUniformityAnalysis.cc
```diff
@@ -279,7 +279,9 @@ bool VariableUniformityAnalysisResult::isUniform(llvm::Function *F,
     return true;
   }
 
-  if (isa<llvm::ConstantInt>(V)) {
+  // We cannot check for Constant here as the magic id variables are external
+  // constants.
+  if (isa<llvm::ConstantData>(V)) {
     setUniform(F, V, true);
```
linehill, line 285: Do you mean `GlobalValue`s by external constants? Why they can't be marked uniform? They are constants and therefore they must be uniform (their address not their contained value).

lib/llvmopencl/Workgroup.cc
```diff
@@ -195,6 +195,53 @@ class WorkgroupImpl {
...
+
+      // The casts look like this:
+      // %55 = ptrtoint ptr addrspace(1) %7 to i64
+      // %56 = inttoptr i64 %55 to ptr addrspace(4)
+
+      PtrToIntInst *PtrToInt = dyn_cast<PtrToIntInst>(&*BI);
+      if (PtrToInt != nullptr &&
+          (PtrToInt->getPointerAddressSpace() == SPIR_ADDRESS_SPACE_GLOBAL ||
+           PtrToInt->getPointerAddressSpace() == SPIR_ADDRESS_SPACE_GENERIC) &&
+          PtrToInt->hasOneUse()) {
```
linehill, line 221: This pass potentially misses the following cast pattern:
```
%1 = ptrtoint ptr addrspace(1) 0 to i64
%2 = inttoptr i64 %1 to ptr addrspace(4)
%3 = inttoptr i64 %1 to ptr addrspace(4)
```

linehill:
On windows:
```
Unhandled exception at 0x00007FF7112370C7 in run_kernel.exe: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x00000099B6C03000).
```
Backtrace points to a line in run_kernel.c with `alloca` call. Probably running out of stack space - consider replacing the `alloca` with heap memory.

lib/llvmopencl/VariableUniformityAnalysis.cc
pjaaskel: Yes, in general, but not in this case because the ID pseudo variables are modeled as global variables as this point, and can thus diverge.

lib/llvmopencl/Workgroup.cc
pjaaskel: Yes. Have you seen it in real codes?

lib/llvmopencl/VariableUniformityAnalysis.cc
linehill: > ID pseudo variables are modeled as global variables as this point, and can thus diverge.

You can still mark them uniform. What should be marked possibly divergent are loads from them. 

```
@some_global_var = external global i32  ;  @some_global_var is uniform.

%1 = load i32, ptr @some_global_var  ; %1 could be divergent.
```

lib/llvmopencl/Workgroup.cc
linehill: > Have you seen it in real codes?

Does it matter? Answering no here doesn't rule out the possibility. We can agree they are unlikely, however. If you are not going to cover this pattern you should at least add a comment about the caveat and put in some alarm (like add `assert(PtrToInt->hasOneUse());`) catch the case.

lib/llvmopencl/VariableUniformityAnalysis.cc
pjaaskel: Yes, that should work. Currently it's recursively done on operands until the leaf and this would add a special case at load handling.  I'll check if it can be done without too intrusive changes.

lib/llvmopencl/Workgroup.cc
pjaaskel: This peephole is for a specific pattern that comes from the CUDA/HIP frontend, not meant to add complexity for the sake of covering unseen cases.  But you are right, a comment and assert should be good.

linehill (approved):

pjaaskel:
The ballot crash should be fixed by https://github.com/pocl/pocl/pull/1840.

PR #1831, opened 2025-03-09
Enable SPIR-V 1.3 with LLVM v19

pjaaskel:
This brings in the subgroup ballot support which chipStar can use.

This exposes an issue in TestBallot (at least for me):

```
/home/pjaaskel/src/chipStar/build-llvm-19/tests/runtime/TestBallot
TestBallot: /home/pjaaskel/src/chipStar/llvm-project/llvm/lib/Transforms/Utils/ValueMapper.cpp:978: void {anonymous}::Mapper::remapInstruction(llvm::Instruction*): Assertion `(Flags & RF_IgnoreMissingLocals) && "Referenced value not in value map!"' failed.
...
(gdb) call I
$1 = (llvm::Instruction *) 0x55556656e5f0
(gdb) call I->dump()
  %22 = load i64, ptr %_local_size_x, align 8

```
Let's see if the CI also fails.

### DIFF
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -1781,16 +1781,21 @@ pocl_init_default_device_infos (cl_device_id dev,
     dev->llvm_target_triplet = "";
 #endif
 
-#if defined(ENABLE_SPIRV) && LLVM_MAJOR >= 20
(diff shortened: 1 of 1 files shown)


franz:
> Nope. Is the LLVM built without asserts?

It is using Ubuntu's LLVM, so most likely yes.

PR #1840, opened 2025-03-13
Fix a regression with subgroup size

pjaaskel:
This was caused by the new privatizeGlobalStores() call setting the insertion pointer back to the beginning of the function.  The special case for SGSize which sets it to local_size_x generated a load from it before the alloca was initialized.

Also clean up the ballot implementation a bit + add a regression test case for it which also tests SGSize in the basic test suite.


tests/workgroup/CMakeLists.txt
tneval, line 1: Is the test case left out on purpose?

tests/workgroup/CMakeLists.txt
```diff
@@ -141,6 +141,10 @@ add_test_pocl(NAME "workgroup/range_md_large_grid"
               EXPECTED_OUTPUT ""
               COMMAND "run_kernel" "range_md.cl" "1000" "128" "1" "1")
 
+add_test_pocl(NAME "workgroup/ballot"
```
franz, line 144: Like Tapio mentioned, ballot.cl is missing, and also please add this to one of the `set_tests_properties(... LABELS "cpu;workgroup")` code blocks.

franz (changes requested):

lib/kernel/subgroups.cl
```diff
@@ -114,19 +114,26 @@ get_first_llid (void)
...
+
   sub_group_barrier (CLK_LOCAL_MEM_FENCE);
   if (get_sub_group_local_id () == 0)
     {
       flags[get_sub_group_id ()] = 0;
-      uint *f = (uint *)(flags + get_sub_group_id ());
-      for (uint i = 0; i < get_sub_group_size () && i < 128; ++i)
+
+      /* The subgroup's uint4 where to flip up the ballot results. */
+      uint *f = (uint *)&flags[get_sub_group_id ()];
```
tneval, line 132:
```suggestion
      /* The subgroup's uint4 where to flip up the ballot results. */
      uint *subgroup_flags = (uint *)&flags[get_sub_group_id ()];
```
Maybe something more descriptive?

tneval (changes requested):
Good to go after adding the missing test case.

franz (approved):

linehill:
FYI, this patch fixes the issue in windows CIs:
```patch
diff --git a/lib/kernel/subgroups.cl b/lib/kernel/subgroups.cl
index ba83fe113..2bd3ffc9a 100644
--- a/lib/kernel/subgroups.cl
+++ b/lib/kernel/subgroups.cl
@@ -33,17 +33,9 @@
 /* Magic variable that is expanded in Workgroup.cc */
 extern uint _pocl_sub_group_size;

-#if _MSC_VER
-/* I wonder if there's a better way to agree on the mangling? */
-#define get_local_id (DIM) __identifier ("?get_local_id@@$$J0YAKI@Z") (DIM)
-#define get_local_linear_id  __identifier ("?get_global_linear_id@@$$J0YAKXZ")
-#define get_local_size (DIM) __identifier ("?get_local_size@@$$J0YAKI@Z") (DIM)
-
-#else
 size_t _CL_OVERLOADABLE get_local_id (unsigned int dimindx);
 size_t _CL_OVERLOADABLE get_local_linear_id (void);
 size_t _CL_OVERLOADABLE get_local_size (unsigned int dimindx);
-#endif

 void _CL_OVERLOADABLE _CL_CONVERGENT work_group_barrier(cl_mem_fence_flags);
 void _CL_OVERLOADABLE
```

The manual mangling is only needed on the C sources (mangling of overloaded functions differ by one character `_` in C vs OpenCL C when targeting a triple with `msvc` component in it).

PR #1841, opened 2025-03-13
WG: Remove the original kernel if not needed any longer

pjaaskel:
This should at speed up at least codegen and dlopen() somewhat as it typically splits the LLVM Module size to about half.


pjaaskel:
The ballot crash could be due to the SGSize issue that is fixed in #1840. I'm still wondering why it hasn't failed before.

tneval (approved):
Good improvement, it also makes analyzing the final bitcode a bit easier.

tneval (commented):
Maybe try again after pulling in #1840.

PR #1842, opened 2025-03-13
pocl_types.h redefined reserve_id_t

pjaaskel:
It seems at least with LLVM 18 it will be defined in the Clang's header (at least when CLC=2), causing a conflict.


pjaaskel:
This problem was exposed by MIVisionX. I'll need to recheck this when I get back to it.

pjaaskel:
It's gone, so likely some local issue I had.

PR #1844, opened 2025-03-14
Bugfixes

franz:


lib/llvmopencl/linker.cpp
```diff
@@ -262,46 +264,65 @@ static void fixCallingConv(llvm::Module *Mod, std::string &Log) {
...
-        FNameSet.insert(Callee->getName());
-        DB_PRINT("search: recursing into %s\n", Name);
-        find_called_functions(Callee, FNameSet);
-      }
+  CallStack.insert(F);
+
+  assert(F->hasName());
+  std::string FName = F->getName().str();
+
+  for (auto &I : instructions(F)) {
```
pjaaskel, line 287: Good to know there's this type of an iterator-generator.

tests/regression/test_issue_1826.cpp
```diff
@@ -0,0 +1,40 @@
+#define CL_HPP_ENABLE_EXCEPTIONS
```
pjaaskel, line 1: The license/copyright header is missing.

pjaaskel (changes requested):

lib/llvmopencl/linker.cpp
franz: Yep this pattern (iterating over all instructions of a function) is so common they've made an iterator for it

tests/regression/test_issue_1826.cpp
franz: fixed

.github/workflows/build_full_cts.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        llvm: [17]
+        llvm: [19]
```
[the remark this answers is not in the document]
pjaaskel: @franz is something breaking with LLVM 20? I think we should test with as new LLVM as possible.

pjaaskel (approved):

PR #1846, opened 2025-03-17
Patches riscv64 platform support, verified on banana-pi-f3 with Bianbu 2.1 (Codename: noble) 

soccercheng:
This pull request contains the necessary patches for  enabling POCL for  riscv64 platform support, It has been verified on  cwith Bianbu 2.1 (Codename: noble).

There are also [WiKi pages](https://github.com/soccercheng/pocl-riscv/wiki) that describe the steps on how to x-build POCL for target device, banana-pi-f3 with Bianbu 2.1, on x86-64 machine.


cmake/bitcode_rules.cmake
```diff
@@ -189,9 +189,13 @@ function(compile_ll_to_bc FILENAME SUBDIR BCLIST)
       endif()
     endif()
 
+    if(RISCV64)
+      set(EXTRA_OPT2 "--data-layout=e-m:e-p:64:64-i64:64-i128:128-n32:64-S128")
```
pjaaskel, line 193: Why is this needed?

cmake/LLVM.cmake
```diff
@@ -809,8 +809,13 @@ endif()
 if(ENABLE_HOST_CPU_DEVICES AND NOT DEFINED CLANG_SUPPORTS_FLOAT16_ON_CPU)
   set(CLANG_SUPPORTS_FLOAT16_ON_CPU 0)
   message(STATUS "Checking Device-side (Clang/LLVM) support for _Float16 type")
-    custom_try_compile_clang_silent("_Float16 callfp16(_Float16 a) { return a * 1.8f16; };" "_Float16 x=callfp16((_Float16)argc);"
-    RESV --target=${LLC_TRIPLE} ${CLANG_MARCH_FLAG}${SELECTED_HOST_CPU})
+    if(RISCV)
+      custom_try_compile_clang_silent("_Float16 callfp16(_Float16 a) { return a * 1.8f16; };" "_Float16 x=callfp16((_Float16)argc);"
+                                      RESV --target=${LLC_TRIPLE} -mcpu=${SELECTED_HOST_CPU} -march=${CLANG_MARCH_FLAG})
```
pjaaskel, line 814: Is this right? CLANG_MARCH_FLAG should be '-march' or similar, not the content of the flag right? Perhaps the detection logic for these flags should be fixed better and then this if(RISCV) would not be needed?

lib/CL/pocl_llvm.h
```diff
@@ -42,6 +42,17 @@ extern "C" {
   POCL_EXPORT
   const char *pocl_get_llvm_cpu_abi ();
 
+#if defined(__riscv)
```
pjaaskel, line 45: Why are these needed for RISC-V but not for other archs?

lib/CL/pocl_llvm_utils.cc
```diff
@@ -174,6 +178,118 @@ const char *pocl_get_llvm_cpu_abi() {
   return nullptr;
 }
 
+#if defined(__riscv)
+
+static char *__model_name = nullptr;
+static char *__isa = nullptr;
+
+static void query_cpu()    {
```
pjaaskel, line 186: Can you heck if you can reuse&improve lib/CL/devices/cpuinfo.c instead so we avoid duplicated code? This function doesn't seem to belong to this file which is for LLVM-related utils.

lib/CL/pocl_llvm_utils.cc
```diff
@@ -174,6 +178,118 @@ const char *pocl_get_llvm_cpu_abi() {
...
+
+const char *pocl_get_llvm_cpu_model()    {
+  query_cpu();
+  return __model_name;
+}
+
+void pocl_get_llvm_cpu_feature(const char *march, char *features, int length)    {
+  std::string targetFeatures = "";
+
+  if(nullptr == march)    {
```
pjaaskel, line 260: This is not common style in PoCL, please use the more "logically readable" march == nullptr (although I've been told what the point of this way around is).

lib/CL/devices/common.c
```diff
@@ -1769,7 +1769,11 @@ pocl_init_default_device_infos (cl_device_id dev,
   dev->kernellib_name = strdup(kernellib);
   if (dev->kernellib_subdir == NULL)
     dev->kernellib_subdir = "host";
-  dev->llvm_abi = pocl_get_llvm_cpu_abi ();
+  dev->llvm_abi = pocl_get_llvm_cpu_abi();
```
pjaaskel, line 1782: Check the C style which is GNU. The clang-format-diff.sh script should fix the basic formatting for you.

lib/CL/pocl_llvm_build.cc
```diff
@@ -91,7 +95,7 @@ POP_COMPILER_DIAGS
 #include "ProgramScopeVariables.h"
 #include "linker.h"
 
-//#define DEBUG_POCL_LLVM_API
+#define DEBUG_POCL_LLVM_API
```
pjaaskel, line 98: Forgot?

lib/CL/pocl_llvm_build.cc
```diff
@@ -520,6 +524,14 @@ int pocl_llvm_build_program(cl_program program,
   if (device->llvm_abi != NULL)
     ss << "-target-abi " << device->llvm_abi << " ";
 
+#if defined(__riscv)
+  if (device->llvm_arch != NULL)    {
+    char features[4096] = { '\0' };
+    pocl_get_llvm_cpu_feature(device->llvm_arch, features, 4096);
```
pjaaskel, line 530: This would be better called only once. Also these device-driver-specific flags should be added in the init_build() hook in the CPU driver. Let's try reduce driver-specific codes in this file which should be quite generic (to all LLVM targets).

lib/CL/pocl_llvm_build.cc
```diff
@@ -688,9 +700,42 @@ int pocl_llvm_build_program(cl_program program,
...
+
+      token = strtok_r(NULL, ",", &saveptr);
+    }
+
+    ta.Features.push_back("experimental");
+    ta.FeatureMap["experimental"] = true;
+  }
 #endif
+
+#if defined(DEBUG_POCL_LLVM_API) && 0
```
pjaaskel, line 728: You can drop && 0. This seems like a useful debug output.

pjaaskel (changes requested):
Thanks! Looks much more readable now. Some minor comments. Also please check that the style adheres to the style guide. There are formatter scripts to help here.

cmake/bitcode_rules.cmake
soccercheng: This because, during "Built target sleef_config_generic-rv64", Compiler complains about the following:

[  0%] Built target sleef_config_generic-rv64
[  0%] Building LL to LLVM bitcode .../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/barrier.ll.bc
[  1%] Building LL to LLVM bitcode .../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/addrspace_operators.ll.bc
[  1%] Building LL to LLVM bitcode .../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/addrspace_operators_spir.ll.bc
[  1%] Linking part of builtins library
warning: Linking two modules of different data layouts: '.../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/svm_atomics_host.cl.bc' is 'e-m:e-p:64:64-i64:64-i128:128-n32:64-S128' whereas 'llvm-link' is 'e-m:e-i64:64-f80:128-n8:16:32:64-S128'

warning: Linking two modules of different data layouts: '.../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/svm_atomics.cl.bc' is 'e-m:e-p:64:64-i64:64-i128:128-n32:64-S128' whereas 'llvm-link' is 'e-m:e-i64:64-f80:128-n8:16:32:64-S128'

warning: Linking two modules of different data layouts: '.../pocl-riscv/build.banana-pi-BPI-F3/lib/kernel/host/generic-rv64/sleef.bc' is 'e-m:e-p:64:64-i64:64-i128:128-n32:64-S128' whereas 'llvm-link' is 'e-m:e-i64:64-f80:128-n8:16:32:64-S128'

cmake/LLVM.cmake
soccercheng: I realize that CLANG_MARCH_FLAG should look like '-march=xxx000', but don't you think "--target=${LLC_TRIPLE} -mcpu=${SELECTED_HOST_CPU} -march=${CLANG_MARCH_FLAG}" looks more unified ?

in current Risc-V world, clang/llvm knows only limited # of vendor CPUs, so developers always use "-mcpu=generic-rv64", accompanied with the right "-march=xxx_000" for specifying the set of extensions that a CPU supports.

So, "${CLANG_MARCH_FLAG}${SELECTED_HOST_CPU}" might NOT be the proper for Risc-V platform.

lib/CL/pocl_llvm.h
soccercheng: It is a little complicated for this, cause it seems, in current POCL implementation and even LLVM, it seems there are no clearly unified definitions/usages for arch/features for all platform, including x86. arm, aarch64 and also Risc-V.  

As for this purpose, my suggestion is to add "llvm_arch" and "llvm_targetmachine_features" to cl_device_id and to query (@ runtime) the right target machine's arch and features in pocl_init_default_device_infos accordingly.

struct _cl_device_id {
  ....
  const char* llvm_cpu; /* the llvm CPU variant to use */
  const char *llvm_abi; /* the ABI to use */
  const char *llvm_arch; /* the ARCH to use, -march */
  const char *llvm_targetmachine_features; /* the LLVM TargetMachine features to use */
  ....
};

cmake/LLVM.cmake
franz: > I realize that CLANG_MARCH_FLAG should look like '-march=xxx000',

With the current code, it doesn't look like that. That CMake variable has *only* the name of the flag (up until '=' character); not including the value. And i don't see any CMake code in this PR that would change the CLANG_MARCH_FLAG (perhaps you forgot to add it?)

So when CMake substitutes strings, this string:

`--target=${LLC_TRIPLE} -mcpu=${SELECTED_HOST_CPU} -march=${CLANG_MARCH_FLAG}`

would become something like:

`--target=<something> -mcpu=<something> -march=-march=`

.. which doesn't make sense.

> So, "${CLANG_MARCH_FLAG}${SELECTED_HOST_CPU}" might NOT be the proper for Risc-V platform. 

That may be true, but the ` -march=${CLANG_MARCH_FLAG}` part still doesn't make sense.

soccercheng: There is a Wiki page [(X‐Build Pocl on Build Machine)](https://github.com/soccercheng/pocl-riscv/wiki/X%E2%80%90Build-Pocl-on-Build-Machine) describing how I configure POCL for X-Build. and "CLANG_MARCH_FLAG" is provide there.

```
export  RISCV_ISA=....
cmake -S .. -B . -G "Unix Makefiles"                   \
    ...
    -DLLC_HOST_CPU=generic-rv64                          \
    -DLLC_HOST_CUSTOM_CPU=spacemit_k1                    \
    -DHOST_CPU_TARGET_ABI=lp64d                          \
    -DCLANG_MARCH_FLAG=${RISCV_ISA}      
```

what I ment by " looking more unified" is that, maybe, just maybe, we can unifind the format we provide FLAG settings by providing the content of the flag only....

--target=<content-only> -mcpu=<content-only> -march=<content-only>

lib/CL/pocl_llvm_utils.cc
soccercheng: Good suggestion and I'll change the code accordingly.

BUT, I've foulnd in current pocl_cpu_init_common (lib/CL/devices/common_utils.c), pocl_init_default_device_infos is called before pocl_cpuinfo_detect_device_info, if we detect LLVM configuratins, in pocl_init_default_device_infos, that depend on the native CPU features collected in pocl_cpuinfo_detect_device_info,  this will be an issue.

PR #1849, opened 2025-03-19
Fix a crash with POCL_TRACING=cq

pjaaskel:
Also widened the kernel name column width.

### DIFF
--- a/tests/regression/CMakeLists.txt
+++ b/tests/regression/CMakeLists.txt
@@ -107,7 +107,8 @@ endif()
 
 add_test_pocl(NAME "regression/test_llvm_segfault_issue_889" COMMAND "test_llvm_segfault_issue_889")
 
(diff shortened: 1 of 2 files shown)


tests/regression/CMakeLists.txt
```diff
@@ -107,7 +107,8 @@ endif()
 
 add_test_pocl(NAME "regression/test_llvm_segfault_issue_889" COMMAND "test_llvm_segfault_issue_889")
 
-add_test_pocl(NAME "regression/test_issue_1711" COMMAND "test_issue_1711")
+# Also smoke test cq-tracing with this case.
```
franz, line 110: Just a nitpick (can be fixed in another next PR): maybe a few words on why this case in particular

franz (approved):

PR #1850, opened 2025-03-20
add retain event during event sync (PR  #1719 + some fixes)

franz:


lib/CL/devices/tbb/tbb.c
```diff
@@ -286,13 +286,28 @@ void pocl_tbb_notify(cl_device_id device, cl_event event, cl_event finished) {
   cl_bool wake_thread = CL_FALSE;
   _cl_command_node *node = event->command;
 
-  if (finished->status < CL_COMPLETE) {
-      pocl_update_event_failed_locked (event);
-      return;
+  if (finished->status < CL_COMPLETE)
+  {
+    /* Unlock the finished event in order to prevent a lock order violation
```
pjaaskel, line 291: I see this part duplicated and I start to think that could we refactor some of the common event handling to the PoCL core side to simplify the drivers and avoid the duplication?

pjaaskel (commented):
It's good to get this in. I only worry that is this a bit late in the release cycle: Are we convinced this fixes races/deadlocks and doesn't introduce new ones. The duplication of event handling code to driver is also a concern here.

franz:
> Are we convinced this fixes races/deadlocks and doesn't introduce new ones

Some of the CIs have been reporting random failures for some time now, and it does seem related to event handling, so i think it's worth a try. There are tests in the OpenCL-CTS that use multithreaded processing heavily, some of those are ran in the CI workflow - in reduced mode, but I can run the full version.

> The duplication of event handling code to driver is also a concern here.

Yes but if you look closer, the duplication of this PR is 70% comments. The actual code is 3 lines / driver. There is more duplication of event-related `device-ops` in general, but i think that needs a separate PR.

PR #1851, opened 2025-03-20
Fix a regression with subgroup size (retake)

linehill:
Retake of #1840 with a patch to fix mangling issues.


pjaaskel (approved):
Thanks Henry for fixing the Windows side!

PR #1876, opened 2025-04-07
CPU: indirect SVM/USM was not set

pjaaskel:
Setting of the indirect pointer info was omitted, likely due to not considering multi-device cases. This was not a problem with CPU-only platforms, but affects migration of indirect-access buffers residing on other devices' memories.

I have additional tests upcoming in a later PR.

### DIFF
--- a/lib/CL/pocl_mem_management.c
+++ b/lib/CL/pocl_mem_management.c
@@ -1105,3 +1105,41 @@ pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id)
   ptr += arg->offset;
   return (void *)ptr;
 }
+
+void
+pocl_reset_indirect_ptrs (cl_kernel kernel, void **ptrs, size_t n)
+{
+  if (kernel->indirect_raw_ptrs != NULL)
+    {
+      struct _pocl_ptr_list_node *n, *tmp;
+      DL_FOREACH_SAFE (kernel->indirect_raw_ptrs, n, tmp)
+        {
+          free (n);
+        }
+      kernel->indirect_raw_ptrs = NULL;
+    }
+
+  for (size_t i = 0; i < n; ++i)
+    {
+      void *ptr = ptrs[i];
+      if (ptr == NULL)
+        continue;
+
+      /* Filter out non-sensical pointers silently. The spec doesn't
+         tell to return an error in this case, perhaps for future
+         compatibility with system allocated (but not migrated)
+         buffers? */
+      pocl_raw_ptr *svm_ptr
+        = pocl_find_raw_ptr_with_vm_ptr (kernel->context, ptr);
+
+      if (svm_ptr == NULL)
+        continue;
+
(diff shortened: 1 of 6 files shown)


lib/CL/pocl_mem_management.c
```diff
@@ -1105,3 +1105,26 @@ pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id)
...
+          free (n);
+        }
+      kernel->indirect_raw_ptrs = NULL;
+    }
+
+  for (size_t i = 0; i < n; ++i)
+    {
+      struct _pocl_ptr_list_node *node
+        = malloc (sizeof (struct _pocl_ptr_list_node));
+      node->ptr = ptrs[i];
```
linehill, line 1126: Should the pointers be checked for NULLs or invalid ones (pointers that are not members of any OpenCL allocations) and filtered out?

pjaaskel: Yes, we could, but not sure if we _should_, the spec doesn't say it explicitly: https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clSetKernelExecInfo.html

pjaaskel: Added a middle-ground solution where I just filter out NULL or unknown pointers for now.

lib/CL/pocl_mem_management.c
```diff
@@ -1105,3 +1105,41 @@ pocl_cpu_get_ptr (struct pocl_argument *arg, unsigned global_mem_id)
   ptr += arg->offset;
   return (void *)ptr;
 }
+
+void
```
linehill, line 1109: For MSVC toolchain you'll need add `POCL_EXPORT` here too.

pjaaskel: All CIs passed. Why didn't MSVC scream here?

linehill: Probably there is no callers from outside libpocl right now so that's why you won't hear any screaming.

pjaaskel: Right. Actually I'd say the POCL_EXPORT is extra here as well since we currently call from libpocl.so only and not from the driver plugins.

pjaaskel:
> However, a thing came up to my mind: What should be done for pointers in `indirect_raw_ptrs` which are freed after? Is there a danger they are dereferenced?

I think that's an app error. It means you SVMFree and after that launch kernels that use the freed buffer, right?

linehill:
> ... It means you SVMFree and after that launch kernels that use the freed buffer, right?

That but the kernels may not dereference them. I recalled a case in chipStar where all SVM pointer are annotated with CL_KERNEL_EXEC_INFO_SVM_PTRS (because it can't tell well which allocations might be accessed the kernels) and in multi-threaded cases lead to this situation where one thread freed an allocation but another enqueued a kernel with the allocation still being annotated but not dereferenced and this caused issues (crashes).

PR #1883, opened 2025-04-09
prerelease fixes 2

franz:


pjaaskel (changes requested):
Can you check if you can use the regex package of C++ to parse that version a bit better/future proof manner? Otherwise good small fixes from coverity mostly.

lib/CL/pocl_llvm_spirv.cc
```diff
@@ -104,6 +104,20 @@ static void handleInOutPathArgs(bool &keepPath, char *Path, char *HiddenPath,
...
                                       cl_uint device_i);
 
+static bool getMaxSpirvVersion(pocl_version_t &MaxVersion,
+                               const char *DevVersions) {
+  MaxVersion.major = 1;
+  MaxVersion.minor = 0;
+  if (DevVersions) {
+    std::string SpirvVersions{DevVersions};
+    if (SpirvVersions.find("SPIR-V_1.") == 0) {
+      MaxVersion.minor = SpirvVersions[9] - '0';
```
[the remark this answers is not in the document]
franz: I only see 1.6 in released versions.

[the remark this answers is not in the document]
franz: fixed

pocld/request_th.cc
```diff
@@ -101,9 +101,14 @@ void RequestQueueThread::readThread() {
       continue;
 
     Request *IncomingRequest = new Request();
+    if (IncomingRequest == nullptr) {
```
linehill, line 104: The `new` operator doesn't return `nullptr` (an exception is thrown instead) but `new(std::nothrow)` does.

linehill, line 104: The `new` operator doesn't return `nullptr` (an exception is thrown instead) but `new(std::nothrow)` does.
```suggestion
    Request *IncomingRequest = new(std::nothrow) Request();
    if (IncomingRequest == nullptr) {
```

linehill (changes requested):
One request. After that, this PR is good for landing.

lib/CL/pocl_llvm_wg.cc
```diff
@@ -762,6 +762,8 @@ void *pocl_llvm_create_context_for_program(char *ProgramBcContent,
   // parse the program's bytes into a llvm::Module
   if (P == nullptr ||
       !P->init(ProgramBcContent, ProgramBcSize, TempBitcodePath)) {
+    if (P)
```
pjaaskel, line 765: Please use the explicit comparison style P != nullptr for pointer checks (just for style uniformity).

pocld/request.hh
```diff
@@ -56,48 +56,48 @@ private:
 class Request {
 public:
   /// Size, in bytes, of the main request body (up to sizeof RequestMsg_t)
-  uint32_t BodySize;
+  uint32_t BodySize = 0;
```
pjaaskel, line 59: It might be best to {} initialize in the instantiator side than rely on these member variable initializers.  https://en.cppreference.com/w/cpp/language/zero_initialization

pjaaskel (approved):
Small things that can be fixed post-pull unless you will push a new version anyhow.

linehill (approved):
LGTM on my side.

PR #1900, opened 2025-04-25
Add environment variables to override vector width

Pennycook:
The changes in this pull request introduce two new environment variables:

- **POCL_FORCE_VECTOR_WIDTH**
  This is equivalent to LLVM's `force-vector-width` option, and forces the LLVM loop vectorizer to use the same vector width for all loops.

- **POCL_PREFER_VECTOR_WIDTH**
  This is equivalent to clang's `prefer-vector-width` option, and overrides clang's preference for 256- or 512-bit vectors on certain x86 targets (e.g., processors previously codenamed Skylake).

It might be a little confusing that one of these environment variables expects the width in loop iterations while the other expects the width in terms of bits.  However, that's how `force-vector-width` and `prefer-vector-width` work, and it would probably be even more confusing to try and come up with some PoCL-specific convention that didn't match LLVM.  I've tried to draw attention to this discrepancy in the documentation.

These changes were motivated by an application I've been working with, where overriding LLVM's heuristics with `POCL_FORCE_VECTOR_WIDTH=16` and `POCL_PREFER_VECTOR_WIDTH=512` results in a 2x speedup.


pjaaskel:
Thanks! Could you add an LLVM IR check case where vectorization to the given length occurs (only) if the env is given?

pjaaskel:
Also a mention in releasenotes would be nice. If you get this done within the next week or so we can pull it to 7.0.

doc/sphinx/source/using.rst
```diff
@@ -407,6 +407,29 @@ pocl.
  When set to 1, prints out remarks produced by the loop vectorizer of LLVM
  during kernel compilation.
 
+- **POCL_FORCE_VECTOR_WIDTH**
+
+ Forces the LLVM loop vectorizer to use the specified vector width, overriding
```
linehill, line 412: Is the "vector width" measured in number of elements or bytes (since the other env variable measures in bits)? Perhaps this should be clarified.

Pennycook: It's a number of loop iterations.  I've tried to clarify that here: ff3299db3391a3aaa9327ed5f5c485eee6667f00

doc/sphinx/source/using.rst
```diff
@@ -407,6 +407,31 @@ pocl.
...
+ by the vectorizer's cost model.
+ The same vector width will be used by all loops in all kernels.
+ Setting the vector width to 1 disables vectorization.
+ If the requested vector width is higher than the machine's native vector
+ width, the vectorizer will also unroll the loop.
+
+- **POCL_PREFER_VECTOR_WIDTH**
+
+ Override the preferred vector width (expressed as a number of **bits**) for
+ x86 targets.
```
pjaaskel, line 423: Is this x86-specific or might it work for other targets such as ARM or RISC-V?

pjaaskel (approved):
LGTM. Please do a rebase instead of a merge commit.

doc/sphinx/source/using.rst
Pennycook: My understanding is that it's x86-specific.  I've only ever seen the prefer-vector-width=256 attribute being set for x86 targets, and only for the subset of x86 architectures where the clang/gcc community decided that 256-bit was a better default.

franz:
I have only one objection: since these variables are very LLVM specific, I suggest we rename them to `POCL_LLVM_FORCE_VECTOR_WIDTH` and `POCL_LLVM_PREFER_VECTOR_WIDTH`.

The current names make it seem like these variables 1) somehow affect the `CL_DEVICE_PREFERRED_VECTOR_WIDTH` properties of `clGetDeviceInfo`, and 2) apply to all devices supported by PoCL - neither of which is true, which IMO makes them somewhat misleading.

pjaaskel:
> I have only one objection: since these variables are very LLVM specific, I suggest we rename them to `POCL_LLVM_FORCE_VECTOR_WIDTH` and `POCL_LLVM_PREFER_VECTOR_WIDTH`.
> 
> The current names make it seem like these variables 1) somehow affect the `CL_DEVICE_PREFERRED_VECTOR_WIDTH` properties of `clGetDeviceInfo`, and 2) apply to all devices supported by PoCL - neither of which is true, which IMO makes them somewhat misleading.

Or how about POCL_VECTORIZER_{PREFER,FORCE}_VECTOR_WIDTH, as there's also POCL_VECTORIZER_REMARKS? We could start a "subcategory" of settings that affect the vectorization process.

Pennycook:
@pjaaskel - Should I basically ignore all the tests that passed, here?  I can't really tell if the tests I added are _always_ failing and some tests are irrelevant, or whether the tests I added sometimes pass.

pjaaskel:
> @pjaaskel - Should I basically ignore all the tests that passed, here? I can't really tell if the tests I added are _always_ failing and some tests are irrelevant, or whether the tests I added sometimes pass.

Some of the CIs do not run the internal tests, this is a bit confusing (@franz is fixing this). Seems there's some genuine problem there in the failing ones as it doesn't find your check string.

Pennycook:
> Some of the CIs do not run the internal tests, this is a bit confusing (@franz is fixing this). Seems there's some genuine problem there in the failing ones as it doesn't find your check string.

Hm.  Okay, I'll take a look at this tomorrow.  I don't understand why it seems to be accepting the options but silently ignoring them.  I would have expected things to blow up more spectacularly than this.

...and of course, things run on my machine. 😆

Pennycook:
I think I fixed the basic tests: the issue wasn't the tests themselves, but that the environment variable wasn't being set correctly by the testing infrastructure. The new failures don't seem to be related to anything I've changed.

Let me know what you decide regarding the renaming and I'll make that change.

pjaaskel:
Please use POCL_VECTORIZER_{PREFER,FORCE}_VECTOR_WIDTH, perhaps a bit more descriptive and we can expand with other vectorizer-affecting properties, some of those properties might be implemented in LLVM upstream, some in our downstream PoCL-passes, so I think adding 'LLVM' to the name is not good.

Pennycook:
> Please use POCL_VECTORIZER_{PREFER,FORCE}_VECTOR_WIDTH, perhaps a bit more descriptive and we can expand with other vectorizer-affecting properties, some of those properties might be implemented in LLVM upstream, some in our downstream PoCL-passes, so I think adding 'LLVM' to the name is not good.

Makes sense to me.  I've changed the names in https://github.com/pocl/pocl/pull/1900/commits/fa75cc19e583105d29866cad59f1fde370984cc6.

Pennycook:
It doesn't look like the failing test is related to anything I've changed.

PR #1903, opened 2025-04-25
 doc: advertise server-side command buffers in 6.1 relnotes

jansol:
DOI is not live yet, let's hope that changes before release.


pjaaskel:
Can you send this against 'main'? release_7_0 is not created yet and release_6_1 is obsolete (we decided to tag it v7.0 afterall).

PR #1904, opened 2025-04-28
relnotes: link to server-side cmdbuf paper

jansol:
Looks like the 7.0 release notes already had a mention of server side command buffers. We'll probably still want to advertise the paper there, too, though.


franz:
"DOI not found" - link not active yet ?

jansol:
Yeah, it's assigned but not published yet.

PR #1909, opened 2025-04-30
Remat: Robustness for AllocaInst handling

pjaaskel:
Sometimes when rematerializing an alloca, the initializer could fail to remat and we detected that too late. Now detects remat-ability early to ensure the rematerialization can be successful.

Also removed the Phi branch as we should not encounter them at this point any longer.

I have tests that trigger without this in the private branch which will be upstreamed Soon.


franz (approved):
LGTM

PR #1920, opened 2025-05-16
Add PowerShell script for code formatting

linehill:


tools/scripts/format-diff.ps1
```diff
@@ -0,0 +1,17 @@
...
+#
+# The former formats staged changes and the latter formats changes since
+# the REF commit.
+#
+# Note that unstaged changes will not be formatted.
+
+git-clang-format.bat `
+  --extensions h,c,cl --style "file:$PSScriptRoot\style.GNU" @args
+git-clang-format.bat `
+  --extensions hh,hpp,cc,cpp --style "file:$PSScriptRoot\style.CPP" @args
```
jansol, line 17: Looking at the bash version of the script, the extensions should be ignored for anything under `lib/llvmopencl` and `lib/CL/devices/tce` and the CPP style should be used for everything there.

jansol (changes requested):

tools/scripts/format-diff.ps1
linehill: Good catch, I'll work this out (may take time).

linehill: Updated to apply the CPP style on files under `lib/llvmopencl` and `lib/CL/devices/tce`.

PR #1926, opened 2025-05-21
The workaround for reduced fast flags not needed for LLVM 18+

pjaaskel:
721733862ad2 introduced a commit to not pass -cl-fast-relaxed-math directly to Clang as it caused to inaccurate divisions in some cases. However, the passed flags do not the introduce essential relaxation attribute 'reassoc' which is required for generating vectorized reduction trees from float reduction loops.

The rounding problems seem to be fixed starting LLVM 18, thus we can disable the workaround for newer LLVMs.

### DIFF
--- a/lib/CL/pocl_llvm_build.cc
+++ b/lib/CL/pocl_llvm_build.cc
@@ -413,7 +413,7 @@ int pocl_llvm_build_program(cl_program program,
   size_t fastmath_flag = user_options.find("-cl-fast-relaxed-math");
 
   if (fastmath_flag != std::string::npos) {
-#ifdef ENABLE_CONFORMANCE
+#if defined(ENABLE_CONFORMANCE) && LLVM_VERSION_MAJOR < 18
     user_options.replace(fastmath_flag, 21,
                          "-cl-finite-math-only -cl-unsafe-math-optimizations");
 #endif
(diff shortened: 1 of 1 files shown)


lib/CL/pocl_llvm_build.cc
```diff
@@ -413,7 +413,7 @@ int pocl_llvm_build_program(cl_program program,
   size_t fastmath_flag = user_options.find("-cl-fast-relaxed-math");
 
   if (fastmath_flag != std::string::npos) {
-#ifdef ENABLE_CONFORMANCE
+#if defined(ENABLE_CONFORMANCE) && LLVM_VERSION_MAJOR < 18
```
franz, line 416: just a consistency nitpick: the macro we use in most of PoCL code is `LLVM_MAJOR`

franz (commented):
LGTM (one minor nitpick)

lib/CL/pocl_llvm_build.cc
pjaaskel: OK. I can fix it in a future commit to avoid retesting for this?

pjaaskel: Maybe we should convert all of those LLVM_VERSION_MAJOR references in a single commit.

franz: Sure, no problem

franz (approved):

PR #1928, opened 2025-05-25
Allow for -DENABLE_ICD=OFF -DRENAME_POCL=ON

vchuravy:
For a deployment situation I would like to both disable ICD support as well as renaming the POCL symbols.
I only want to access the pthread devices and want to avoid any symbol collisions with other OpenCL libraries
that may be loaded.

### DIFF
--- a/lib/CL/pocl_cl.h
+++ b/lib/CL/pocl_cl.h
@@ -279,8 +279,14 @@ extern pocl_obj_id_t last_object_id;
 
 #  define POname(name) PO##name
 
+#if defined(RENAME_POCL) && !defined(BUILD_ICD)
+#define POdeclsym(name) POCL_EXPORT __typeof__ (name) PO##name;
+#define POdeclsymExport(name) POdeclsym(name)
+#else
 #define POdeclsym(name) __typeof__ (name) PO##name;
 #define POdeclsymExport(name) POCL_EXPORT POdeclsym(name)
+#endif
+
 #  define POCL_ALIAS_OPENCL_SYMBOL(name)                                \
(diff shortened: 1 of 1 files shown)


vchuravy:
The logic split in https://github.com/vchuravy/pocl/blob/02103927ba899b4bcb60ef02d3eedb4a6644cd50/lib/CL/pocl_cl.h#L260-L276 is a bit unfortunate here since it means that `-DRENAME_POCL` has no effect and leads to a failure like:

```
[16:41:41] ld64.lld: error: undefined symbol: POclGetDeviceInfo
[16:41:41] >>> referenced by poclu/libpoclu.a(misc.c.o):(symbol poclu_get_multiple_devices+0x278)
[16:41:41] >>> referenced by poclu/libpoclu.a(misc.c.o):(symbol poclu_supports_opencl_30+0xdc)
[16:41:41] >>> referenced by poclu/libpoclu.a(misc.c.o):(symbol poclu_supports_opencl_30+0x50)
[16:41:41] >>> referenced 12 more times
[16:41:41] 
[16:41:41] ld64.lld: error: undefined symbol: POclGetProgramBuildInfo
[16:41:41] >>> referenced by poclu/libpoclu.a(misc.c.o):(symbol poclu_show_program_build_log+0x114)
[16:41:41] >>> referenced by poclu/libpoclu.a(misc.c.o):(symbol poclu_show_program_build_log+0xe0)
```

pjaaskel (approved):
The CIs pass so I think it's safe to pull in. Thanks!

vchuravy:
Thanks! This only makes it work for Linux, figuring out MacOs and Windows is still something I need to do.

PR #1934, opened 2025-05-29
add CPU label by default in add_test_pocl + fix the remat test output 

pjaaskel:
...this is to avoid forgetting to add the label when adding new tests, like I seem to tend to do.

'cpu' label means that the test is expected to pass with host CPUs. Add a new label 'xfail' that should be added to tests
that are known broken at the moment.

I don't think we need test-type labels (kernel, workgroup) since the tests have the type in their name and can be
filtered by it.  What do you think?

Also fix test_rematerialized_alloca_load_with_outside_pr_users. Its validation was dependent on device-specific output strings. Appeared in #1931.


franz:
> 'cpu' label means that the test is expected to pass with host CPUs. Add a new label 'xfail' that should be added to tests
that are known broken at the moment.

I think "xfail" should be renamed to something like "cpu_fail" to make it clear the test fails on the CPU device (other devices could still pass the test). If the test itself is broken (= fails with all devices) then we should either disable it with `DISABLED` property, fix it, or  remove the test entirely.

Also I think one label per device is enough: ctest can include or exclude tests by label, hence there's no need to have both positive (cpu) and negative (cpu_fail) labels for the same device. Since CPU driver passes 99.9% of tests, it makes sense to use only the negative label.

> I don't think we need test-type labels (kernel, workgroup) since the tests have the type in their name and can be
filtered by it.

Yeah i agree this "subgrouping" of tests isn't too useful. The CI only uses it only in one instance (ASan sanitizer) which we can change to regexp match.

.github/workflows/build_linux.yml
```diff
@@ -165,9 +165,9 @@ jobs:
           elif [ "${{ matrix.config }}" == "asan" ]; then
             runCTest -L "runtime|asan"
```
franz: For Asan, this need changing to: `runCTest -R runtime` ... or we can try running all cpu tests with just `runCTest`. Since the Asan+Conformance build now [passes](https://github.com/pocl/pocl/actions/runs/15436252536) all "cts_micro" tests, there shouldn't be many failures with the internal testsuite either.

.github/workflows/build_linux.yml
```diff
@@ -165,9 +165,9 @@ jobs:
...
           elif [ "${{ matrix.config }}" == "tsan" ]; then
-            runCTest -L "cpu|tsan"
+            runCTest -L "tsan"
```
franz: Same as Asan, both Tsan and Ubsan need changing to just plain `runCTest` (= run same tests as default CPU).

If you look at the current sanitizer CI runs, they have zero tests:
https://github.com/pocl/pocl/actions/runs/15469075013/job/43548428766?pr=1934

(TBH it's a bit annoying that CTest does not return error on zero tests selected)

edit: Ok it seems 3.26 added a new ctest option `--no-tests=<X>` and environment variable `CTEST_NO_TESTS_ACTION`. We should set the env var to "error" in the runner script.

.github/workflows/build_linux_san.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         run: |
           runCTest() {
             # limit jobs to 8; ctest seems to have some memleak or something, it eats up 10+ GBytes of memory
```
franz: you can remove this line about memleak, it's been solved

cmake/add_test_pocl.cmake
```diff
@@ -34,12 +34,22 @@ include(CMakeParseArguments)
...
 # will be added that runs the test with the LLVM IR tester script using the
-# loopvec method."
-
+# loopvec method.
+#
+# If ONLY_FILECHECK is set to 1, the test is only added as an LLVM IR filecheck
+# which runs the program and validates the parallel.bc IR. Otherwrise, if LLVM_FILECHECK
+# is given, the execution test is added also separately.
+#
+# By default the function adds the 'cpu' label to the test, meaning that the
```
franz: I think this paragraph is now outdated ?

tests/regression/CMakeLists.txt
```diff
@@ -83,7 +89,11 @@ add_test_pocl(NAME "regression/test_issue_757" COMMAND "test_issue_757")
 
 add_test_pocl(NAME "regression/test_issue_1435" COMMAND "test_issue_1435")
 
-add_test_pocl(NAME "regression/test_issue_1525" COMMAND "test_issue_1525" LLVM_FILECHECK "test_issue_1525.fc" WORKITEM_HANDLER "loopvec")
+# fails on windows/MinGW with "access violation"
+add_test_pocl(NAME "regression/test_issue_1525" LABELS "mingw_fail win_fail"
```
franz: IIRC the LABELS property is supposed to be a CMake list type, so ";" delimited not space delimited

franz (changes requested):
Looking good mostly, minor fixes needed

franz (approved):

PR #1942, opened 2025-06-04
Improve VM/BDA pointer lookup

linehill:
This reduced time of `clEnqueueNDRangeKernel()` calls from ~20us to few microseconds in OpenVINO which allocates lots of USM allocations (~700 for resnet50 inference).

Additionally,

* Require pocl_raw_ptr::device to be set if pocl_raw_ptr::dev_ptr is non-NULL. The new pointer lookup uses it to disambiguate aliasing device pointers.

* Copy headers from [STC](https://github.com/stclib/STC) for implementing the new pointer lookup and maybe other things in the future.

* Fix proactively clMemReleaseMemObject() did not removed all `pocl_raw_ptr` objects created at `clCreateBuffer()` with `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT`.

* Fix proactively `clMemFreeIntel()` did not released its shadow cl_mem objects.


lib/CL/clSetKernelExecInfo.c
```diff
@@ -111,9 +123,11 @@ POname(clSetKernelExecInfo)(cl_kernel kernel,
         for (size_t i = 0; i < param_value_size / sizeof (void *); ++i)
           {
             void *dev_ptr = ptrs[i];
-            pocl_raw_ptr *raw_ptr_info
-                = pocl_find_raw_ptr_with_dev_ptr (kernel->context, dev_ptr);
-            POCL_RETURN_ERROR_ON ((raw_ptr_info == NULL), CL_INVALID_VALUE,
+            // TODO: cl_ext_buffer_device_address 1.0.1 allows invalid
+            // pointers in clSetKernelArgDevicePointerEXT - does it
+            // apply on CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT too?
```
pjaaskel: This is not specified to my recollection. I think it's best to return invalid value here. Allowing it as kernel arguments is justified with C/C++ which allow passing garbage pointers around until they are dereferenced.

pjaaskel: Also please use the old /* */ style in C comments.

lib/CL/pocl_raw_ptr_set.h
```diff
@@ -0,0 +1,106 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  typedef struct pocl_raw_ptr_set pocl_raw_ptr_set;
+
+  /** Creates new empty pocl_raw_ptr_set.
+   */
+  POCL_EXPORT
```
pjaaskel: Can you not indent due to the extern "C" block? I know clang-format wants to indent here and I couldn't find a way to disable it.

lib/CL/pocl_raw_ptr.h
```diff
@@ -0,0 +1,47 @@
+
```
pjaaskel, line 2: License missing.

lib/CL/pocl_util.c
```diff
@@ -2375,26 +2375,9 @@ pocl_svm_check_get_pointer (cl_context context, const void *svm_ptr, size_t size
                             size_t *buffer_size, void** actual_ptr)
 {
 
-  /* TODO we need a better data structure than linked list,
```
pjaaskel, line 2378: Thanks for adding it ;-)

lib/CL/pocl_mem_management.c
```diff
@@ -1066,34 +1066,20 @@ pocl_raw_ptr *
 pocl_find_raw_ptr_with_vm_ptr (cl_context context, const void *host_ptr)
```
pjaaskel, line 1066: Do we still need these old functions for something if these just lock and delegate?

pjaaskel (changes requested):
Very nice! Small things only.

lib/CL/clSetKernelExecInfo.c
linehill: Addressed comment style.

lib/CL/pocl_raw_ptr_set.h
linehill: Fixed.

> I know clang-format wants to indent here and I couldn't find a way to disable it.

This seems to fix the indentation (included in this PR):
```diff
-IndentExternBlock: AfterExternBlock
+IndentExternBlock: NoIndent
```

lib/CL/pocl_raw_ptr.h
linehill: Addressed.

lib/CL/pocl_mem_management.c
linehill: There are plenty of `pocl_find_raw_ptr_{vm,dev}_ptr()` users - removing these and hoisting their code to the callers seems undesired due to code bloat it would add.

lib/CL/pocl_raw_ptr_set.c
```diff
@@ -0,0 +1,248 @@
...
+  free (ptr);
+}
+
+POCL_EXPORT void
+pocl_raw_ptr_set_erase_all (pocl_raw_ptr_set *set)
+{
+  assert (set && "Invalid pocl_raw_ptr_set argument!");
+
+  pocl_raw_ptr *ptr;
+  DL_FOREACH (set->head, ptr)
```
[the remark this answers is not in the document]
linehill: Addressed.

pjaaskel:
The SVM test failures are suspicious this PR touching that part a lot.

linehill:
> The SVM test failures are suspicious this PR touching that part a lot.

I have been resolving the issue. The latest changes should fix it.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2389,18 +2389,11 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq, int locked)
...
-              POCL_UNLOCK_OBJ (event->context);
-              // Leads to 'undefined symbol: POclReleaseContext'
-              // POname (clReleaseContext) (event->context);
-
-              dev->ops->svm_free (dev, ptr);
-            }
+	      void *ptr = cmd->svm_free.svm_pointers[i];
+	      /* This updates bookkeeping associated with the 'ptr'
+		 done by the PoCL core. */
+	      POname (clSVMFree) (event->context, ptr);
```
pjaaskel, line 2399: I have a vague feeling it's not good to call API functions from the driver layer due to some troubles with the ICD loader or such, but cannot remember exactly. Maybe @Kerilk remembers?

Kerilk: As long as you call them through the pocl private aliases, you should be fine. You should not call them through the public signatures, otherwise you may get reentrancy. I think the macro `POname` targets the non public version on platform where alias are supported.

Here are other examples of use:
https://github.com/pocl/pocl/blob/c72d4f3cd4b8892890f55b2d03a5e64e9e6ead2d/lib/CL/devices/level0/level0-driver.cc#L428

https://github.com/pocl/pocl/blob/c72d4f3cd4b8892890f55b2d03a5e64e9e6ead2d/lib/CL/pocl_img_buf_cpy.c#L250

So it wouldn't be a new occurrence, but I wonder what would be the potential issues on MacOS and windows, though I don't know how the linker behaves on these platforms.

franz: AFAIK calling via `POname(clFunc)` is equivalent to calling `POclFunc` which always resolves to libpocl symbol, IOW it never goes through the ICD loader (only calling `clFunc` without `PO` would). We're already doing this in a few places in the driver layer (mainly `clRelease*` calls).

Kerilk: For reference here is the code of the macros:
https://github.com/pocl/pocl/blob/c72d4f3cd4b8892890f55b2d03a5e64e9e6ead2d/lib/CL/pocl_cl.h#L260-L299

pjaaskel: OK. Thanks. Should be OK then.

pjaaskel:
Pls rebase to get the Julia CI fix in.

pjaaskel (approved):

PR #1952, opened 2025-06-11
minor CI fixes

franz:


lib/kernel/cuda/barrier.ll
```diff
@@ -3,7 +3,7 @@ declare void @llvm.nvvm.barrier0()
 ; this is merely to enforce non-opaque-pointers
 ; required because there are no other pointers in the code,
 ; and llvm-link just sets opaque-pointers setting to whatever default
-declare void @__pocl_unused(i8* %arg1) noduplicate
+declare void @__pocl_unused(ptr  %arg1) noduplicate
```
linehill, line 6: This probably is not needed anymore and could be deleted.

linehill (approved):
Only one small suggestion. Otherwise, LGTM!

PR #1963, opened 2025-06-27
sycl-bench + openvx-cts

franz:


lib/CL/devices/pthread/pthread.c
pjaaskel, line 1: This is interesting. I wonder how well the runtime estimation will work in practice and could it lead to perf impact on cases which would benefit from parallel execution from the start? I wonder could we implement multiple work queues instead and the usual work stealing approach: Start executing on the host thread only while starting up the worker threads with all WGs initially in the host thread's WQ and then let the workers steal work in chunks from it. If the kernel command is small enough it should execute all the WGs in the host thread before the workers get to it. Just an idea.

Also seems to lack clang-format, but I see it's a draft.

.github/workflows/build_linux.yml
```diff
@@ -48,10 +43,13 @@ jobs:
         # chipstar 1.1 only supports LLVM up to 17
         # chipstar 1.2 only supports LLVM up to 19
         # asan, tsan, ubsan = sanitizers
+        # openvx = OpenVX-CTS
         config: [cts_spirv, tier1, asan, tsan, ubsan]
         include:
           - llvm: 19
             config: chipstar
+          - llvm: 19
```
pjaaskel: I've tested OpenVX with LLVM 20.

lib/CL/devices/pthread/pthread.c
franz: Some details of your idea description are unclear to me, but mainly I think it focuses on the wrong part (thread switching overhead).

It's not obvious, but the main improvement is actually coming from a single line change, `POCL_BROADCAST_COND` to `POCL_SIGNAL_COND` in `pthread_scheduler_push_command`. This (broadcast) was done because of the way subdevices were implemented. I won't go into details, but in short: all worker threads are woken up for each new command. With certain sycl-bench benchmarks, which had a large number of very "small" commands this 1) caused a fairly large overhead when worker threads would repeatedly wake up & discover nothing to do & go to sleep, 2) prevented the CPU of worker thread which did the actual work from entering turbo state, because the other threads were consuming energy & creating heat. This caused significant  slowdown (hundreds of %) in those benchmarks (measured on a 16C/24T CPU).

So WRT your idea: "Start executing on the host thread only while starting up the worker threads" - the "starting up the worker threads" part is the main thing I'm trying to avoid / minimize (hence the time measuring). The idea makes sense in theory, but in practice it ignores the overhead & energy required to start the worker threads.

> I wonder how well the runtime estimation will work in practice

Good question, it does need more tuning. IIRC at last sycl-bench run, this PR was something like 300-500% better than `main` in benchmarks with many tiny commands, and about 5% slower overall (geomean). Perhaps i should move this commit into a separate PR, and we can discuss more when you're back.

> could it lead to perf impact on cases which would benefit from parallel execution from the start

Kernels which have zero measurements (have never been executed), are executed the same way as current main = with all worker threads. Then multiple measurements are taken & averaged to fix some jitter, but it's still fairly primitive and could be improved.

pjaaskel: > Perhaps i should move this commit into a separate PR, and we can discuss more when you're back.

Let's.

franz: Done

pjaaskel (approved):

PR #1968, opened 2025-07-01
Improve error message if recursion is detected

fweig:
Improves the error message if recursion is detected in the device code linker:
- Demangles OpenCL C++ symbols before printing (requires a hack to remove address space qualifiers from the mangled name...)
- Prints the specific function causing the recursion in addition to the original function where recursion was detected

The old error message looked like this:
```
Recursion detected in function: '_ZNU3AS42o23gpu7GPUdEdx11fillClusterEffihffRU3AS4KNS0_23GPUCalibObjectsTemplateINS0_8ConstPtrEEEfff'
```

New error message looks like this:
```
Recursion detected in function: 'o2::gpu::GPUdEdx::fillCluster(float, float, int, unsigned char, float, float, o2::gpu::GPUCalibObjectsTemplate<o2::gpu::ConstPtr> const&, float, float, float)'
-> Infriging function: 'o2::gpu::MultivariatePolynomialParametersHelper::factorial(unsigned int)'
```


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -110,8 +112,36 @@ static std::vector<Use *> findInstructionUses(GlobalVariable *GVar) {
 }
 
 
+// Remove address space qualifiers like U3AS4, U3AS1, etc. from mangled symbol
+static std::string stripAddressSpaces(const std::string &mangledName) {
+
+    // Pattern: U followed by digits, then AS, then a digit
+    std::regex asPattern(R"(U\d+AS\d)");
+    std::string result = std::regex_replace(mangledName, asPattern, "");
```
linehill: FWIW, I think this could unintentionally replace parts of the original function name. E.g. `_Z17my_function_U1AS1PU3AS1i` could get stripped and demangled to `my_function_` but the correct demangling should be `my_function_U1AS1` (or alternatively the demangling fails because the function name length, `_Z17`, is no longer correct).

fweig: Yes, that's a big issue. My hope is, that this name clash happens very rarely. So this hack can be used until the LLVM demangler eventually implements address space qualifiers. I don't see another workaround that doesn't involve implementing at least a partial demangler...

linehill: Sounds good to me.

lib/llvmopencl/LLVMUtils.cc
```diff
@@ -110,8 +112,36 @@ static std::vector<Use *> findInstructionUses(GlobalVariable *GVar) {
 }
 
 
+// Remove address space qualifiers like U3AS4, U3AS1, etc. from mangled symbol
+static std::string stripAddressSpaces(const std::string &mangledName) {
```
pjaaskel: Can you use the LLVM coding guidelines here? It means CamelCase variable names.

lib/llvmopencl/LLVMUtils.cc
```diff
@@ -125,21 +155,21 @@ regenerate_kernel_metadata(llvm::Module &M, FunctionMapping &kernels)
 {
   // reproduce the opencl.kernel_wg_size_info metadata
   NamedMDNode *wg_sizes = M.getNamedMetadata("opencl.kernel_wg_size_info");
-  if (wg_sizes != NULL && wg_sizes->getNumOperands() > 0) 
+  if (wg_sizes != NULL && wg_sizes->getNumOperands() > 0)
```
pjaaskel: This is a cleanup-only change, but it's OK. In that case can you add a bit of further cleaning and rename wg_size to WGSizes to avoid another cleanup commit to do that? Thanks!

lib/llvmopencl/LLVMUtils.cc
```diff
@@ -125,21 +155,21 @@ regenerate_kernel_metadata(llvm::Module &M, FunctionMapping &kernels)
...
+                 e = kernels.end(); i != e; ++i)
             {
               Function *old_kernel = (*i).first;
               Function *new_kernel = (*i).second;
               Function *func_from_md;
               func_from_md = dyn_cast<Function>(
                 dyn_cast<ValueAsMetadata>(wgsizeMD->getOperand(0))->getValue());
               if (old_kernel == new_kernel || wgsizeMD->getNumOperands() == 0 ||
-                  func_from_md != old_kernel) 
+                  func_from_md != old_kernel)
```
pjaaskel: Same here. FuncFromMD, OldKernel.  Or you can just remove the whitespace cleanup changes from this commit. We aim to improve the cleanliness/style of the code with each commit but try to avoid large cleanup-only commits that make the commit history harder to browse.

lib/llvmopencl/linker.cpp
```diff
@@ -691,9 +693,14 @@ int link(llvm::Module *Program, const llvm::Module *Lib, std::string &Log,
 
     SmallFunctionSet CallStack;
     llvm::SmallVector<llvm::Function *> Callees;
-    if (find_called_functions(&*FI, Callees, CallStack)) {
+    if (auto *R = find_called_functions(&*FI, Callees, CallStack); R != nullptr) {
```
pjaaskel: What's the "R != nullptr" part there? Haven't seen this before.

lib/llvmopencl/LLVMUtils.cc
fweig: Fixed.

lib/llvmopencl/LLVMUtils.cc
fweig: Fixed. Didn't really intend for that change. Just configured my editor to strip whitespaces at end of lines...

lib/llvmopencl/LLVMUtils.cc
fweig: Fixed.

lib/llvmopencl/linker.cpp
fweig: This is a C++17 feature for an init-statement before the branching condition. I used this to make the pointer check explicit here. LLVM style guide seems (?) to prefer the implicit nullptr check. So i removed the `R != nullptr`.

franz:
LGTM. The SYCL-CTS fail is unrelated to this PR (unreliable test).

pjaaskel (approved):
Thanks!

PR #1974, opened 2025-07-21
WILoops: fall to CBS when encountering unreachables

pjaaskel:
For WIL currently we can only delete BBs with 'unreachable' instructions in them to produce clean control flow graphs for the loop-based methods. However, deletion of the whole basic block (and redirecting the control to another one) is not correct when relying on 'unreachable' to equal to 'abort', like Julia does. Therefore, disable this behavior by default, making PoCL to fall back to CBS when encountering unreachable instructions. See, for example the Github Issue #1971 and a bunch of others.  

Unreachable/abort keeps causing headaches due to the LLVM dominator analysis which get confused with unreachable terminators. Something to optimize when there's an official abort (extension) implemented. 

Also allow CBS to be ran on kernels without barriers, so it can be used more widely as a fallback until the new fibers method is pushed.


tneval:
@pjaaskel could this be fallbacking CBS failing to execute some Julia test cases?

vchuravy:
The failing tests are the same as mentioned by @maleadt here https://github.com/pocl/pocl/issues/1871#issuecomment-2769083558

pjaaskel:
> @pjaaskel could this be fallbacking CBS failing to execute some Julia test cases?

@tneval, most likely yes.

> The failing tests are the same as mentioned by @maleadt here https://github.com/pocl/pocl/issues/1871#issuecomment-2769083558

The reproducer and the regression test we added from it do seem to work. So it must be some other root cause.

pjaaskel:
@maleadt what do you think: Is the fallback behavior for 'unreachables' better than the brutal one we have now?

I haven't yet come up with a solution for unreachables that would not kill vectorization performance combined with graceful return in cases with barriers. This will likely have to wait until we implement the CUDA-style partial barrier feature before it.

In loopvec-next we have a 'fibers' method which is more robust fallback than CBS, but one still doesn't get cross-WI vectorization out of it.

PR #1982, opened 2025-08-11
LLVM 21 support

franz:


lib/llvmopencl/linker.cpp
```diff
@@ -670,6 +679,47 @@ int link(llvm::Module *Program, const llvm::Module *Lib, std::string &Log,
 
   llvm::Module::iterator FI, FE;
 
+#if LLVM_MAJOR > 20
+  // For now, remove all Clang-generated kernel wrappers
```
pjaaskel: What are these? Something introduced in 21?

franz: Yes, this is completely new in 21.

pjaaskel: I wonder what's their purpose and could/should we use them in the future.

franz: It was introduced by LLVM commit [642481a4286c9006958274531](https://github.com/llvm/llvm-project/commit/642481a4286c9006958274531ee173b347866c50) to this PR: https://github.com/llvm/llvm-project/pull/115821

It's AMDGPU related Codegen fix to allow kernels calling other kernels. IIRC SPIR-V translator does something similar (creating stubs). Perhaps we could use it, right now we just inline all kernels that are called by other kernels (in lib/llvmopencl/InlineKernels.cc).

TBH i only noticed this change, because it broke our metadata parser that relies on OpenCL metadata attached to kernels (they're only partially copied to the stub function).

pjaaskel: OK. I see. Something to migrate to at some point if the feature sticks, but not urgent.

pocld/daemon.cc
```diff
@@ -332,6 +338,7 @@ static AvahiAdvertise *StartAvahiAdvert(addrinfo *RA,
     }
   }
 
+  POCL_MSG_ERR("@@@@@@@@@@@ IfIndex: %u  IpProto: %u \n", (unsigned)IfIndex, (unsigned)IpProto);
```
pjaaskel: Perhaps drop the @@@@@@ prefix?

pjaaskel (approved):

pocld/daemon.cc
franz: yeah these debug msgs were accidental commit

PR #1991, opened 2025-08-21
Expand matmul DBK on PoCL-L0-NPU for Llama.cpp

linehill:
Expand matmul DBK on PoCL-Level0-npu device by supporting `CL_TENSOR_LAYOUT_BLAS_EXP` datalayout (support is limited to row-major and alikes), supporting mixed input-output element type and adding new sub-buffer creation type for viewing regions of buffers as tensors.

Other changes:

* Fix tensor datalayouts were leaked (`cl_mem::tensor_layout`).

* Fix segfault when releasing an export allocation in level0 driver when a context, the allocation was bound to, was released.

* Enable cl_exp_tensor on host device without dependencies to external libraries such as libxssm, opencv and so on which are meant for DBKs.

* Fix off-by-one errors.

* Add `CL_EXP_{TENSOR,DEFINED_BUILTIN_KERNELS}_EXTENSION_VERSION`.


pjaaskel (approved):
Good work, LGTM. Perhaps something could be added to the release notes' experimental features section?

PR #1994, opened 2025-08-27
Fix device uninit and ensure vt_dev_close is called

menxb:
Problem
The Ventus device fails to properly perform uninit cleanup operations in the pocl environment, resulting in issues such as vt_dev_close not being called.
By analyzing the logic of the pocl_uninit_devices function on lines 384-385 of devices.c, it was found that the reinit function was missing, and the condition check failed and the cleanup operation was skipped.

Solution
Added a full implementation of the pocl_ventus_reinit function:

Key changes:
Added pocl_ventus_reinit function: Implement device reinitialization logic, including device data structure reconstruction, device handle reopening, lock initialization, etc
Update the device operation table: Set ops->reinit = pocl_ventus_reinit in pocl_ventus_init_device_ops
Add Function Declaration: Add the corresponding function declaration in pocl_ventus.h


pjaaskel:
Was this accidental? I advise to upstream the Ventus driver so it's easy to keep up with the upstream PoCL.

PR #2031, opened 2025-09-26
random fixes for 7.1

franz:


lib/CL/pocl_llvm_utils.cc
```diff
@@ -365,6 +365,22 @@ void cpu_setup_vector_widths(cl_device_id dev) {
     if (Features["sve"] || Features["sve2"] || Features["neon"])
       lane_width = 16;
 #endif
+#if defined(__linux__) && defined(__riscv)
```
pjaaskel, line 368: Can we add a mention about the now-verified RVV output to the release notes?

cmake/LLVM.cmake
```diff
@@ -166,8 +166,9 @@ endif()
 string(REPLACE " -pedantic" "" LLVM_CXXFLAGS "${LLVM_CXXFLAGS}")
 
 # Convert the LLVM's include path to -isystem so the headers are
+# disabled, this breaks with GCC 15
```
pjaaskel: Oh? How does it break? I thought this is a common way to mark include paths as non-project ones. Can we make it conditional. The LLVM warnings are rather annoying and tend to hide useful warnings.

lib/kernel/host/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ clamp_int.cl
 clz.cl
 ctz.cl
 convert_type.cl
+copysign.cl
```
pjaaskel, line 42: Do all of the native half CTS pass now with these? What is missing?

CREDITS
```diff
@@ -37,6 +37,7 @@ Krishnaraj Bhat
 Martin Hauke
 Volkan Keleş
 Lassi Koskinen
+Jakub Klinkovský
```
pjaaskel, line 40: I think this (long-unmaintained) CREDITS list was in the order of first contribution. I'm not sure should we try to keep maintaining it. It might miss other authors too.

pjaaskel (changes requested):
The LLVM warning commit would be nice to unrevert and use it at least for GCC < 15 if no universal solution is found. And perhaps mention RVV improvements in the releasenotes. Otherwise LGTM.

cmake/LLVM.cmake
franz: IIRC it breaks on `include_next <stdio.h>` or similar.

Edit: found a [report](https://stackoverflow.com/questions/37218953/isystem-on-a-system-include-directory-causes-errors) of almost exact same error from ~9 years ago (also with LLVM).

TLDR: the problem is that `-isystem` also changes search order of include directories. I'll adjust the commit to only disable this isystem hack when the LLVM includedir is `/usr/include`.

linehill: Would `-idirafter` help here? IIUC, it's same as `-isystem` but the it puts the include search directory after standard system directories so it should avoid the issue described in the SO link.

franz: @linehill  thanks, `-idirafter` indeed fixes the problem.

lib/kernel/host/CMakeLists.txt
franz: No unfortunately, there are a lot more functions missing. The short answer is that each function using SLEEF or libclc implementation is missing the FP16 version. Those using libclc, might already have a fp16 version in the upstream libclc, i'm not sure if upstream SLEEF has fp16. The functions we currently support with FP16 are 1) the ones that have a corresponding Clang `__builtin_xyzf16()` and 2) the ones that are simple enough to be defined as expression or bit of OpenCL C (like `max`, `min`, `select` and such).

pjaaskel: OK, maybe updating libclc would be useful.

pjaaskel: Since native half is "a common need" nowadays.

CREDITS
pjaaskel: Pls just move the name to the end of the list if you will add more commits to this PR anyhow.

PR #2042, opened 2025-10-15
fix(macos): Add SDK sysroot to kernel linker flags

### LINKED ISSUE #2041: macOS: JIT kernel linking fails with "ld: library 'System' not found"
macOS (15.0+), pocl's JIT compiler fails to link kernel shared objects with the error:

```
ld: library 'System' not found
error: linker command failed with exit code 1
```

pvelesko:
On modern macOS, libSystem is provided by the SDK as libSystem.tbd rather than being in /usr/lib. Without specifying the sysroot, clang's linker driver cannot find libSystem when linking kernel shared objects, resulting in "ld: library 'System' not found" errors.

This commit adds the `-isysroot` flag to DEFAULT_HOST_LD_FLAGS for macOS, pointing to the SDK path obtained via `xcrun --show-sdk-path`.

Fixes JIT compilation errors on macOS 15 (Sequoia) and later.

Error before fix:
  ld: library 'System' not found
  error: linker command failed with exit code 1

Tested on:
  - macOS 15.0.0 (Sequoia)
  - Apple M4 Pro (arm64)
  - Xcode 16

Fixes #2041

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1150,7 +1150,9 @@ if(UNIX)
     # MacOS ld outputs useless warnings like
     # ld: warning: -macosx_version_min not specificed, assuming 10.7
     # suppress them with -w.
-    set(DEFAULT_HOST_LD_FLAGS "-dynamiclib -w -nostartfiles")
+    # Use execute_process to get SDK path and add -isysroot so clang can find libSystem.tbd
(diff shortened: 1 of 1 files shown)


tneval:
@pvelesko can this cause problems if one wants to use custom-built/homebrew LLVM instead of the system LLVM?

tneval:
> Ok yeah I think the fix should be in LLVM but I would suggest at least providing a warning. It took me a while to figure out what is going

Yeah, I feel you. This needs to be addressed in the docs, and maybe think of some kind compile-time check to catch this situation.

PR #2052, opened 2025-10-30
Add POCL_DEBUG_LLVM_OPTS environment variable

linehill:
Add new debug environment variable `POCL_DEBUG_LLVM_OPTS` for setting LLVM defined `cl::opt` options.
  
 The variable supports:
* multiple options separated by `,`s,
* options with values in a form of `some-option=with-some-value` and 
* options with or without leading `-`s.
    
A primer for this is to control the amount of LLVM pass debug prints. E.g. print only when IR changes and print IR only before and after some passes.

Also, 
*  register standard pass instrumentations on PoCL's pass pipelines so LLVM's print IR options work on them.
*  change PoCL's pass pipelines to verify IR after every pass in Debug builds or when `POCL_LLVM_VERIFY=1`.

### DIFF
--- a/doc/sphinx/source/using.rst
+++ b/doc/sphinx/source/using.rst
@@ -170,6 +170,19 @@ pocl.
  'refcounts', 'timing', 'hsa', 'tce', 'cuda', 'vulkan', 'proxy' and 'all'.
  Note: setting POCL_DEBUG to 1 still works and equals error+warning+general.
 
+- **POCL_DEBUG_LLVM_OPTS**
+
+  Set LLVM options for debugging purposes. Multiple options are
+  supported and are separated by ``,``. Options taking values are
+  defined as ``-some-option=some-value``. Leading dashes on the option
+  names are optional.
+
+  Most of the options listed in ``opt-XX --help`` and ``opt-XX
+  --help-hidden`` are assumed to work but some of them may be
+  unavailable due to being tied to specific tools and not to the LLVM
+  library or depending on the configuration of LLVM (e.g. ``-debug``
+  may not work in release builds of LLVM).
+
 - **POCL_DEBUG_LLVM_PASSES**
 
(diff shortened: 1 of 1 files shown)


lib/CL/pocl_llvm_utils.cc
```diff
@@ -563,6 +598,10 @@ void InitializeLLVM() {
       O->addOccurrence(1, StringRef("debug-only"), StringRef("inline"), false);
 #endif
     }
+    if (auto *llvm_opts_string =
```
pjaaskel: Can you convert the variables you touch in your diff to CamelCase C++ style?

lib/CL/pocl_llvm_wg.cc
```diff
@@ -106,6 +106,16 @@ POP_COMPILER_DIAGS
 
 using namespace llvm;
 
+static bool verifyIR() {
+  return pocl_get_bool_option("POCL_LLVM_VERIFY", LLVM_VERIFY_MODULE_DEFAULT);
```
pjaaskel, line 110: Seems this old env misses docs.

pjaaskel (changes requested):
Very useful feature. Small feedback only.

lib/CL/pocl_llvm_utils.cc
linehill: Done.

lib/CL/pocl_llvm_wg.cc
linehill: There is documentation for [this](https://github.com/pocl/pocl/blob/1f16b540cde02ec2ec3c73ee1a315a39b1d1a9fc/doc/sphinx/source/using.rst?plain=1#L291-L296).

pjaaskel (approved):

franz:
I've merged #2058 which should resolve Julia CI failures.

PR #2053, opened 2025-10-31
cuda: more subgroup collaboration functions & arithmetic mappings

jansol:
~~sub_group_reduce_min is still doing something weird and the sync variants are untested (hence draft)~~

EDIT: all reduces and scans now pass the CTS tests on PTX6.0+ targets.

### DIFF
--- a/doc/sphinx/source/cuda.rst
+++ b/doc/sphinx/source/cuda.rst
@@ -117,6 +117,10 @@ The following is a non-comprehensive list of known issues in the CUDA backend:
 
 * image types and samplers are unimplemented
 * printf format support is incomplete
+* subgroup reduce is not supported with CUDA versions older than 9.0
+* subgroup ballot is only supported on SM_30 - SM_64 targets with CUDA 9.0+
+* LLVM 21 is not supported due to an upstream bug. Use LLVM 17-20 instead. For
+  details, see https://github.com/llvm/llvm-project/issues/154772
 
 Additionally, there has been little effort to optimize the performance of this
 backend so far - the current effort is on implementing remaining functionality.
--- a/doc/sphinx/source/notes_7_2.rst
+++ b/doc/sphinx/source/notes_7_2.rst
@@ -1,0 +1,17 @@
+**************************
+Release Notes for PoCL 7.2
+**************************
+
+===========================
+Driver-specific features
+===========================
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+CUDA driver
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Implement sub_group_{reduce,scan_exclusive,scan_inclusive}_* and
+  sub_group_{all,any,broadcast}. (reduce is only available for PTX6.0+)
+* Note that CUDA driver does not support LLVM 21, due to a bug
+  in upstream Clang code. Users must use LLVM 17 to 20 with CUDA. For details,
+  see https://github.com/llvm/llvm-project/issues/154772
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2202,7 +2202,7 @@ if(ENABLE_CUDA)
   endif()
 
   if(LLVM_VERSION_MAJOR EQUAL 21)
-    message(FATAL_ERROR "CUDA driver does not work with LLVM/Clang 21")
+    message(FATAL_ERROR "CUDA driver does not work with LLVM/Clang 21: https://github.com/llvm/llvm-project/issues/154772")
   endif()
 
   set(OCL_DRIVERS "${OCL_DRIVERS} cuda")


lib/kernel/cuda/subgroup_sync.cl
```diff
@@ -0,0 +1,149 @@
...
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+   IN THE SOFTWARE.
+*/
+
+int _CL_OVERLOADABLE
+sub_group_all (int predicate)
+{
+  uint active = __nvvm_activemask ();
+  return __nvvm_vote_all_sync (active, !!predicate);
```
linehill: Does the `__nvvm_activemask()` behave the same way as CUDA's `__activemask()`? If so, the `sub_group_all` may behave unexpectedly on modern architectures. The `__activemask()` does not necessarily return a mask for threads in warp but a possibly a subset of it depending on the scheduling.

Since all WIs in the subgroup must encounter the `sub_group_all()` (same for other collective sub-group functions) this probably can be just implemented as:
```suggestion
  return __nvvm_vote_all_sync (0xFFFFFFFF, !!predicate);
```

pjaaskel:
Is this still a draft on purpose? Can you rebase?

jansol:
Redid the reduce operations (except for <PTX6.0), reorganized a bit and went ahead and implemented the scan operations for good measure. Now it passes the `subgroup_functions_core` portion of the CTS (tested on Pascal and Ampere).

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -1053,3 +1056,26 @@ int pocl_cuda_get_ptr_arg_alignment(void *LLVM_IR, const char *KernelName,
...
+  std::unique_ptr<llvm::Module> Magic{
+      new llvm::Module("<pocl-cuda-subgroup-size>", Module->getContext())};
+  llvm::IRBuilder<> Builder(Magic->getContext());
+  llvm::GlobalVariable *Variable = new llvm::GlobalVariable(
+      Builder.getInt32Ty(), true,
+      llvm::GlobalVariable::LinkageTypes::ExternalLinkage,
+      Builder.getInt32(SGSize), "_pocl_sub_group_size");
+  Magic->insertGlobalVariable(Variable);
+  llvm::Linker Link(*Module);
+  Link.linkInModule(std::move(Magic));
```
linehill: This could be simply done by setting an initializer. Something like:
```c++
SGSizeVar->setInitializer(ConstantInt::get(SGSizeVar->getValueType(), SGSize));
```

lib/CL/devices/cuda/pocl-ptx-gen.cc
```diff
@@ -1053,3 +1056,26 @@ int pocl_cuda_get_ptr_arg_alignment(void *LLVM_IR, const char *KernelName,
   std::memcpy(Alignments, AVec.data(), sizeof(size_t) * AVec.size());
   return 0;
 }
+
+int pocl_cuda_override_sub_group_size(void *llvm_module, int SGSize) {
```
linehill: A bit misnomer here as the _pocl_sub_group_size is not defined IUCC. So maybe:
```suggestion
int pocl_cuda_define_sub_group_size(void *llvm_module, int SGSize) {
```

linehill:
Should [this line](https://github.com/pocl/pocl/blob/ab1c837188c4ad38ce21d64648c199450424bf85/lib/CL/devices/cuda/pocl-cuda.c#L811) changed as follows:
```diff
-  dev->max_num_sub_groups = dev->max_work_group_size / 32;
+  dev->max_num_sub_groups = dev->max_work_group_size / data->warp_size;
```
?

doc/sphinx/source/notes_7_2.rst
```diff
@@ -0,0 +1,16 @@
...
+===========================
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+CUDA driver
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Implement sub_group_{reduce,scan_exclusive,scan_inclusive}_* and
+  sub_group_{all,any,broadcast}. (reduce is only available for PTX6.0+)
+* Note that CUDA driver does not support LLVM 21, due to a bug
+  in upstream Clang code. Users must use LLVM 17 to 20 with CUDA.
```
linehill: Would be nice if there were a configure-time check for the unsupported combinations. 

Could you mention the bug in the `cuda.rst`'s known issues section with a link to the upstream issue?

franz: The problem of CUDA + LLVM 21 is actually already mentioned in 7.1 notes. There is also a CMake check (with FATAL_ERROR) for CUDA + LLVM 21.

jansol:
@linehill I fell into a bit of a rabbit hole looking into whether reduce could be supported for < PTX 6.0 after all. And the comment in `subgroup_ballot.cl` seems a bit odd as [the clang builtins listing](https://github.com/llvm/llvm-project/blob/llvmorg-17.0.0/clang/include/clang/Basic/BuiltinsNVPTX.def#L622) doesn't actually mention a minimum PTX for `__nvvm_vote_ballot`. Did you perhaps mean `__nvvm_vote_ballot_sync` with that comment, as that *does* have a PTX6.0+ requirement (as do the `.sync` versions of pretty much everything else)? Looking at https://docs.nvidia.com/cuda/archive/8.0/parallel-thread-execution/index.html#release-notes `vote.ballot` dates back all the way to SM2.0 so it should be accessible no problem.

At the same time `subgroup.ll` refers to `@llvm.nvvm.bar.warp.sync(i32)`, [which does have a PTX6.0 requirement listed in clang](https://github.com/llvm/llvm-project/blob/llvmorg-17.0.0/clang/include/clang/Basic/BuiltinsNVPTX.def#L589) as well as [NVIDIA's PTX ISA documentation](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-bar-warp-sync) so I'm a bit suspicious about the correctness of the target level split in the first place.

If my understanding is correct I'd reorganize things so that
* sm30-ptx60lt => `subgroup_ballot.cl` + `subgroup_legacy.cl` + (maybe everything folded into `subgroup_legacy.cl` along with a `sub_group_barrier` impl that uses `__nvvm_bar` instead of `@llvm.nvvm.bar.warp.sync`, getting rid of `subgroup.ll` entirely)
* sm30-ptx60-63 => `subgroup_sync.cl` with a CL C version of `sub_group_barrier`
* sm70-ptx64 => also `subgroup_sync.cl` (is it really needed then?), or if we are really ambitious and there are no big reasons *not* to do so, this could be bumped to SM 8.0 and PTX 8.8 which adds dedicated reduce and scan intrinsics (or PTX8.0 if we don't mind keeping the current implementation for floats)

linehill:
Welcome to the rabbit hole - it goes deeper :). It’s been a while I have worked on this area so I’ll try to remember why things are arranged here as so.

> @linehill I fell into a bit of a rabbit hole looking into whether reduce could be supported for < PTX 6.0 after all. And the comment in `subgroup_ballot.cl` seems a bit odd as [the clang builtins listing](https://github.com/llvm/llvm-project/blob/llvmorg-17.0.0/clang/include/clang/Basic/BuiltinsNVPTX.def#L622) doesn't actually mention a minimum PTX for `__nvvm_vote_ballot`. Did you perhaps mean `__nvvm_vote_ballot_sync` with that comment, as that _does_ have a PTX6.0+ requirement (as do the `.sync` versions of pretty much everything else)? 
> Looking at https://docs.nvidia.com/cuda/archive/8.0/parallel-thread-execution/index.html#release-notes `vote.ballot` dates back all the way to SM2.0 and PTX 1.2 so it definitely should be accessible no problem.

There is a [quirk in the LLVM NVPTX backend](https://github.com/llvm/llvm-project/blob/6459f39c377dc8b7d5d81ef365553c8625fb4def/llvm/lib/Target/NVPTX/NVPTXIntrinsics.td#L231-L258) that the PTX emission for the  __nvvm_vote_ballot{,_sync} intrinsic is only enabled when targeting >= PTX 6.0 and  >= SM 3.0. Below those versions, the compilation probably result in `fatal error: error in backend: Cannot select: intrinsic %llvm.nvvm.vote.ballot` error or similar. 

> 
> At the same time `subgroup.ll` refers to `@llvm.nvvm.bar.warp.sync(i32)`, [which does have a PTX6.0 requirement listed in clang](https://github.com/llvm/llvm-project/blob/llvmorg-17.0.0/clang/include/clang/Basic/BuiltinsNVPTX.def#L589) as well as [NVIDIA's PTX ISA documentation](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-bar-warp-sync) so I'm a bit suspicious about the correctness of the target level split in the first place.

I'm not sure if I have noticed the use of `@llvm.nvvm.bar.warp.sync(i32)` or it haven't mattered until now.

> If my understanding is correct I'd reorganize things so that
> 
>     * sm30-ptx60lt => `subgroup_ballot.cl` + `subgroup_legacy.cl` + (maybe everything folded into `subgroup_legacy.cl` along with a `sub_group_barrier` impl that uses `__nvvm_bar` instead of `@llvm.nvvm.bar.warp.sync`, getting rid of `subgroup.ll` entirely)

Probably won't work out because of the LLVM backend quirk I mentioned.

>     * sm30-ptx60-63 => `subgroup_sync.cl` with a CL C version of `sub_group_barrier`

The use of _sync variant will probably cause regressions - and we might not have CIs to catch them for this specific SM-PTX version combination. OpenCL's subgroup shuffles and ballots calls don't require that all WIs of the sub-group encounter them. For CUDA device, we can implement this semantic using the non-`_sync` variants. If we were to use the `_sync` variants instead, we would need to somehow know the exact thread mask instead of `FULL_MASK` in scenarios like `if (divergent_condition) { sub_group_ballot(...); ... }`.

>     * sm70-ptx64 => also `subgroup_sync.cl` (is it really needed then?), or if we are really ambitious and there are no big reasons _not_ to do so, this could be bumped to SM 8.0 and PTX 8.8 which adds dedicated reduce intrinsics (or PTX8.0 if we don't mind keeping the current implementation for floats)

Maybe it's OK to implement shuffles and ballots with `subgroup_sync.cl` when PoCL is configured with `ENABLE_CONFORMANCE=OFF` for use case the shuffle and ballot functions are called by all WIs in the subgroups. Otherwise, I don't think we want to confuse users with sub-group shuffle and ballot support which works some times / in certain setups.

linehill:
> Subgroups for the CUDA driver are gated behind `-DENABLE_CUDA_EXPERIMENTAL_EXTENSIONS=ON`

Got it. Yeah, the subgroup tests passes.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2937,3 +2938,67 @@ pocl_cuda_get_device_info_ext (cl_device_id device, cl_device_info param_name,
...
+        for (size_t i = 0; i < 3; ++i)
+          {
+            if (input_value_size / sizeof (size_t) > i)
+              ndcount[i] = ((size_t *)input_value)[i];
+          }
+        size_t b = ndcount[0] * ndcount[1] * ndcount[2];
+        if (b % sg_size == 0)
+          b = b / sg_size;
+        else
+          b = b / sg_size + 1;
```
linehill: Could be simplified to:
```suggestion
        b = (b + sg_size - 1) / sg_size;
```

doc/sphinx/source/cuda.rst
```diff
@@ -117,6 +117,10 @@ The following is a non-comprehensive list of known issues in the CUDA backend:
 
 * image types and samplers are unimplemented
 * printf format support is incomplete
+* subgroup reduce is not supported on pre-PTX6.0 targets
+* subgroup ballot is only supported on PTX 6.0 - 6.3
```
linehill: IMO, this doesn't communicate well if users' GPU supports these features - the supported GPUs should probably expressed in terms of compute capability (or `sm_##`). 

Be aware that PoCL chooses a PTX version to emit code for and prefers to use a version that supports the OpenCL C features the most (with possible performance trade-offs). For detail, see: [choose_sm_ptx_version_combo()](https://github.com/pocl/pocl/blob/806bef35cd80036c3f1d1e532a5e60bb9d010825/lib/CL/devices/cuda/pocl-cuda.c#L324).

linehill:
LGTM, thanks.

PR #2059, opened 2025-11-07
rearrange GH workflows + fix device-side printf for 32bit devices (ALMAIF)

franz:


lib/CL/devices/printf_buffer.c
```diff
@@ -487,30 +487,30 @@ __pocl_printf_format_full (param_t *p, char *buffer, uint32_t buffer_size)
...
+                        if (vector_length <= pointer_size_bytes)
                           {
-                            alloca_length = 8;
+                            alloca_length = pointer_size_bytes;
                           }
                         break;
                       case 2:
                         __pocl_print_ints_ushort (p, buffer, vector_length,
                                                   is_unsigned);
-                        if (vector_length <= 4)
```
pjaaskel, line 503: Do we want to use 64b for storing these < 32b integers in any target? Could we simplify and use 32b for all <= 32b types and 64b for the rest?

franz: > Could we simplify and use 32b for all <= 32b types and 64b for the rest?

I think this might be a problem with 64bit RISCV. IIRC some 32bit type (`half2` vector ? can't remember exactly) was passed as single 64bit integer in the ABI. So i would have to detect this in the printf arguments, and reduce the int64 to int32. So it would probably not make the code simpler (at least the EmitPrintf code).

PR #2075, opened 2025-12-02
Command buffers: Allow mutable updates to pending cmd bufs

TopiLeppanen:
This makes the clUpdateMutableCommandsKHR readiness check match the check in clEnqueueCommandBufferKHR, so that mutable updates are allowed even if the cmd buf is still pending (and the cmd buf is simultaneous).

The check is refactored into an utility function, as it is now the same in these two files.

Relevant Spec portions:
For clEnqueueCommandBufferKHR:
"CL_INVALID_OPERATION if command_buffer was not created with the CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR flag and is in the Pending state."

And for the clUpdateMutableCommandsKHR:
"Using this function when command_buffer is in the pending state and not created with the CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR flag causes undefined behavior."

So I think both imply that the operation should succeed when cmd buf is in Pending state and was created with the SIMULTANEOUS-flag.

### DIFF
--- a/lib/CL/pocl_cmdbuf.c
+++ b/lib/CL/pocl_cmdbuf.c
@@ -61,9 +61,9 @@ pocl_cmdbuf_create_command (_cl_command_node **cmd,
   (*cmd)->buffered = 1;
 
   /* pocl_cmdbuf_choose_recording_queue should have been called to ensure we
-   * have a valid command queue, usually via CMDBUF_VALIDATE_COMMON_HANDLES
-   * but at that time *cmd was not allocated at that time, so find the queue
-   * index again here */
+   * have a valid command queue, usually via
+   * pocl_cmdbuf_validate_common_handles but at that time *cmd was not
+   * allocated at that time, so find the queue index again here */
   for (unsigned i = 0; i < command_buffer->num_queues; ++i)
     {
       if (command_buffer->queues[i] == command_queue)
@@ -201,6 +201,44 @@ pocl_cmdbuf_get_property (cl_command_buffer_khr command_buffer,
         return command_buffer->properties[2 * i + 1];
     }
   return 0;
+}
+
+cl_int
+pocl_cmdbuf_validate_common_handles (cl_command_buffer_khr command_buffer,
+                                     cl_command_queue *command_queue,
+                                     cl_mutable_command_khr *mutable_handle)
+{
+  assert (command_queue != NULL);
+  cl_command_queue q = *command_queue;
+
+  POCL_RETURN_ERROR_COND ((!IS_CL_OBJECT_VALID (command_buffer)),
+                          CL_INVALID_COMMAND_BUFFER_KHR);
+  cl_device_id dev = command_buffer->queues[0]->device;
+  if (strstr (dev->extensions, "cl_khr_command_buffer_multi_device"))
+    {
+      POCL_RETURN_ERROR_COND ((q == NULL && command_buffer->num_queues > 1),
+                              CL_INVALID_COMMAND_QUEUE);
+      int queue_in_buffer = 0;
+      for (unsigned ii = 0; ii < command_buffer->num_queues; ++ii)
+        {
+          queue_in_buffer |= (q == command_buffer->queues[ii]);
+        }
+      POCL_RETURN_ERROR_COND ((q != NULL && !queue_in_buffer),
+                              CL_INVALID_COMMAND_QUEUE);
+    }
+  else
+    {
+      POCL_RETURN_ERROR_ON (
+        (q != NULL), CL_INVALID_COMMAND_QUEUE,
+        "device does not support cl_khr_command_buffer_multi_device");
+    }
+  if (dev->cmdbuf_mutable_dispatch_capabilities == 0)
+    {
+      POCL_RETURN_ERROR_COND ((mutable_handle != NULL), CL_INVALID_VALUE);
+    }
+  cl_int errcode
+    = pocl_cmdbuf_choose_recording_queue (command_buffer, command_queue);
+  return errcode;
 }
 
 /**
--- a/lib/CL/pocl_cmdbuf.h
+++ b/lib/CL/pocl_cmdbuf.h
@@ -66,48 +66,14 @@ pocl_cmdbuf_get_property (cl_command_buffer_khr command_buffer,
 /* Returns 1 if command buffer is ready to be enqueued or mutated */
 int pocl_cmdbuf_is_ready (cl_command_buffer_khr command_buffer);
 
+cl_int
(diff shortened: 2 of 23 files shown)


lib/CL/pocl_util.h
```diff
@@ -159,6 +159,8 @@ cl_command_buffer_properties_khr
 pocl_cmdbuf_get_property (cl_command_buffer_khr command_buffer,
                           cl_command_buffer_properties_khr name);
 
+int pocl_is_cmdbuf_ready (cl_command_buffer_khr command_buffer);
```
pjaaskel: Seems the other cmdbuf helpers are named pocl_cmdbuf. While at it, can you extract those to pocl_cmdbuf.h to reduce pocl_util.h bloating? Also please add a comment to the new function.

TopiLeppanen:
- Renamed pocl_is_cmdbuf_ready to pocl_cmdbuf_is_ready
- Extracted cmdbuf-specific utils from pocl_util.* to pocl_cmdbuf.*
- Added a comment

lib/CL/pocl_cmdbuf.h
```diff
@@ -0,0 +1,110 @@
...
+extern "C"
+{
+#endif
+
+cl_int pocl_command_record (cl_command_buffer_khr command_buffer,
+                            _cl_command_node *cmd,
+                            cl_sync_point_khr *sync_point);
+
+cl_int
+pocl_create_recorded_command (_cl_command_node **cmd,
```
pjaaskel: This and the above function should be renamed to the same naming convention as the other cmdbuf functions? pocl_cmdbuf_create_recorded_command() and pocl_cmdbuf_record_command()

jansol: Or just `pocl_cmdbuf_create_command`, agreed. The original name is just to mirror the function in `pocl_util.c` that this was originally derived from and located next to.

lib/CL/pocl_cmdbuf.h
```diff
@@ -0,0 +1,110 @@
...
+                          cl_command_buffer_properties_khr name);
+
+/* Returns 1 if command buffer is ready to be enqueued or mutated */
+int pocl_cmdbuf_is_ready (cl_command_buffer_khr command_buffer);
+
+#ifdef __cplusplus
+}
+#endif
+
+#define CMDBUF_VALIDATE_COMMON_HANDLES                                        \
```
jansol: I think this is long enough that it warrants being an actual function by now. Doesn't have to be done as part of this PR though.

lib/CL/pocl_cmdbuf.c
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Check if command buffer is ready to be enqueued or mutated
+ *
+ * The command buffer must either be in executable state,
+ * or it can be in a pending state, if the cmd buf was created with the
+ * simultaneous-flag
+ *
+ */
+int
```
jansol, line 252: Ditto about the doc comment.

lib/CL/pocl_cmdbuf.c
```diff
@@ -0,0 +1,231 @@
...
+       * references to their buffers and they are expected to live until the
+       * entire command buffer is destroyed. */
+      command_buffer->migr_infos = pocl_append_unique_migration_info (
+        command_buffer->migr_infos, mi->buffer, mi->read_only);
+    }
+  LL_APPEND (command_buffer->cmds, cmd);
+
+  if (sync_point != NULL)
+    *sync_point = command_buffer->num_syncpoints + 1;
+  command_buffer->num_syncpoints++;
```
jansol, line 127: Maybe move this above the if clause and get rid of the duplicate increment.

lib/CL/pocl_cmdbuf.c
```diff
@@ -0,0 +1,231 @@
...
+#include "utlist.h"
+
+#include "pocl_cmdbuf.h"
+
+/**
+ * Create a command buffered command node.
+ *
+ * The node contains the minimum information to "clone" launchable
+ * commands in clEnqueueCommandBufferKHR.c.
+ */
```
jansol: Hmm should internal API doc comments be in the header instead of the implementation? Now that these have been broken out of util.c I'd kind of look at the header first for quick documentation.

Also it *might* make sense to mention that this is analogous to pocl_create_command in util.c, since it was initially derived from that.

lib/CL/pocl_cmdbuf.h
TopiLeppanen: renamed pocl_command_record -> pocl_cmdbuf_record_command
and pocl_create_recorded_command -> pocl_cmdbuf_create_command

lib/CL/pocl_cmdbuf.h
TopiLeppanen: Changed the pragma to a function. It does have a slighly weird looking code where it mutates the incoming function argument. I think it is to implement the possible command_buffer == NULL. I didn't change this behaviour, but I think at least it is now a bit easier to understand than inside the previous pragma.

lib/CL/pocl_cmdbuf.c
TopiLeppanen: Did this

lib/CL/pocl_cmdbuf.c
TopiLeppanen: Moved this comment to a header, and mentioned the link to pocl_create_command

lib/CL/pocl_cmdbuf.h
jansol: Thanks, I do also think it's much nicer to read like this. As for the mutated argument, I think you mean the command_*queue*? That one is allowed to be NULL on single-queue command buffers (since it's obvious). From the spec of clCommandCopyBufferKHR:

```
command_queue specifies the command-queue the command will be recorded to.
If the cl_khr_command_buffer_multi_device extension is not supported, only a single command-queue is supported, and command_queue must be NULL.
If the cl_khr_command_buffer_multi_device extension is supported and command_queue is NULL, then only one command-queue must have been set on command_buffer creation; otherwise, command_queue must not be NULL.
```

It's only meant to mutate the local copy of the argument internally in the function, not the caller's copy. The reason it does this is so that the rest of the function is guaranteed to have a valid pointer to the command queue. Maybe it would be a bit clearer if the *actual* command queue was extracted to a separate local variable in each of the functions, but functionally it should not make a difference. Alternatively it would probably be a good idea to document this mutation in a doc comment of the validation function.

pjaaskel (approved):

PR #2080, opened 2025-12-10
Fill image info before ops->alloc_mem_obj

jansol:
Optimized texture allocations need info about the format and dimensions so it's better to have this available in the mem object already at device memory allocation time. This doesn't meaningfully change much right now, but it should noticeably simplify adding image support for targets that have specialized handling for sampling 2D images in a cache-friendly way (that is, image storage in non-linear order), such as Vulkan, CUDA and (I think) LevelZero.

One thing that this *does* meaningfully change is image allocations on the proxy driver, which had assumed that image info is filled out prior to the per-device function being called and ended up feeding uninitialized (well, zeroed by `calloc`) values to the underlying driver.

### DIFF
--- a/lib/CL/clCreateBuffer.c
+++ b/lib/CL/clCreateBuffer.c
@@ -35,6 +35,7 @@ pocl_create_memobject (cl_context context,
                        cl_mem_flags flags,
                        size_t size,
                        cl_mem_object_type type,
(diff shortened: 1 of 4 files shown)


linehill (approved):
LGTM. 

The CI failures seems to be unrelated to this PR.

Issue #2024, opened 2025-09-17
[CUDA] Forced to run with POCL_CUDA_DISABLE_QUEUE_THREADS=0


arrebarritra:
The CUDA backend defaults to POCL_CUDA_DISABLE_QUEUE_THREADS=1. With this setting I invariably run into various failed asserts within queue handling, such as
```bash
pocl_update_event_running_unlocked: Assertion `event->status == CL_SUBMITTED' failed
```
```bash
 |      CUDA | Error during cuEventSynchronize
CUDA_ERROR_INVALID_HANDLE: invalid resource handle
```
It seems the driver is trying to handle an event that has already been handled. Could it be perhaps there is some place in the code that doesn't check `POCL_CUDA_DISABLE_QUEUE_THREADS` and are running with threads somewhere?

This mostly happens in an area of code where there are a lot of buffer writes and copies. I can try to produce an MRE.

jansol:
I remember seeing some event synchronization issues back when I implemented peer-to-peer migrations for the CUDA driver, but not having time to really solve them.

franz:
I haven't looked at the CUDA driver for a long while, but yeah there were race conditions and i'm 99% sure they're still there.

I'm guessing DISABLE_QUEUE_THREADS=1 is the default because of having lowest submission latency, and it works well if the user program uses a single queue, but if it uses more queues simultaneously, it likely hits those race conditions.

pjaaskel:
Sounds reasonable. Perhaps try it out and send a PR if it works?

PR #2088, opened 2025-12-16
[CUDA] Thread safe queue join and wait events

arrebarritra:
Closing #2024

This PR makes events and queues thread safe on the CUDA driver with `POCL_CUDA_DISABLE_QUEUE_THREADS=1` (default setting). Specifically, event/queue synchronisation commands `clFinish` (which is also used by blocking commands) and `clWaitForEvents` can be called from multiple user threads. Note that this is not the case for threaded event handling (ie. `POCL_CUDA_DISABLE_QUEUE_THREADS=0`), that is still not thread safe but could be probably be fixed in a similar way.

### Problem
Previously a race conditions occurred when the user used a queue on multiple threads and performed any synchronisation. When `clFinish` was called from different threads, there was a race condition if one thread was in the process of submitting, while another completed the event. One of two errors would occur:
1. The completed thread has freed the CUDA event, so the submititng thread gets `CUDA_ERROR_INVALID_HANDLE` in `pocl_cuda_finalize_command`
2. The submitting thread fails the `event->status == SUBMITTED` assert in `pocl_update_event_running`
Similarly, this would occur when calling `clWaitForEvents`.

### Solution
The solution is to make sure only one thread calls `pocl_cuda_finalize_command` for any given thread. This is achieved by holding a lock on the queue data of the CUDA driver for the duration of `pocl_cuda_wait_event`. Any subsequent threads trying to wait for this event find it has been completed.

### Changes
- Queue data lock is also created for non threaded event handling
- A lock on queue data is obtained in `pocl_cuda_join` and calls `pocl_wait_event_recurse`
- `pocl_wait_event` also obtains a lock on the queue data, this is only called from `clWaitForEvents`
- Some whitespace fixes

### Testing
I have evaluated this on WSL Ubuntu on a GTX 1070 and RTX 4070, as well as on Ubuntu on a Jetson Orin Nano. I have tested with a small script [opencl-multithread-cpp.txt](https://github.com/user-attachments/files/24189563/opencl-multithread-cpp.txt). As mentioned this (probabalistically) fails in `main`, but is fixed in my PR. I usually run with $10000$ iters to most likely get an error.

### DIFF
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -625,11 +625,11 @@ pocl_cuda_init (unsigned j, cl_device_id dev, const char *parameters)
     {
       result = cuDriverGetVersion(&driver_version);
       if (CUDA_CHECK_ERROR (result, "cuDriverGetVersion"))
-	{
+        {
           ret = CL_INVALID_DEVICE;
-	}
+        }
       else
-	{
+        {
 #if CUDA_VERSION >= 11010
           if (driver_version >= 11010)
             {
@@ -857,9 +857,9 @@ pocl_cuda_init_queue (cl_device_id device, cl_command_queue queue)
   queue_data->use_threads
       = !pocl_get_bool_option ("POCL_CUDA_DISABLE_QUEUE_THREADS", 1);
 
+  PTHREAD_CHECK (pthread_mutex_init (&queue_data->lock, NULL));
   if (queue_data->use_threads)
     {
-      PTHREAD_CHECK (pthread_mutex_init (&queue_data->lock, NULL));
       PTHREAD_CHECK (pthread_cond_init (&queue_data->pending_cond, NULL));
       PTHREAD_CHECK (pthread_cond_init (&queue_data->running_cond, NULL));
       int err = pthread_create (&queue_data->submit_thread, NULL,
@@ -2406,11 +2406,11 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq, int locked)
         {
           for (unsigned int i = 0; i < cmd->svm_free.num_svm_pointers; i++)
             {
-	      void *ptr = cmd->svm_free.svm_pointers[i];
-	      /* This updates bookkeeping associated with the 'ptr'
-		 done by the PoCL core. */
-	      POname (clSVMFree) (event->context, ptr);
-	    }
+          void *ptr = cmd->svm_free.svm_pointers[i];
+          /* This updates bookkeeping associated with the 'ptr'
+         done by the PoCL core. */
+          POname (clSVMFree) (event->context, ptr);
+        }
         }
       break;
     case CL_COMMAND_READ_IMAGE:
@@ -2598,14 +2598,24 @@ pocl_cuda_update_event (cl_device_id device, cl_event event)
     }
 }
 
+/* Waits for event and dependencies to be completed */
+/* Should be called with queue->data locked */
 void
 pocl_cuda_wait_event_recurse (cl_device_id device, cl_event event)
 {
   while (event->wait_list)
     pocl_cuda_wait_event_recurse (device, event->wait_list->event);
 
-  if (event->status > CL_COMPLETE)
-    pocl_cuda_finalize_command (device, event);
+  assert (event->status > CL_COMPLETE);
+  /* If another thread has handled submission, event data might not have been created yet */
+   while (!event->data)
+     ;
+  pocl_cuda_event_data_t *e_d = (pocl_cuda_event_data_t *)event->data;
+  while (!e_d->events_ready)
+    ;
+  assert (event->status == CL_SUBMITTED);
+  pocl_cuda_finalize_command (device, event);
+  assert (event->status == CL_COMPLETE);
 }
 
 void
@@ -2620,9 +2630,10 @@ pocl_cuda_notify_event_finished (cl_event event)
 void
 pocl_cuda_wait_event (cl_device_id device, cl_event event)
 {
+  pocl_cuda_queue_data_t *q_d = (pocl_cuda_queue_data_t *)event->queue->data;
   pocl_cuda_event_data_t *e_d = (pocl_cuda_event_data_t *)event->data;
 
-  if (((pocl_cuda_queue_data_t *)event->queue->data)->use_threads)
+  if (q_d->use_threads)
     {
       /* Wait until background thread marks command as complete */
       POCL_LOCK_OBJ (event);
@@ -2635,8 +2646,11 @@ pocl_cuda_wait_event (cl_device_id device, cl_event event)
     }
   else
     {
-      /* Recursively finalize commands in this thread */
-      pocl_cuda_wait_event_recurse (device, event);
+      PTHREAD_CHECK (pthread_mutex_lock (&q_d->lock));
+      /* Ensure this event is completed */
+      if (event->status > CL_COMPLETE)
+        pocl_cuda_wait_event_recurse (device, event);
+      PTHREAD_CHECK (pthread_mutex_unlock (&q_d->lock));
     }
 }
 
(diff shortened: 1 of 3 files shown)


lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2584,14 +2584,24 @@ pocl_cuda_update_event (cl_device_id device, cl_event event)
...
 pocl_cuda_wait_event_recurse (cl_device_id device, cl_event event)
 {
   while (event->wait_list)
     pocl_cuda_wait_event_recurse (device, event->wait_list->event);
 
-  if (event->status > CL_COMPLETE)
-    pocl_cuda_finalize_command (device, event);
+  assert (event->status > CL_COMPLETE);
+  /* If another thread has handled submission, event data might not have been created yet */
+   while (!event->data)
```
pjaaskel: Should this be an atomic read?

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2584,14 +2584,24 @@ pocl_cuda_update_event (cl_device_id device, cl_event event)
...
+     ;
+  pocl_cuda_event_data_t *e_d = (pocl_cuda_event_data_t *)event->data;
+  while (!e_d->events_ready)
```
pjaaskel: Same here.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2621,8 +2632,11 @@ pocl_cuda_wait_event (cl_device_id device, cl_event event)
     }
   else
     {
```
pjaaskel: Is there a use case for the threaded version? Can we simplify? CUDA's command streams are async anyhow to my understanding, what's the benefit of the added complexity here?

lib/CL/devices/cuda/pocl-cuda.c
pjaaskel: Also check the indent (clang-format-branch.sh)

pjaaskel (commented):
Also please add a mention of this fix to the release notes. Thanks for contributing! Stabilization of the CUDA driver has been long wished for.

lib/CL/devices/cuda/pocl-cuda.c
arrebarritra: No, I am happy to remove the threaded version. I just wasn't sure if we were ready to do that or if we wanted to keep it.

lib/CL/devices/cuda/pocl-cuda.c
arrebarritra: > Should this be an atomic read?

Do you mean the read inside the while loop? I guess that makes sense since the other thread is writing to data.

> Also check the indent (clang-format-branch.sh)

Will do!

lib/CL/devices/cuda/pocl-cuda.c
pjaaskel: @isuruf can you confirm the above?

isuruf: cc @inducer

arrebarritra:
Changes after review:
- Atomic read of event/event data params that might be written to in submitting thread
- Remove threaded queue handling

I haven't been able to run the formatting script yet on my Windows/WSL machine, but I think it will be easier to do it later today when I have access to my linux machine.

After review I can remove/fixup commits as needed depending on if we wan't to keep threaded queue handling or not.

One thing that would be good for someone to have an extra look at is external events. We had a warning that this is not handled for non threaded

> "Can't handle non-CUDA dependencies without queue threads"

but I couldn't see any special handling being done in the threaded handling so I removed it, but I might be wrong.

lib/CL/devices/cuda/pocl-cuda.c
```diff
@@ -2584,46 +2513,36 @@ pocl_cuda_update_event (cl_device_id device, cl_event event)
...
-{
+  assert (event->status > CL_COMPLETE);
+  /* If another thread has handled submission, event data might not have been created yet */
+  while (!POCL_ATOMIC_LOAD (event->data))
+    ;
   pocl_cuda_event_data_t *e_d = (pocl_cuda_event_data_t *)event->data;
-
-  if (((pocl_cuda_queue_data_t *)event->queue->data)->use_threads)
-    PTHREAD_CHECK (pthread_cond_broadcast (&e_d->event_cond));
+  while (!POCL_ATOMIC_LOAD (e_d->events_ready))
```
pjaaskel, line 2554: Is the write to it also atomic?

arrebarritra: Thanks! Now that I think about it, don't we need atomic ops everywhere where the submitting thread writes to variables the finishing thread reads? Even further upstream ex. functions in `pocl_util.c` called by the submitting thread won't we need atomics to flush writes to ex. `event->status` and `event->wait_list`. Or maybe this is already handled, and I just need to read these atomically in `pocl-cuda.c`?

pjaaskel: Right. I'm a bit unsure of the volatile's guarantees in the C standards and whether problems like tearing can occur on some platforms if there're no extra precautions (basically explicit fences etc. produced by atomic loads/stores). I'd expect no problems with aligned variables and coherent caches, so likely not a real issue with this.

franz: Atomics is not the way concurrency is handled in the rest of the pocl codebase. We use per-object locking. Each `cl_mem`, `cl_event`, etc object has its own lock. Some of the driver callbacks are called with an event already locked, this is described in comments of `struct pocl_device_ops` in `pocl_cl.h`.

pjaaskel: You are right that the per object locking is the safest/portable way here, but the question in this particular case was if the volatile could be enough to safely enough just read normally here without locking the object as it just tracks for non null.

arrebarritra: I have given some thought to this discussion and I think the current approach is mostly fine. We just have to be careful since we don't obtain a lock to `event`/`event_data`, but atomic loads should suffice here. We just need to identify any variables in `pocl-cuda.c` that are being read by the finalizing thread that may be concurrently written to by the submitting thread. Here I would add atomic loads/stores. I can add this at some point this week (probably even today).

arrebarritra: @pjaaskel or @franz could you weigh in on the following since I'm not super experienced with CPU multithreading. I'm not sure if I'm overthinking this. In `pocl_cuda_wait_event_recurse` we do:

```c
while (event->wait_list)
  pocl_cuda_wait_event_recurse (device, event->wait_list->event);
```

When a command is enqueued in the submitting thread, `pocl_create_event_sync` creates the `wait_list` and sets its fields. Since we don't acquire a lock on the event in `pocl_cuda_wait_event_recurse`, the submitting thread could have allocated the `wait_list` variable but not yet set `wait_list->event` and we would be passing a `event == nullptr` to `pocl_cuda_wait_event_recurse`. To correct this we would have to do something like this:

```c
while (POCL_ATOMIC_LOAD (event->wait_list))
  {
    while (!POCL_ATOMIC_LOAD (event->wait_list->event))
      ;
    pocl_cuda_wait_event_recurse (device, event->wait_list->event);
  }
```

or something similar with locks (though that seems more expensive). Similarly I would have to update [this bit of code](https://github.com/arrebarritra/pocl/blob/0051c9cd6c0a2e092b85cde2ccadfbd85ec5fe4e/lib/CL/devices/cuda/pocl-cuda.c#L2587-L2592) in `pocl_cuda_join`. We didn't do this before so I'm not sure if I'm thinking correctly here?

franz: TBH if the code has to evolve into these nested atomic while loops to deal with complications, I would just go with the object locking. At least on Linux, a pthread lock is implemented with a [Futex](https://en.wikipedia.org/wiki/Futex), which means a lock operation on uncontested lock will be about as expensive as one atomic cmpxchg instruction plus overhead of a function call. The atomic while loops are harder to reason about, and probably not significantly faster.

pjaaskel: Is this PR pending only this change now?

arrebarritra: Yes, I have not had time to work on it

doc/sphinx/source/notes_7_2.rst
```diff
@@ -36,6 +36,10 @@ CUDA driver
 * Note that CUDA driver does not support LLVM 21, due to a bug
   in upstream Clang code. Users must use LLVM 18 to 20 with CUDA. For details,
   see https://github.com/llvm/llvm-project/issues/154772
+* Make event synchronisation (clFinish and clWaitEvents) thread safe with non
+  threaded queue handling
+* Remove threaded queue handling and POCL_DISABLE_QUEUE_THREADS environment
```
pjaaskel:
```suggestion
* Remove threaded queue handling and POCL_CUDA_DISABLE_QUEUE_THREADS environment 
```

pjaaskel:
Sorry, this was left hanging and didn't make it to 7.2. Can you rebase to the current main and move the release notes to notest_7_3.rst.

arrebarritra:
Sure I'm happy to do that. I see there is no notes_7_3.rst yet though, do you want me to create it?

arrebarritra:
I rebased onto master and created/updated notes for 7.3.

pjaaskel (approved):

PR #2094, opened 2026-01-03
Enforce Vectorization Proposal

dsding2:
# POCL pitch

## Motivation

I would like to propose a new, optional compiler feature to PoCL that seeks to to provide an alternative to the existing PoCL vectorization strategy. The current vectorization strategy relies on various LLVM vectorization passes, which can make vectorization potentially unpredictable, and thus, difficult to rely upon. However, programs in SPMD languages such as OpenCL are implicitly parallel, a property that compilers like [ISPC](https://pharr.org/matt/assets/ispc.pdf) can exploit to consistently enforce vectorization. The proposed compiler feature aims to utilize the same structural parallelism for PoCL, placing vectorization back under user control.

## Proposed Solution
 
I propose to add an optional pass to the compilation pipeline, which will begin by converting the global/local id of workgroup dimension zero to a vector, before recursively changing all dependent instructions to accept a vector as an input, and return a vector as output. Instructions that cannot be vectorized will instead be unrolled. 

Attached is an example of a simple proof of concept. While it is missing several key features, most notably efficient support for loads and stores, as well as the ability to toggle the application of this pass, it should serve to illustrate the intent and approach.

### Future Work

To handle diverging control flow while maintaining vectorization, we plan to utilize masking to determine which vector entries are actually updated. This masking system can also be extended to handle cases where the vector size does not evenly divide the dimension size.

We would also aim to significantly expand the number of efficiently supported instructions, such as allowing vectorized stores and loads to be converted to wide loads and stores or gather/scatter instructions where possible.

## Conclusion
Most of all, with this proof of concept, we are hoping to get your opinion about whether this would fit into PoCL overall. Is this approach worth pursuing further? If so, how should it best be integrated or enabled? Thank you for taking a look.

cc @inducer

### DIFF
--- a/lib/llvmopencl/EnforceVectorization.cc
+++ b/lib/llvmopencl/EnforceVectorization.cc
@@ -23,10 +23,13 @@
 // THE SOFTWARE.
 
 #include "CompilerWarnings.h"
-#include "llvm-c/Core.h"
-#include "llvm/Support/GraphWriter.h"
+#include <llvm/ADT/APInt.h>
+#include <llvm/IR/Constants.h>
 #include <llvm/IR/DerivedTypes.h>
+#include <llvm/IR/Instruction.h>
 #include <llvm/IR/Intrinsics.h>
+#include <unordered_map>
+#include <unordered_set>
 IGNORE_COMPILER_WARNING("-Wmaybe-uninitialized")
 #include <llvm/ADT/Twine.h>
 POP_COMPILER_DIAGS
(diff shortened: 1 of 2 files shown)


lib/llvmopencl/EnforceVectorization.cc
pjaaskel, line 1: I'd just call this WholeFunctionVectorization.cc or similar.

PR #2099, opened 2026-01-19
LLVM 22 support

franz:


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -270,6 +270,23 @@ bool isGVarUsedByFunction(llvm::GlobalVariable *GVar, llvm::Function *F) {
   return false;
 }
 
+void eraseLifetimeMarkers(llvm::Function *F) {
+    // Erase lifetime.start markers which reference invalid values
+    // Only valid values are Poison & Alloca
```
linehill: I think, it's better if the function is named to reflect its intent. Also, maybe turn the comment to documentation.
```suggestion
/// Erase lifetime.start markers which reference invalid values.
///
/// Only valid values are Poison & Alloca.
void eraseInvalidLifetimeMarkers(llvm::Function *F) {
```

PR #2110, opened 2026-02-02
Fix missing LLVM rpath when building with STATIC_LLVM=OFF

### LINKED ISSUE #2098: macOS: libOpenCL fails to load LLVM dependencies with @rpath
On macOS, when building pocl with shared LLVM libraries (STATIC_LLVM=OFF), the resulting libOpenCL.dylib has dependencies on @rpath/libclang-cpp.dylib and @rpath/libLLVM.dylib, but the library itself has no LC_RPATH entries to resolve these @rpath references.
This causes runtime failures like:

```
dyld: Library not loaded: @rpath/libclang-cpp.dylib
  Referenced from: .../libOpenCL.2.15.0.dylib
```

On macOS, @rpath resolution only uses LC_RPATH entries from the dependency chain and does not fall back to DYLD_LIBRARY_PATH. So even if LLVM libraries are in DYLD_LIBRARY_PATH, they won't be found because @rpath can't be resolved without LC_RPATH entries.

Currently working around by manually patching with install_name_tool -add_rpath or changing install names from @rpath/... to bare filenames. 

Has anyone encountered this

Noerr:
## Summary

When building PoCL with dynamic LLVM linking (`STATIC_LLVM=OFF`), the resulting library depends on LLVM shared libraries (`libclang-cpp`, `libLLVM`). However, the library had no rpath entries to locate these dependencies at runtime.

This causes runtime failures unless users manually set `LD_LIBRARY_PATH` (Linux) or `DYLD_LIBRARY_PATH` (macOS) to include LLVM's lib directory.

## The Fix

Adds `INSTALL_RPATH` and `BUILD_RPATH` properties to the library target when `STATIC_LLVM=OFF`, pointing to `LLVM_LIBDIR`.

```cmake
if(NOT STATIC_LLVM AND LLVM_LIBDIR)
  set_target_properties("${POCL_LIBRARY_NAME}" PROPERTIES
    INSTALL_RPATH "${LLVM_LIBDIR}"
    BUILD_RPATH "${LLVM_LIBDIR}")
endif()
```

This is preferable to the global `-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` workaround because:
- It only adds the rpath when actually needed (`STATIC_LLVM=OFF`)
- It targets only the LLVM dependency, not all link directories
- Users don't need to know about the workaround — it just works

## Testing

Tested on macOS (Apple Silicon) with LLVM 21 and chipStar. The built `libOpenCL.dylib` now includes the LLVM lib path in its rpath, and programs using HIP runtime compilation work without requiring manual rpath workarounds.

Fixes #2098

### DIFF
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -39,12 +39,28 @@ if(ENABLE_SPIRV)
   harden(spirv_fix_atomic_compare_exchange)
   target_link_libraries(spirv_fix_atomic_compare_exchange poclu ${OPENCL_LIBS})
   target_include_directories(spirv_fix_atomic_compare_exchange PRIVATE "${SPVSRC}")
+  # Installed executables need rpath to find the PoCL library.
+  # CMAKE_INSTALL_RPATH_USE_LINK_PATH handles external dependencies like LLVM,
+  # but the PoCL library is linked from the build tree, so its install
+  # location must be added explicitly.
+  if(NOT STATIC_LLVM)
+    set_target_properties(spirv_fix_atomic_compare_exchange PROPERTIES
+      INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")
+  endif()
 endif()
 
 if(ENABLE_POCLCC)
   add_executable(poclcc poclcc.c)
   harden(poclcc)
   target_link_libraries(poclcc poclu ${OPENCL_LIBS})
+  # Installed executables need rpath to find the PoCL library.
+  # CMAKE_INSTALL_RPATH_USE_LINK_PATH handles external dependencies like LLVM,
+  # but the PoCL library is linked from the build tree, so its install
+  # location must be added explicitly.
+  if(NOT STATIC_LLVM)
+    set_target_properties(poclcc PROPERTIES
+      INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")
+  endif()
 
   install(TARGETS "poclcc" RUNTIME
           DESTINATION "${POCL_INSTALL_PUBLIC_BINDIR_REL}" COMPONENT "poclcc")
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -39,12 +39,26 @@ if(ENABLE_SPIRV)
   harden(spirv_fix_atomic_compare_exchange)
   target_link_libraries(spirv_fix_atomic_compare_exchange poclu ${OPENCL_LIBS})
   target_include_directories(spirv_fix_atomic_compare_exchange PRIVATE "${SPVSRC}")
+  # On macOS, installed executables need rpath to find the PoCL library.
+  # CMAKE_INSTALL_RPATH_USE_LINK_PATH handles external dependencies like LLVM,
+  # but the PoCL library is linked from the build tree, so its install
+  # location must be added explicitly.
+  if(APPLE AND NOT STATIC_LLVM)
+    set_target_properties(spirv_fix_atomic_compare_exchange PROPERTIES
+      INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")
+  endif()
 endif()
 
 if(ENABLE_POCLCC)
   add_executable(poclcc poclcc.c)
   harden(poclcc)
   target_link_libraries(poclcc poclu ${OPENCL_LIBS})
+  # On macOS, installed executables need rpath to find the PoCL library.
+  # (Same rationale as spirv_fix_atomic_compare_exchange above.)
+  if(APPLE AND NOT STATIC_LLVM)
+    set_target_properties(poclcc PROPERTIES
+      INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")
+  endif()
 
   install(TARGETS "poclcc" RUNTIME
           DESTINATION "${POCL_INSTALL_PUBLIC_BINDIR_REL}" COMPONENT "poclcc")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,6 +330,13 @@ endif()
 
 option(STATIC_LLVM "If ON, link to static LLVM libraries. OFF (default) = link to shared LLVM libraries." ${DEFAULT_STATIC_LLVM})
 
+# When dynamically linking LLVM, add rpaths to installed binaries so they can
+# find LLVM libraries without requiring LD_LIBRARY_PATH/DYLD_LIBRARY_PATH.
+# This is especially important on macOS where DYLD_LIBRARY_PATH doesn't help
+# resolve @rpath references.
+if(NOT STATIC_LLVM)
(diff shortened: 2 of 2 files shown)


Noerr:
Adding second commit because I've realized the executable poclcc experiences the same issue.

Noerr:
# Updated Approach

Thanks for the review  @linehill ! You're right that `CMAKE_INSTALL_RPATH_USE_LINK_PATH` handles the LLVM dependency. I ran some tests to understand the behavior better.

## Test Results

| Approach | libOpenCL.dylib | poclcc | Runtime |
|----------|-----------------|--------|---------|
| Baseline (main) | No rpath | No rpath | - |
| `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` only | LLVM ✅ | LLVM only | **Fails** - can't find PoCL library |
| Updated PR (hybrid) | LLVM ✅ | LLVM + PoCL lib ✅ | Works ✅ |

## Why the executables need explicit rpath

`CMAKE_INSTALL_RPATH_USE_LINK_PATH` adds rpaths based on where libraries were found **at link time**. For `poclcc`:

- LLVM libs are in an external install dir → rpath added automatically ✅
- PoCL library is in the **build tree** at link time → install location not captured ❌

After installation without the added INSTALL_RPATH property, `poclcc` and `spirv_fix_atomic_compare_exchange` can't find the PoCL library because their install location differs from their build location.

## Updated Approach

I've updated the PR to use a hybrid approach:

1. **Enable `CMAKE_INSTALL_RPATH_USE_LINK_PATH`** when `STATIC_LLVM=OFF` - handles LLVM automatically, and only change needed to fix-up PoCL library dependency on LLVM.
2. **Add `POCL_INSTALL_PUBLIC_LIBDIR`** explicitly for executables only - handles PoCL library

This is cleaner than my original approach - no changes to `lib/CL/CMakeLists.txt`, and it relies better on CMake's built-in mechanism for external dependencies.

bin/CMakeLists.txt
```diff
@@ -39,12 +39,28 @@ if(ENABLE_SPIRV)
...
   target_link_libraries(spirv_fix_atomic_compare_exchange poclu ${OPENCL_LIBS})
   target_include_directories(spirv_fix_atomic_compare_exchange PRIVATE "${SPVSRC}")
+  # Installed executables need rpath to find the PoCL library.
+  # CMAKE_INSTALL_RPATH_USE_LINK_PATH handles external dependencies like LLVM,
+  # but the PoCL library is linked from the build tree, so its install
+  # location must be added explicitly.
+  if(NOT STATIC_LLVM)
+    set_target_properties(spirv_fix_atomic_compare_exchange PROPERTIES
+      INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")
+  endif()
```
linehill: I think the repetition of setting the INSTALL_RPATH can by avoided by adding `set(CMAKE_INSTALL_RPATH "${POCL_INSTALL_PUBLIC_LIBDIR}")` before the first `add_executable()` [[1](https://cmake.org/cmake/help/latest/prop_tgt/INSTALL_RPATH.html)] in this file.

linehill (approved):
LGTM but I'm not sure if setting `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` has negative impact on users packaging PoCL for binary distribution.

pjaaskel:
@anbe42 any opinions?

pjaaskel:
@Noerr is this still relevant for you?

Noerr:
Yes, still relevant @pjaaskel  — this is load-bearing for macOS builds and I'd like to see it or something similar land.

On my side it's not hypothetical: my chipStar-on-Apple-Silicon build pipeline cherry-picks this branch on every build, and the resulting `libOpenCL.dylib` carries exactly one added rpath entry — the LLVM libdir — which is what lets `@rpath/libLLVM.dylib` resolve. On macOS `DYLD_LIBRARY_PATH` can't substitute for a missing `LC_RPATH`, so there's no non-invasive workaround short of `install_name_tool` post-processing (which is what Homebrew's long-standing patch avoids). Homebrew installs LLVM and PoCL to non-system prefixes for the same reason this is needed.

On @franz's Fedora concern — I think it doesn't actually materialize, and it's testable rather than a matter of opinion. `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` does **not** add directories that are in `CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES` to the rpath; CMake intentionally filters those out. On Fedora, `libLLVM.so` lives in `/usr/lib64`, which *is* an implicit link directory — so no `/usr/lib64` rpath should be emitted, and the PR is effectively a no-op there. It only adds an rpath when LLVM genuinely lives outside the system link paths (macOS/Homebrew, or any custom `--prefix` build), which is exactly the case where you want it.

That matches @xanderlent's earlier `ldd` output showing the Fedora libs resolve with no rpath at all. @xanderlent — if you still have a moment to build current HEAD *with* this patch, `chrpath -l` / `readelf -d` on the resulting `libpocl.so` should show no `/usr/lib64` DT_RUNPATH, which would settle this concretely.

If the maintainers would still rather be conservative about touching Linux rpaths at all, I'm happy to scope the change so it only takes effect when `LLVM_LIBDIR` is not already a system link directory — but per the above, CMake should already be giving that behavior for free.

franz:
> CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON does not add directories that are in CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES to the rpath; CMake intentionally filters those out

Sorry for delay. CMake does not document this well, but based on my quick search you're right, it's filtered.

However, this filtering does *not* happen on `INSTALL_RPATH` target property. I verified it by building pocl with your PR and installing poclcc:

```
$ readelf -d /usr/bin/poclcc

Dynamic section at offset 0x9c48 contains 29 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libOpenCL.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [/usr/lib/x86_64-linux-gnu]
```

If you change condition on  `set(INSTALL_RPATH ..)` to `if(APPLE AND (NOT STATIC_LLVM))`, I will merge this.

Noerr:
@franz Thanks for testing that — you're right, `INSTALL_RPATH` doesn't get the implicit-directory filtering that `CMAKE_INSTALL_RPATH_USE_LINK_PATH` does. I've scoped the `INSTALL_RPATH` sets to `if(APPLE AND NOT STATIC_LLVM)` and force-pushed. The top-level `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` stays cross-platform since CMake already filters system paths there.

Tested locally: `poclcc` gets both LC_RPATH entries on macOS (LLVM libdir + POCL libdir), `libOpenCL.dylib` gets the LLVM libdir. Everything resolves without `DYLD_LIBRARY_PATH`.

PR #2135, opened 2026-03-02
Implement proper device-specific kernel metadata

jansol:
Hashcat has some kernels with very high register pressure where it checks this value because they'll fail if it tries to run them with the device's general maximum workgroup size (which can be several times higher than for the problematic kernels).

Draft/HACK because this 1) reports the minimum across all devices instead of accurate per-device values (can differ a lot in heterogeneous setups) and 2) quietly overwrites kernel metadata set by other drivers when e.g. building the same kernel for a remote and a proxy device


jansol:
Gave this a stab (well, a few) after some offline discussion, let's what I missed...

PR #2164, opened 2026-04-28
[Host] Implement `cl_khr_kernel_clock` extension

ping-ee:
## Description
This PR implements the `cl_khr_kernel_clock` extension for PoCL's host (CPU) devices. This extension allows OpenCL kernels to sample high-resolution hardware clocks at different scopes (device, work-group, and sub-group), which is essential for fine-grained performance analysis and profiling within kernels.

### 1. OpenCL C Built-ins Implementation
- Added 6 new built-in functions in `lib/kernel/kernel_clock.cl`:
    - Standard `ulong` variants: `clock_read_device()`, `clock_read_work_group()`, `clock_read_sub_group()`.
    - HiLo `uint2` variants: `clock_read_hilo_device()`, `clock_read_hilo_work_group()`, `clock_read_hilo_sub_group()`.
- These functions leverage Clang's `__builtin_readcyclecounter()` to provide high-precision hardware clock data on CPU backends.
- Updated `include/_clang_opencl.h` and `include/_builtin_renames.h` for proper function declaration and internal renaming.

### 2. Runtime & Device Info Support
- **Mandatory Query**: Implemented `CL_DEVICE_KERNEL_CLOCK_CAPABILITIES_KHR` in `clGetDeviceInfo.c` as required by the specification.
- **Initialization**: Updated `struct _cl_device_id` in `pocl_cl.h` and initialized the capabilities in `lib/CL/devices/common_utils.c`.
- **Feature Macros**: Enabled associated OpenCL C feature macros (`__opencl_c_kernel_clock_scope_*`) in `CMakeLists.txt` for host devices.

### 3. Testing & Validation
- **New Kernel Test**: Added `tests/kernel/test_kernel_clock.cl` to verify the functionality of all new built-ins.
- **Runtime Validation**: Enhanced `tests/runtime/test_clGetDeviceInfo.c` to ensure the new device capability query is correctly handled and reports valid data.
- **CI Integration**: Updated `tests/kernel/CMakeLists.txt` to include the new test in the standard test suite.

## Specification Reference
- [Khronos OpenCL SDK: cl_khr_kernel_clock](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/cl_khr_kernel_clock.html)

## Checklist
- [x] Implementation follows the Khronos specification.
- [x] All 6 built-in functions (including HiLo variants) are implemented.
- [x] `clGetDeviceInfo` support for `CL_DEVICE_KERNEL_CLOCK_CAPABILITIES_KHR` is added.
- [x] Added/Updated tests to verify both Kernel and Runtime changes.

### DIFF
--- a/tests/kernel/test_kernel_clock.cl
+++ b/tests/kernel/test_kernel_clock.cl
@@ -1,0 +1,40 @@
+#define HILO_TO_ULONG(v) ((ulong)(v).x | ((ulong)(v).y << 32))
+
+kernel void test_kernel_clock() {
+    ulong dev0 = clock_read_device();
+    ulong dev1 = clock_read_device();
+    ulong wg0 = clock_read_work_group();
+    ulong wg1 = clock_read_work_group();
+    ulong sg0 = clock_read_sub_group();
+    ulong sg1 = clock_read_sub_group();
+
+    ulong dev_hilo0 = HILO_TO_ULONG(clock_read_hilo_device());
+    ulong dev_hilo1 = HILO_TO_ULONG(clock_read_hilo_device());
+    ulong wg_hilo0 = HILO_TO_ULONG(clock_read_hilo_work_group());
+    ulong wg_hilo1 = HILO_TO_ULONG(clock_read_hilo_work_group());
+    ulong sg_hilo0 = HILO_TO_ULONG(clock_read_hilo_sub_group());
+    ulong sg_hilo1 = HILO_TO_ULONG(clock_read_hilo_sub_group());
+
+    if (dev0 > dev1) {
+        printf("Device clocks: FAIL (%lu > %lu)\n", dev0, dev1);
+    }
+    if (wg0 > wg1) {
+        printf("Work-group clocks: FAIL (%lu > %lu)\n", wg0, wg1);
+    }
+    if (sg0 > sg1) {
+        printf("Sub-group clocks: FAIL (%lu > %lu)\n", sg0, sg1);
+    }
+
+    if (dev_hilo0 > dev_hilo1) {
+        printf("Device hilo clocks: FAIL (%lu > %lu)\n",
+               dev_hilo0, dev_hilo1);
+    }
+    if (wg_hilo0 > wg_hilo1) {
+        printf("Work-group hilo clocks: FAIL (%lu > %lu)\n",
+               wg_hilo0, wg_hilo1);
+    }
+    if (sg_hilo0 > sg_hilo1) {
+        printf("Sub-group hilo clocks: FAIL (%lu > %lu)\n",
+               sg_hilo0, sg_hilo1);
+    }
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1659,6 +1659,9 @@ endif()
 set(HOST_DEVICE_CL_VERSION_MAJOR 3)
 set(HOST_DEVICE_CL_VERSION_MINOR 0)
 
+# Probe native host CPU support for cl_khr_kernel_clock
+include(CheckHostCpuKernelClock)
+
 # Host CPU device: list of extensions that are always enabled, for both OpenCL 1.2 and 3.0
 # NOTE: update README.md when changing the list
 set(HOST_DEVICE_EXTENSIONS "cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics \
@@ -1765,6 +1768,18 @@ endif()
 if(HOST_CPU_ENABLE_CL_KHR_FP64)
   set(HOST_DEVICE_EXTENSIONS "${HOST_DEVICE_EXTENSIONS} cl_khr_fp64")
   set(HOST_DEVICE_FEATURES "${HOST_DEVICE_FEATURES} __opencl_c_fp64")
+endif()
+
+if(HOST_CPU_ENABLE_CL_KHR_KERNEL_CLOCK)
+  set(HOST_DEVICE_EXTENSIONS "${HOST_DEVICE_EXTENSIONS} cl_khr_kernel_clock")
+endif()
+
+if(HOST_DEVICE_EXTENSIONS MATCHES "cl_khr_kernel_clock")
+  set(HOST_DEVICE_FEATURES "${HOST_DEVICE_FEATURES} \
+    __opencl_c_kernel_clock_scope_device \
+    __opencl_c_kernel_clock_scope_work_group \
+    __opencl_c_kernel_clock_scope_sub_group"
+  )
 endif()
 
 set(HOST_DEVICE_SPV_EXTENSIONS "+SPV_KHR_no_integer_wrap_decoration,+SPV_KHR_linkonce_odr,+SPV_KHR_expect_assume,+SPV_INTEL_fp_fast_math_mode,+SPV_INTEL_unstructured_loop_controls,+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_memory_access_aliasing,+SPV_INTEL_inline_assembly,+SPV_KHR_integer_dot_product")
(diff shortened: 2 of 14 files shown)


ping-ee:
This is my first contribution to the PoCL project. If I've missed any specific requirements or if there are better ways to implement this, please let me know. Thanks.

tests/kernel/test_kernel_clock.cl
```diff
@@ -0,0 +1,23 @@
...
+    ulong t1 = clock_read_device();
+    ulong t2 = clock_read_work_group();
+    ulong t3 = clock_read_sub_group();
+
+    uint2 hilo1 = clock_read_hilo_device();
+    uint2 hilo2 = clock_read_hilo_work_group();
+    uint2 hilo3 = clock_read_hilo_sub_group();
+
+    // Verification: clocks should generally be non-zero on a running system
+    if (t1 > 0 && t2 > 0 && t3 > 0) {
```
jansol: As a smoke test it would probably be good to also check that the returned timestamps are strictly increasing between subsequent calls to the same clock_* function.

jansol:
Are you planning to make major changes or why did you mark this as a draft?

ping-ee:
> Are you planning to make major changes or why did you mark this as a draft?

I marked it as a draft because I found a small section that needed a minor change. The update has now been completed, and no major changes are planned. Please feel free to take another look—thanks!

franz:
Mostly LGTM, just one thing I'm wondering about is the availability of the builtin. Clang documentation says:

> Query for this feature with __has_builtin(__builtin_readcyclecounter). Note that even if present, its use may depend on run-time privilege or other OS controlled state.

Shall we enable this feature unconditionally for CPU devices ? @ping-ee have you tested on other than x86-64 CPUs. like ARM or RISCV ?

tests/kernel/test_kernel_clock.cl
```diff
@@ -0,0 +1,38 @@
...
+    ulong sg1 = clock_read_sub_group();
+
+    ulong dev_hilo0 = HILO_TO_ULONG(clock_read_hilo_device());
+    ulong dev_hilo1 = HILO_TO_ULONG(clock_read_hilo_device());
+    ulong wg_hilo0 = HILO_TO_ULONG(clock_read_hilo_work_group());
+    ulong wg_hilo1 = HILO_TO_ULONG(clock_read_hilo_work_group());
+    ulong sg_hilo0 = HILO_TO_ULONG(clock_read_hilo_sub_group());
+    ulong sg_hilo1 = HILO_TO_ULONG(clock_read_hilo_sub_group());
+
+    if (!(dev0 > 0 && wg0 > 0 && sg0 > 0)) {
```
linehill: I think this check will fail on CPUs that don't support `__builtin_readcyclecounter()` as stated in Clang documentation:

>  When not supported by the target, the return value is always zero.

tests/kernel/test_kernel_clock.cl
```diff
@@ -0,0 +1,38 @@
...
+        printf("Standard clocks: FAIL (dev=%lu, wg=%lu, sg=%lu)\n",
+               dev0, wg0, sg0);
+    }
+
+    if (!(dev1 > dev0 && wg1 > wg0 && sg1 > sg0)) {
```
linehill: Is it guaranteed that `__builtin_readcyclecounter()` is strictly increasing? This check could be flaky if the implementation of the builtin is monotonic.

ping-ee:
> Mostly LGTM, just one thing I'm wondering about is the availability of the builtin. Clang documentation says:
> 
> > Query for this feature with __has_builtin(__builtin_readcyclecounter). Note that even if present, its use may depend on run-time privilege or other OS controlled state.
> 
> Shall we enable this feature unconditionally for CPU devices ? @ping-ee have you tested on other than x86-64 CPUs. like ARM or RISCV ?

Since I only have x86-64 for local testing and cannot verify the runtime privilege requirements on other architectures, I have two questions regarding the implementation:

1. Does our CI include runners for ARM or RISC-V that could help verify this? If not, it might be safer to restrict this feature for now.

2. Should I use CMake (e.g., check_c_source_compiles) to perform a compile-time test during kernel compilation for this builtin, or should I simply restrict it to the x86-64 platform only to be conservative?

tests/kernel/test_kernel_clock.cl
ping-ee: OpenCL spec says: [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#kernel-clock-functions)

> An idealized clock is an unbounded unsigned scalar integer tick count increasing monotonically over time. A clock’s rate of progress may vary within the lifetime of a work-item, may vary across different executions of the program, and may be affected by conditions beyond the control of the programmer. The sampled value read by this function consists of the least significant bits of the idealized clock’s tick count at the time the instruction was executed. In particular, an observer may see sampled values wrap around zero.

I think we should not assume `cl_khr_kernel_clock` functions is strictly increasing.

franz:
>     1. Does our CI include runners for ARM or RISC-V that could help verify this? If not, it might be safer to restrict this feature for now.

Yes we have both ARM and RISCV 64bit CIs. Looking at them, they  are currently failing. According to [this LLVM issue](https://github.com/llvm/llvm-project/issues/117701), the situation on ARM/RISCV is a mess.
 
>   to perform a compile-time test during kernel 

A CMake compile-time would be good, but `check_c_source_compiles` is not enough; you could compile it and  still get a "illegal instruction" at runtime. The test would have to be compiled & run.

Alternatively, you could make it a CMake `option` which defaults to ON on X86 and OFF on other archs.

ping-ee:
> > ```
> > 1. Does our CI include runners for ARM or RISC-V that could help verify this? If not, it might be safer to restrict this feature for now.
> > ```
> 
> Yes we have both ARM and RISCV 64bit CIs. Looking at them, they are currently failing. According to [this LLVM issue](https://github.com/llvm/llvm-project/issues/117701), the situation on ARM/RISCV is a mess.
> 
> > to perform a compile-time test during kernel
> 
> A CMake compile-time would be good, but `check_c_source_compiles` is not enough; you could compile it and still get a "illegal instruction" at runtime. The test would have to be compiled & run.
> 
> Alternatively, you could make it a CMake `option` which defaults to ON on X86 and OFF on other archs.

I changed this to a host-CPU-specific probe. We now only attempt to enable cl_khr_kernel_clock for native x86 host builds, and the check is compile+run with HOST_CLANG rather than compile-only. Other architectures keep the feature disabled by default for now.

tests/kernel/test_kernel_clock.cl
ping-ee: I updated the test to match the OpenCL CTS behavior for cl_khr_kernel_clock: it now samples each clock twice and only fails if the second sample is less than the first, i.e. it checks for non-decreasing values rather than strict increase. This avoids assuming that two consecutive reads must always return different timestamps.

The CTS does the same thing for both the scalar and hilo variants in each supported scope:
  https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_kernel_clock/kernel_clock.cpp

So the test now aligns with the CTS expectation instead of enforcing a stronger condition that could be flaky.

ping-ee:
Thanks for the review and suggestions earlier!

I've addressed the comments and force-pushed the updated version.
This PR should now be ready for another round of review.

Please let me know if there's anything else I should improve 🙂

pjaaskel:
I think we should pull this in for the next release that will be tagged pretty soon. Can you add a mention of this feature in the 7.2 release notes file? Also, if you have used AI assistance, please mention the model etc. in your commit message.

linehill (approved):
LGTM.

pjaaskel:
Thanks!

PR #2182, opened 2026-05-29
WorkitemLoops: fix stale global id in the peeled work-item

maleadt:
When a kernel returns from some work-items before a barrier, WorkitemLoops peels work-item (0,0,0) into its own copy so it only resolves the branch once. The peeled copy's prologue resets the local ids to 0, but the global id counters (`_global_id_{x,y,z}`) are running globals that `createLoopAround()` only re-bases for the loop iterations. So the peeled copy kept whatever value the loop left behind, and `get_global_id()` returned the work-group size instead of 0 for the first work-item. That work-item then read or wrote the wrong element.

The fix re-bases the global ids next to the local ids in the peeled-region prologue.

This came from Julia, where we've recently switched to lowering unreachable control flow to `ret` from the kernel, see https://github.com/pocl/pocl/pull/2126#issuecomment-4562039891

x-ref https://github.com/JuliaGPU/KernelAbstractions.jl/pull/700

## Test

`tests/regression/test_peeled_wi_global_id` (loopvec and cbs), reduced from a Julia kernel with `llvm-reduce`. Per work-item:

```
if (get_local_id(0) > 7) return;   // dead for local size 8
barrier();
A[get_global_id(0)] = get_global_id(0) + 1;
```

Run as `global = local = 8`, the result must be `{1,..,8}`. Before the fix, element 0 was left untouched.


pjaaskel:
Thanks. WILoop peeling was removed in the new kernel compiler (LVN), so this bug might have disappeared there or needs to be fixed again.

pjaaskel:
The two CIs do not like the spv test case, them not finding the barrier. Perhaps a SPIRV wrapper issue @franz ?

franz:
The problem is that the test doesn't check the presence of `cl_khr_subgroup`. Something like this should work (copypasted from another test):

```C
  if (poclu_supports_extension(Devices.at(0).get(), "cl_khr_subgroups") == 0) {
      std::cout << "this test requires cl_khr_subgroups, test SKIPPED\n";
      return 77;
  }
```

pjaaskel (approved):

PR #2188, opened 2026-06-04
`clSetKernelExecInfo`: fix out-of-bounds read of the USM indirect pointer list.

maleadt:
The `CL_KERNEL_EXEC_INFO_USM_PTRS_INTEL` branch of `clSetKernelExecInfo` passes `param_value_size` (a byte count) to `pocl_reset_indirect_ptrs()`, which expects an element count. It therefore iterates past the end of the application's pointer array, reading uninitialized/unmapped memory. With a large indirect-pointer list this runs off the allocation and segfaults; with a small one it silently reads garbage (and can spuriously register stray words as indirect USM pointers).

The SVM branch a few lines up already divides by `sizeof(void*)` correctly; this just brings the USM branch in line.

Introduced in 588c6cc8a (#1876); affects v7.0, v7.1 and main.

Reported downstream as JuliaGPU/OpenCL.jl#317 (frequent crashes in Windows CI when launching kernels with USM buffers; the heap layout there makes the over-read fault, whereas it usually stays within mapped memory on Linux).

### DIFF
--- a/doc/sphinx/source/notes_7_2.rst
+++ b/doc/sphinx/source/notes_7_2.rst
@@ -22,6 +22,10 @@ Notable bugfixes
   (some corner-cases in clSetKernelArg, clSetKernelExecInfo, clCreateContext,
    clCreateContextFromType, clGetDeviceIDs, clSetKernelArgSVMPointer,
    clEnqueueNDRange with local_size == NULL and nonzero reqq-wg-size)
+* Fixed an out-of-bounds read in `clSetKernelExecInfo` when registering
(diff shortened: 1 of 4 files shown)


pjaaskel (changes requested):
Great catch, thanks! Can you add this significant bug to the ReleaseNotes? Also can you add "Co-authored by:" in the commit for your LLM and its model etc. for traceability?

maleadt:
Done.

pjaaskel (approved):

pjaaskel:
Thanks!

PR #2189, opened 2026-06-08
Support RENAME_POCL on macOS

maleadt:
On Apple, pocl_cl.h hardcoded `POname(name) name`, so building with -DENABLE_ICD=OFF -DRENAME_POCL=ON had no effect there: the library still exported the unprefixed cl* entrypoints instead of the PO-prefixed ones.

Gate the Apple symbol macros on `RENAME_POCL && !BUILD_ICD`, mirroring the ELF path: rename to PO<name>, forward-declare the PO-prefixed entrypoints via POdeclsym (so internal callers see a prototype), and export them through POCL_EXPORT's default visibility. The ICD build (BUILD_ICD defined) is unaffected. Apple has no ELF-style aliases, so the unprefixed cl* names are not provided in this mode (as is already the case on ELF with RENAME_POCL).

### DIFF
--- a/lib/CL/pocl_cl.h
+++ b/lib/CL/pocl_cl.h
@@ -262,9 +262,22 @@ extern pocl_obj_id_t last_object_id;
 #ifdef __APPLE__
 /* Note: OSX doesn't support aliases because it doesn't use ELF */
 
+#if defined(RENAME_POCL) && !defined(BUILD_ICD)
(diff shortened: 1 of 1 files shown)


maleadt:
This has been tested downstream, and seems like it's working.

tneval (approved):
Looks good to me.

PR #2192, opened 2026-06-09
Apply #1929 to `main`

maleadt:
https://github.com/pocl/pocl/pull/1929 was only applied to `release_7_0`, and in Julia we've been carrying this patch for our 7.1 build. Is there a reason it was not applied to `main`?


franz:
> Is there a reason it was not applied to main?

Probably just forgot.

franz (approved):

PR #2194, opened 2026-06-10
Link CPU kernel binaries in-process through lld

maleadt:
https://github.com/pocl/pocl/pull/2190 was getting unwieldy, so I split off one of the changes: generalize the use of in-process `lld` to link kernels. It currently was only enabled for MSVC, but I see no reason to not use it on other platforms, where it also avoids having to call into `clang` or `lld`. One exception is MinGW, but that's a pre-existing bug (which I hope to look into later).

---

The CPU drivers used to hand each compiled kernel object to the Clang driver, which execs a linker and pulls in the C startup files, so kernel compilation needed a host toolchain at run time. Windows (MSVC) builds already avoided that by calling lld-link through its library API.

Generalize that path to ELF and Mach-O hosts: when lld is available at build time (CPU_USE_LLD_LINK, set up in cmake/SetupLLD.cmake), kernel objects are linked into shared libraries in-process, with the Clang driver kept as a fallback. Symbols the kernel binary leaves undefined resolve at dlopen() time, so the link needs no C runtime or startup files.

Related changes:
- replace the CPU drivers' finalize_binary hook with a direct call in the common codegen path, extracted into pocl_link_final_binary()
- mark the work-group function dllexport on MinGW triples too (instead of MSVC only), since lld-link never auto-exports and MinGW linkers stop auto-exporting once any symbol is marked
- fix the _fltused triple check for LLVM 21's Triple-returning Module::getTargetTriple()


maleadt:
Okay, I think this is ready for review. The main changes are:

- Adds `cmake/SetupLLD.cmake` to decide whether `CPU_USE_LLD_LINK` can be enabled and to locate the right lld libraries.
- Refactors final binary linking into `pocl_link_final_binary()`: try device-specific finalize hooks first, then in-process lld, then Clang driver fallback.
- Implements `pocl_invoke_lld_link()` for COFF, ELF, and Mach-O.
- Adjusts CMake link lines so `libpocl` can link lld correctly, especially with static LLVM and CMake-package LLVM discovery.
- Preserves absolute-path vector math libraries such as SVML/SLEEF/libmvec for the lld path.
- Keeps Windows/MSVC runtime-free kernel DLL linking with bundled helper objects.
- Marks work-group entry points `dllexport` for all Windows triples, not only MSVC.
- Adds a fallback to Clang when unsupported final linker flags are provided
- Updates docs/release notes to describe the new CPU kernel linking behavior.

maleadt:
Any update here? This shouldn't be too controversial, as it generalizes an existing mechanism (LLD was already being used for MSVC). This would allow me to move forwards on https://github.com/pocl/pocl/pull/2190, which greatly simplifies our use of PoCL in Julia.

franz (approved):
LGTM, however it updates documentation for 7.2 notes - the current main branch is 8.0-pre.

@maleadt  do you want this backported into to 7.2 ?

pjaaskel:
I think this and the FP16 updates could go to 7.2. @franz  I'd just merge again from main and revert the version change commit. @maleadt after this is pulled in, can you send a PR that enables this by default for all LLVMs where it works?

franz:
@pjaaskel IIUC this is enabled by default. However it depends on presence of lld libraries & cmake files (`liblld-XY-dev` on Ubuntu), not the version of LLVM.

maleadt:
Thanks!

> do you want this backported into to 7.2 ?

Ideally yes, but I'm maintaining my own backport branches anyway, so it's not strictly required.

> IIUC this is enabled by default. However it depends on presence of lld libraries & cmake files (`liblld-XY-dev` on Ubuntu), not the version of LLVM.

Correct. The API is available from LLVM 17 onwards, which is the lower-bound of what PoCL supports anyway, so I didn't bothered adding a version gate.

franz:
OK, this applies cleanly on 7.2 branch, is relatively small and useful, so I'm adding it. Also the MacOS RENAME_POCL PR.

I'm holding off merging recent FP16 PRs into 7.2 branch for now, since it's a lot of churn and doesn't (yet) make much  difference on the overall status of FP16 support.

FWIW, I tested today the (almost) full OpenCL-CTS (on CPU) with main branch + LLVM 23 support + a build of LLVM 23, and it passed all tests. Since LLVM 23 support commits are quite small, we can add them to the 7.2 branch.

However, I haven't tested CUDA and LevelZero with LLVM 23; I expect at least CUDA might require some fixes (LevelZero probably just works since it inputs SPIRV).

Issue #2181, opened 2026-05-28
`clGetKernelSubGroupInfo` ignores `intel_reqd_sub_group_size`


maleadt:
When setting `intel_reqd_sub_group_size`, the host-side and device-side sub-group queries seem to disagree.

<details>

<summary>MWE</summary>

```c
#define CL_TARGET_OPENCL_VERSION 210
#include <CL/cl.h>

#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static const char *SRC_WITH =
    "__attribute__((intel_reqd_sub_group_size(32)))\n"
    "__kernel void f(__global uint *a) {\n"
    "    a[get_global_linear_id()] = get_num_sub_groups();\n"
    "}\n";

static const char *SRC_WITHOUT =
    "__kernel void f(__global uint *a) {\n"
    "    a[get_global_linear_id()] = get_num_sub_groups();\n"
    "}\n";

#define CHECK(expr) do {                                                       \
    cl_int _err = (expr);                                                      \
    if (_err != CL_SUCCESS) {                                                  \
        fprintf(stderr, "%s:%d: %s -> %d\n", __FILE__, __LINE__, #expr, _err); \
        exit(1);                                                               \
    }                                                                          \
} while (0)

static const size_t LOCAL_SIZES[][3] = {
    {1, 1, 1}, {3, 1, 1}, {1, 3, 1}, {32, 1, 1}, {64, 1, 1},
};

void list_platforms(void) {
    cl_uint num_platforms;
    cl_int err = clGetPlatformIDs(0, NULL, &num_platforms);
    CHECK(err);

    cl_platform_id *platforms = (cl_platform_id *)malloc(sizeof(cl_platform_id) * num_platforms);
    err = clGetPlatformIDs(num_platforms, platforms, NULL);
    CHECK(err);

    printf("Available platforms:\n");
    for (cl_uint i = 0; i < num_platforms; i++) {
        char platform_name[128];
        err = clGetPlatformInfo(platforms[i], CL_PLATFORM_NAME,
                                sizeof(platform_name), platform_name, NULL);
        CHECK(err);
        printf("  %d: %s\n", i, platform_name);
    }
    free(platforms);
}

cl_platform_id get_platform(int index) {
    cl_uint num_platforms;
    cl_int err = clGetPlatformIDs(0, NULL, &num_platforms);
    CHECK(err);

    if ((cl_uint)index >= num_platforms) {
[...]

PR #2195, opened 2026-06-10
Make host- and device-side sub-group queries agree

maleadt:
When a kernel sets `intel_reqd_sub_group_size`, `clGetKernelSubGroupInfo` ignored the attribute (it assumes SG == WG_x) while the device-side builtins honored it, so host and device disagreed. That breaks code which sizes per-sub-group state from the host query and indexes it on the device with `get_sub_group_id()`, e.g. the random number generator we have in OpenCL.jl.

The device-side builtins also got non-divisible work-group sizes wrong: `get_num_sub_groups()` truncated, returning 0 for work-groups smaller than the sub-group size, and `get_sub_group_size()` didn't account for a smaller trailing sub-group, which also made the sub-group reductions/scans/ballot read past the end of the work-group. `get_enqueued_num_sub_groups()` hardcoded 1, and the CUDA `get_num_sub_groups()` counted the whole NDRange rather than one work-group.

The first commit fixes the kernel library: the sub-group count rounds up, and `get_sub_group_size()` returns the actual size of the current sub-group. The second commit extracts `intel_reqd_sub_group_size` into the kernel metadata (also stored in pocl binaries) and uses it in the CPU drivers' sub-group queries. It also implements `CL_KERNEL_COMPILE_SUB_GROUP_SIZE_INTEL`, and adds a regression test that checks host/device agreement with and without the attribute; the test fails without these fixes.

The CPU device now gives the same answers as Intel's OpenCL implementations for the MWE in #2181.

cc @christiangnrd

### DIFF
--- a/lib/CL/clGetKernelSubGroupInfo.c
+++ b/lib/CL/clGetKernelSubGroupInfo.c
@@ -70,7 +70,8 @@ CL_API_ENTRY cl_int CL_API_ENTRY POname (clGetKernelSubGroupInfo) (
   POCL_RETURN_ERROR_ON ((dev_i == CL_UINT_MAX), CL_INVALID_KERNEL,
                         "the kernel was not built for this device\n");
 
-  if (param_name == CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE)
+  if (param_name == CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE
+      || param_name == CL_KERNEL_SUB_GROUP_COUNT_FOR_NDRANGE)
     {
       POCL_RETURN_ERROR_ON ((input_value == NULL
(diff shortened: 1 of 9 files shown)


pjaaskel:
@tneval I recall you recently fixed this in LVN? Can you check this to avoid conflicts.

tneval:
> @tneval I recall you recently fixed this in LVN? Can you check this to avoid conflicts.

I have implemented some parts of this in the PR that is WIP on LVN. I think this can be merged here, but I'll give it a closer look.

tneval (approved):
This looks solid. 
Should we also validate that the `reqd_sub_group_size` is somehow legal? 
IIRC, the kernel compiler only supports single fixed subgroup size. Having smaller 'leftover' subgroup will probably break something (At least this is the case in LVN). If this is the case, it can be addressed later.

PR #2196, opened 2026-06-16
SubCFGFormation: fall back to work-item loops when a kernel has no exit

maleadt:
canHandleKernel() only checked for barriers, so a kernel that has a barrier but no exit block (an infinite loop like for(;;)) was still handed to CBS. formSubCfgs() then takes the "Invalid kernel! No kernel exits!" path and calls llvm_unreachable, which lowers to a trap without assertions.

The work-item loops generator already compiles these kernels (the regression/infinite_loop test relies on it), and the chooser already sends barrier-free kernels there. Send kernels with no exit block there too.

Fixes the failure seen in https://github.com/pocl/pocl/pull/2194


pjaaskel (approved):
Thanks!

PR #2197, opened 2026-06-16
WorkitemHandlerChooser: keep the chosen handler stable

maleadt:
The work-item handler is chosen once, before any kernel transformation, and the whole work-group pipeline commits to it. As a plain cached analysis it could be recomputed mid-pipeline: ImplicitLoopBarriers adds implicit barriers (for the work-item-loops path) that flip hasWorkgroupBarriers(), so a recompute picks CBS for a kernel originally classified LOOPS. Whether that recompute happens is incidental, environment-dependent cache invalidation -- which surfaced regressions on some hosts but not others. Pin the result so it is never invalidated.

I think this is the reason https://github.com/pocl/pocl/pull/2196 happened on https://github.com/pocl/pocl/pull/2194 but not on main.

### DIFF
--- a/lib/llvmopencl/WorkitemHandlerChooser.cc
+++ b/lib/llvmopencl/WorkitemHandlerChooser.cc
@@ -109,12 +109,18 @@ WorkitemHandlerChooser::run(llvm::Function &F,
 }
 
 bool WorkitemHandlerResult::invalidate(
-    llvm::Function &F, const llvm::PreservedAnalyses PA,
-    llvm::AnalysisManager<llvm::Function>::Invalidator &Inv) {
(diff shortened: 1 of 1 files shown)


linehill:
In out private branch we have made the work-item (WI) method selection sticky. Also, the selector is no longer an analysis pass so our approach to remember the chosen WI method is to store it in a function attribute. Once the work in the private branch is good enough, we pull it on the main branch.

That said, the patch is so small that I don't think its a big deal to pull it in.

linehill (approved):
LGTM. 

Let's pull this in after the CI issues have been solved which probably means pulling #2196 first.

PR #2198, opened 2026-06-16
cmake: gate the SIGFPE handler on x86-64, not all x86

maleadt:
The SIGFPE handler (lib/CL/devices/signal_handlers.c) skips the faulting floating-point instruction from within the handler, which is implemented only for x86-64: the relevant code, including pocl_ignore_sigfpe_for_thread(), is guarded by `#ifdef __x86_64__`. The CMake default was gated on X86, which also matches 32-bit x86 (i686). There the option defaulted on, so the pthread scheduler compiled its `#ifdef ENABLE_SIGFPE_HANDLER` call to pocl_ignore_sigfpe_for_thread(), which has no definition on i686.

This stays latent while the CPU driver is a separate dlopen()ed module (shared objects tolerate undefined symbols), but breaks linking once the driver is linked into libpocl (ENABLE_LOADABLE_DRIVERS=OFF): the poclcc and example executables then fail with
`undefined reference to pocl_ignore_sigfpe_for_thread`.


maleadt:
CI failure unrelated, fixed in https://github.com/pocl/pocl/pull/2196.

franz (approved):

PR #2205, opened 2026-06-23
Fix EmitPrintf.cc compilation with LLVM 23.0

woodychow:
This change fixes a compilation error in EmitPrintf.cc due to a constructor signature change for llvm::DataExtractor in the modern LLVM project (which added the 2-argument constructor and removed the 3-argument one): https://github.com/llvm/llvm-project/commit/7319a3c

An LLVM version guard is added in EmitPrintf.cc because LLVM versions before 23 do not have a 2-argument constructor for llvm::DataExtractor (see https://github.com/llvm/llvm-project/blob/release/17.x/llvm/include/llvm/Support/DataExtractor.h). The 2-argument constructor is supported in LLVM 23.0 (see https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/DataExtractor.h).

### DIFF
--- a/lib/llvmopencl/EmitPrintf.cc
+++ b/lib/llvmopencl/EmitPrintf.cc
@@ -399,7 +399,11 @@ static void processConstantStringArg(StringData *SD, IRBuilder<> &Builder,
                                      SmallVectorImpl<Value *> &WhatToStore) {
   std::string Str(SD->Str.str() + '\0');
 
+#if LLVM_MAJOR >= 23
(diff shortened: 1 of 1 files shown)


woodychow:
~I still have some LLVM 22/23 compile fixes. I will send out PRs for them soon. Thanks!~

Actually, https://github.com/pocl/pocl/pull/2204 covers them. Nice.

franz (approved):

PR #2206, opened 2026-06-23
Fix floating point to normalized integer rounding in pocl_image_util.c

woodychow:
This change implements true round-to-nearest-even (RTE) rounding for float-to-normalized-integer image conversions in pocl_image_util.c by using rintf instead of truncation (which was performed via direct integer casts). See: https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#converting-floating-point-values-to-normalized-integer-channel-data-types

A comprehensive unit test is added to test both round-up-to-even and round-down-to-even halfway cases across all OpenCL-supported normalized formats (signed/unsigned 8 and 16-bit integers).

### DIFF
--- a/lib/CL/pocl_image_util.c
+++ b/lib/CL/pocl_image_util.c
@@ -25,6 +25,7 @@
 #include "assert.h"
 #include "pocl_cl.h"
 #include "pocl_cl_half_util.h"
+#include <math.h>
 
 cl_int pocl_opencl_image_type_to_index (cl_mem_object_type  image_type)
 {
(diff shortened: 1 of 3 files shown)


franz (changes requested):
LGTM overall, however there are some missing things:

1) the test needs to be added to `tests/runtime/CMakeLists.txt` so that it's actually built & run

2) PoCL compiled with ENABLE_CONFORMANCE supports less image formats than the non-conformant build. So clCreateImage can fail with CL_INVALID_OPERATION or CL_INVALID_IMAGE_FORMAT_DESCRIPTOR in which case it should skip the test.

woodychow:
Updated. Please take another look. Thanks!

franz (approved):
LGTM, test is passing now

PR #2210, opened 2026-06-24
Add half (fp16) conversion functions

woodychow:
Add explicit type conversion functions (e.g. convert_halfN, convert_typeN(halfN)) between half (fp16) types and other OpenCL C types.

Why this patch is needed:
1. convert_type.py needs to map OpenCL C 'half' vector boolean comparisons to 'shortN' (16-bit signed vector types) instead of 'intN' (32-bit vector types) to comply with the OpenCL C specification for relational operators on 16-bit float vectors.
   See: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#relational-and-logical-operators
2. It generates explicit convert_half*() and convert_*_sat_*(half*) conversion functions.
   See: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#explicit-conversions

### DIFF
--- a/lib/kernel/convert_type.cl
+++ b/lib/kernel/convert_type.cl
@@ -53937,3 +53937,6069 @@ double16 convert_double16_rtn(double16 x)
   return res;
 }
 #endif
+
+#if defined(cl_khr_fp16)
+_CL_ALWAYSINLINE _CL_OVERLOADABLE _CL_READNONE
+half convert_half_rtz(char x)
+{
+  return convert_half(x);
+}
+#endif
+
+#if defined(cl_khr_fp16)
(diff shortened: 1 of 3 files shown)


franz:
@maleadt  is there a way to make Julia CI to not treat "unexpected pass" test results as an error ?

maleadt:
> is there a way to make Julia CI to not treat "unexpected pass" test results as an error ?

Not easily, it's there by design to spot changes in test behavior. But I could improve the condition on which things are marked "broken" (e.g. once this is merged) or temporarily skip those tests for now (e.g. on pocl-dev in general).

woodychow:
Updated. Please take another look!

pjaaskel (approved):

pjaaskel:
Thanks!

PR #2211, opened 2026-06-24
Fix float-to-half rounding behavior in pocl_cl_half_util.c

woodychow:
This change fixes the buggy custom bitwise emulation logic in pocl_float_to_half to use standard, correct calls to the official Khronos cl_half.h library functions, enforcing proper round-to-nearest-even (RTE) rounding during float-to-half conversion.

The original bitwise helper truncated floating-point values instead of performing IEEE-754 rounding.

According to the OpenCL C Programming Language Specification:
- Conversions to floating-point types follow IEEE-754 rounding rules.
- If no rounding mode is specified, the default rounding mode is round-to-nearest-even (RTE). See Section "Explicit Conversions": https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#explicit-conversions

### DIFF
--- a/lib/CL/pocl_cl_half_util.c
+++ b/lib/CL/pocl_cl_half_util.c
@@ -23,89 +23,17 @@
    IN THE SOFTWARE.
 */
 
-/**
- * TODO: see if this library can be dropped for the include/CL/cl_half.h
- *  library. The question is which rounding mode should be used in that
- *  case.
- */
-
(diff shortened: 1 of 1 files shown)


franz (approved):

woodychow:
Added `Co-authored-by: Gemini 3.5 Flash` to commit message.

PR #2214, opened 2026-06-24
Fail commands whose wait-list dependency is already failed

bricevideau-ai:
[pocl-failed-dep-repro.zip](https://github.com/user-attachments/files/29291305/pocl-failed-dep-repro.zip)
## Problem

A command whose wait-list dependency is **already in a failed (negative) status** at the moment its event-sync edge is wired runs anyway, on memory the terminated dependency already released. On the CPU (pthread) device this manifests as a SIGSEGV in a worker thread:

```
__memcpy_evex_unaligned_erms
pocl_exec_command            (libpocl)
pocl_pthread_driver_thread   (libpocl-devices-pthread)
```

It can also silently *succeed*, having run on terminated data. Per the OpenCL spec (§5.10, `clSetUserEventStatus` / event execution status), completing an event with a negative value terminates it, and commands waiting on it must themselves be terminated with the error propagated — they must **not** execute.

## Root cause

`pocl_create_event_sync` treated an already-**failed** notifier identically to a **completed** one — skipping creation of the `wait_list` edge in both cases:

```c
if (notifier_event->status < 0 || notifier_event->status == CL_COMPLETE)
    goto FINISH;   /* "it can start right away" */
```

That is correct for `CL_COMPLETE`, but wrong for a negative status: with no edge, the waiter ends up with an empty `wait_list`, `pocl_command_is_ready()` returns true, and the device submit path runs the command. pocl already handles a dependency that fails *after* the edge exists (via `pocl_broadcast` → `device->ops->notify` → `pocl_update_event_failed`); only the *already-failed-at-wiring-time* case bypassed it.

## Fix

The waiter cannot be failed inline in `pocl_create_event_sync` (it runs under the command-queue lock held across the `create_event_sync` calls in `pocl_command_enqueue`; `pocl_update_event_failed` re-locks the same cq → deadlock). So the failure is **recorded and acted on later**, mirroring how pocl already defers failure propagation:

- Add `cl_event::broken_dependency`, initialized to 0 in `pocl_create_event` (events recycled from the mem-manager free list aren't memset, so the reset is explicit).
- Split the condition in `pocl_create_event_sync`: `CL_COMPLETE` keeps the safe skip; `status < 0` sets `waiting_event->broken_dependency = 1` (under the event lock) and skips the edge.
- `pocl_broadcast` also sets `broken_dependency` on the target when the notifier finishes failed, closing the window where a command becomes ready via a *different* completed dependency while a concurrent failed-dependency broadcast is mid-flight.
- The **pthread** and **basic** device submit/notify paths honor the flag before pushing a ready command, failing it via `pocl_update_event_failed` with the same unlock/fail(/relock) dance the existing failed-finished notify branch uses. A `pocl_command_honor_broken_dependency()` helper encapsulates the submit-path dance.

## Validation

Reproduced with a minimal concurrent reproducer (a host-callback graph that aborts its downstream by completing a wait-list user event negative), built against this same source with and without the fix:

| | abort path ×60 | success control |
|---|---|---|
| **before** (clean main) | 35 pass / 19 ran-on-garbage / **6 SIGSEGV** | — |
| **after** (this PR) | **60 pass** / 0 fail / 0 crash | 10/10 |

After the fix the downstream command aborts cleanly with `-14` (`CL_EXEC_STATUS_ERROR_FOR_EVENTS_IN_WAIT_LIST`) instead of crashing or running on terminated data; the success path still completes normally.

**Full pocl test suite: 311/311 passed (12 skipped), no regressions.**

### DIFF
--- a/lib/CL/devices/basic/basic.c
+++ b/lib/CL/devices/basic/basic.c
@@ -543,6 +543,19 @@ pocl_basic_submit (_cl_command_node *node, cl_command_queue cq)
     }
 
   node->state = POCL_COMMAND_READY;
+
+  /* If a wait-list dependency has failed, this command must not run: fail it via
+   * the error cascade. There is no point waiting for it to become otherwise
+   * ready first. pocl_update_event_failed takes the cq + event locks itself, so
+   * unlock the event we hold and return without unlocking again. */
+  if (node->sync.event.event->failed_dependency)
+    {
+      POCL_UNLOCK_OBJ (node->sync.event.event);
+      pocl_update_event_failed (CL_FAILED, NULL, 0, node->sync.event.event,
+                                NULL);
+      return;
+    }
+
   POCL_LOCK (d->cq_lock);
   pocl_command_push(node, &d->ready_list, &d->command_list);
 
@@ -580,7 +593,10 @@ pocl_basic_notify (cl_device_id device, cl_event event, cl_event finished)
   pocl_basic_data_t *d = (pocl_basic_data_t *)device->data;
   _cl_command_node * volatile node = event->command;
 
-  if (finished->status < CL_COMPLETE)
+  /* Fail the command if the notifier finished failed, or if any wait-list
+   * dependency has already failed (recorded on the event). In either case the
+   * command MUST NOT execute, so there is no point checking readiness first. */
+  if (finished->status < CL_COMPLETE || event->failed_dependency)
     {
       /* Unlock the finished event in order to prevent a lock order violation
        * with the command queue that will be locked during
--- a/lib/CL/devices/common.c
+++ b/lib/CL/devices/common.c
@@ -871,6 +871,20 @@ pocl_broadcast (cl_event brc_event)
             }
         }
 
+      /* If the notifier finished in a failed (negative) status, the target
+       * command must NOT run: per the OpenCL spec a command whose wait-list
+       * event has a negative status must be terminated with the error cascade.
+       * The device notify call below fails the target directly when it is in a
+       * failable state. But the target may transiently be in another state
+       * (e.g. a concurrent broadcast for a *different*, completed dependency is
+       * mid-flight and about to submit/push it), so also record the failed
+       * dependency on the target; the device submit/notify paths check the flag
+       * and fail the command rather than execute it on freed/aborted memory.
+       * We hold the target's event lock here, so this is safe against those
+       * readers. */
+      if (brc_event->status < 0)
+        target->event->failed_dependency = 1;
+
       if ((target->event->status == CL_SUBMITTED)
           || (target->event->status == CL_QUEUED))
         {
--- a/lib/CL/devices/pthread/pthread.c
+++ b/lib/CL/devices/pthread/pthread.c
@@ -216,6 +216,17 @@ void
 pocl_pthread_submit (_cl_command_node *node, cl_command_queue cq)
 {
   node->state = POCL_COMMAND_READY;
+  /* If a wait-list dependency has failed, this command must not run: fail it via
+   * the error cascade. There is no point waiting for it to become otherwise
+   * ready first. pocl_update_event_failed takes the cq + event locks itself, so
+   * unlock the event we hold and return without unlocking again. */
+  if (node->sync.event.event->failed_dependency)
+    {
+      POCL_UNLOCK_OBJ (node->sync.event.event);
+      pocl_update_event_failed (CL_FAILED, NULL, 0, node->sync.event.event,
+                                NULL);
+      return;
+    }
   if (pocl_command_is_ready (node->sync.event.event))
     {
       pocl_update_event_submitted (node->sync.event.event);
@@ -256,7 +267,10 @@ pocl_pthread_notify (cl_device_id device, cl_event event, cl_event finished)
 {
   _cl_command_node *node = event->command;
 
-  if (finished->status < CL_COMPLETE)
+  /* Fail the command if the notifier finished failed, or if any wait-list
+   * dependency has already failed (recorded on the event). In either case the
+   * command MUST NOT execute, so there is no point checking readiness first. */
+  if (finished->status < CL_COMPLETE || event->failed_dependency)
     {
       /* Unlock the finished event in order to prevent a lock order violation
        * with the command queue that will be locked during
--- a/lib/CL/pocl_cl.h
+++ b/lib/CL/pocl_cl.h
@@ -2137,6 +2137,28 @@ struct _cl_event {
 
   /* The execution status of the command this event is monitoring. */
   cl_int status;
+
+  /* Set to 1 when one of this command's wait-list dependencies has failed
+   * (finished with a negative status) and therefore this command must not run
+   * but must instead be terminated with the error cascade (OpenCL spec: a
+   * command whose wait-list event has a negative status is terminated and the
+   * error propagated). It is set in two situations the normal
+   * broadcast -> notify failure path cannot cover on its own:
+   *   - In pocl_create_event_sync, when the dependency is ALREADY failed at the
+   *     moment the sync edge would be wired. No edge is created (the notifier
+   *     has already broadcast and will not do so again), so without this flag
+   *     the waiter would be left ready with an empty wait_list and run.
+   *   - In pocl_broadcast, when a dependency finishes failed but the target is
+   *     transiently not in a failable state (e.g. a concurrent broadcast for a
+   *     different, completed dependency is mid-flight), so the immediate notify
+   *     cannot fail it.
+   * The device submit and notify paths check this flag and fail the command via
+   * pocl_update_event_failed. The flag (rather than an inline fail at the set
+   * sites) is required because both set sites hold locks (the command-queue lock
+   * in pocl_create_event_sync, the event lock in pocl_broadcast) under which
+   * failing the event inline would deadlock. */
+  cl_int failed_dependency;
+
   /* implicit event = an event for pocl's internal use, not visible to user */
   char implicit_event;
 
--- a/lib/CL/devices/cuda/pocl-cuda.c
+++ b/lib/CL/devices/cuda/pocl-cuda.c
@@ -2249,6 +2249,15 @@ pocl_cuda_submit_node (_cl_command_node *node, cl_command_queue cq, int locked)
   cl_device_id dev = node->device;
   _cl_command_t *cmd = &node->command;
 
+  /* If a wait-list dependency has already failed, this command must not run:
+   * skip issuing any device work for it (its inputs may have been aborted), but
+   * still record the end event below so pocl_cuda_finalize_command can sync and
+   * clean up on valid CUDA handles. finalize_command honors failed_dependency
+   * and terminates the command via the error cascade. Unlike the other drivers
+   * CUDA does not call pocl_update_event_failed from submit: its finalize path
+   * already drives the failure, and bailing out here would leave the end event
+   * unrecorded (finalize would cuEventSynchronize on a garbage handle). */
+  if (!event->failed_dependency)
   switch (node->type)
     {
     case CL_COMMAND_READ_BUFFER:
@@ -2534,8 +2543,14 @@ pocl_cuda_finalize_command (cl_device_id device, cl_event event)
 
   /* Handle failed events */
 
+  /* Read the failed-dependency flag before pocl_update_event_running, which
+   * transitions the event to CL_RUNNING and so would mask a status that
+   * pocl_cuda_notify set to a negative value. A failed dependency means the
+   * command was not issued (see pocl_cuda_submit_node) and must be terminated. */
+  int failed_dependency = event->failed_dependency;
+
   pocl_update_event_running (event);
-  if (event->status < 0)
+  if (event->status < 0 || failed_dependency)
     POCL_UPDATE_EVENT_FAILED_MSG (CL_FAILED, event, "CUDA event failed");
   else
     POCL_UPDATE_EVENT_COMPLETE_MSG (event, "CUDA event");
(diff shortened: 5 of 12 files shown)


lib/CL/pocl_util.h
```diff
@@ -209,6 +209,41 @@ pocl_command_is_ready(cl_event event)
   return event->wait_list == NULL;
 }
 
+POCL_EXPORT void
+pocl_update_event_failed (cl_int status,
+                          const char *func,
+                          unsigned line,
+                          cl_event event,
+                          const char *msg);
+
```
jansol: This is redundant, the function is already declared in the same file.

lib/CL/devices/pthread/pthread.c
```diff
@@ -282,6 +288,19 @@ pocl_pthread_notify (cl_device_id device, cl_event event, cl_event finished)
 
   if (pocl_command_is_ready (node->sync.event.event))
     {
+      /* The command may have become ready while still carrying a broken
+       * (already-failed at wiring time) dependency recorded on another wait
+       * edge. Fail it instead of running it. pocl_broadcast calls us with both
+       * 'event' and 'finished' locked and expects them locked on return, so
+       * replicate the unlock/fail/relock dance used for the failed-finished
+       * case above. */
+      if (event->broken_dependency)
```
jansol: If we already know that a command MUST NOT execute, there is no point in waiting for it to become otherwise ready. I'd move this check up as part of the  `finished->status < CL_COMPLETE` clause and skip the readiness check.

lib/CL/devices/basic/basic.c
```diff
@@ -543,6 +543,15 @@ pocl_basic_submit (_cl_command_node *node, cl_command_queue cq)
     }
 
   node->state = POCL_COMMAND_READY;
+
+  /* A dependency that was already failed when its sync edge was wired up leaves
+   * the command ready with no wait_list. Fail it instead of running it on
+   * potentially freed/aborted memory. The helper unlocks and fails the event,
+   * so return immediately without unlocking again. */
+  if (pocl_command_is_ready (node->sync.event.event)
+      && pocl_command_honor_broken_dependency (node->sync.event.event))
```
jansol: If we already know that a command MUST NOT execute, there is no point in waiting for it to become otherwise ready before failing it.

lib/CL/pocl_cl.h
```diff
@@ -2137,6 +2137,17 @@ struct _cl_event {
...
+
+  /* Set to 1 if, at the time a sync edge to a dependency was being wired up,
+   * that dependency was already in a failed (negative) status. In that case no
+   * sync edge is created (the notifier will not broadcast again), so the
+   * failure cannot propagate through the normal pocl_broadcast -> notify path.
+   * The device submit path must check this flag and fail the command via the
+   * error cascade instead of running it. This deferral is required because the
+   * flag is set under the command-queue lock (in pocl_create_event_sync) where
+   * failing the event inline would deadlock. */
+  cl_int broken_dependency;
```
jansol: `failed_dependency` seems like a more descriptive name, the dependency was still a valid event even if it was marked as failed.

lib/CL/devices/pthread/pthread.c
```diff
@@ -218,6 +218,12 @@ pocl_pthread_submit (_cl_command_node *node, cl_command_queue cq)
   node->state = POCL_COMMAND_READY;
   if (pocl_command_is_ready (node->sync.event.event))
     {
+      /* A dependency that was already failed when its sync edge was wired up
+       * leaves the command ready with no wait_list. Fail it instead of running
+       * it on potentially freed/aborted memory. The helper unlocks and fails
+       * the event, so return immediately without unlocking again. */
+      if (pocl_command_honor_broken_dependency (node->sync.event.event))
```
jansol: No need to wait for the command to become ready here either.

lib/CL/devices/basic/basic.c
```diff
@@ -605,6 +614,19 @@ pocl_basic_notify (cl_device_id device, cl_event event, cl_event finished)
 
   if (pocl_command_is_ready (event))
     {
+      /* The command may have become ready while still carrying a broken
+       * (already-failed at wiring time) dependency recorded on another wait
+       * edge. Fail it instead of running it. pocl_broadcast calls us with both
+       * 'event' and 'finished' locked and expects them locked on return, so
+       * replicate the unlock/fail/relock dance used for the failed-finished
+       * case above. */
+      if (event->broken_dependency)
```
jansol: If we already know that a command MUST NOT execute, there is no point in waiting for it to become otherwise ready. I'd move this check up as part of the `finished->status < CL_COMPLETE` clause and skip the readiness check.

lib/CL/pocl_cl.h
```diff
@@ -2137,6 +2137,17 @@ struct _cl_event {
...
   cl_int status;
+
+  /* Set to 1 if, at the time a sync edge to a dependency was being wired up,
+   * that dependency was already in a failed (negative) status. In that case no
+   * sync edge is created (the notifier will not broadcast again), so the
+   * failure cannot propagate through the normal pocl_broadcast -> notify path.
+   * The device submit path must check this flag and fail the command via the
+   * error cascade instead of running it. This deferral is required because the
+   * flag is set under the command-queue lock (in pocl_create_event_sync) where
+   * failing the event inline would deadlock. */
```
jansol: This comment does not match how the variable is actually used.

lib/CL/devices/common.c
```diff
@@ -871,6 +871,20 @@ pocl_broadcast (cl_event brc_event)
...
+       * command must NOT run: per the OpenCL spec a command whose wait-list
+       * event has a negative status must be terminated with the error cascade.
+       * The device notify callback fails the target when it is in a failable
+       * state (QUEUED/SUBMITTED). However, the target may transiently be in
+       * another state (e.g. a concurrent broadcast for a *different*, completed
+       * dependency is mid-flight and about to push it), so record the broken
+       * dependency on the target as well. Every device submit / ready check
+       * honors this flag and fails the command instead of executing it on
+       * potentially freed/aborted memory. We are holding the target's event
+       * lock here, so this is safe against the submit/notify readers. */
```
jansol: This contradicts the comment in pocl_cl.h

jansol (changes requested):

lib/CL/pocl_util.h
bricevideau-ai: Good catch — fixed. The forward declaration was only there because the helper I had added below it called pocl_update_event_failed; since I removed that helper (see the related comments), the redundant declaration is gone too.

lib/CL/devices/pthread/pthread.c
bricevideau-ai: Done. The notify-path check is now folded into the existing `if (finished->status < CL_COMPLETE || event->failed_dependency)` failure clause, before the readiness check, and shares that branch's unlock/fail/relock dance. Note the `|| event->failed_dependency` is load-bearing: pocl_broadcast can set the flag for a notifier that itself finished *completed* (the concurrent-broadcast-of-a-different-dep window), so the flag has to be OR'd into the failure condition rather than nested under the negative-status branch.

lib/CL/devices/basic/basic.c
bricevideau-ai: Done — the basic submit path now fails on the flag unconditionally, with no readiness check first.

lib/CL/pocl_cl.h
bricevideau-ai: Agreed, renamed to `failed_dependency` throughout — it reads better and is accurate (the dependency was a valid event that failed).

lib/CL/devices/pthread/pthread.c
bricevideau-ai: Done. The submit path now checks the flag and fails the command unconditionally (no `pocl_command_is_ready` guard around it), and I removed the `pocl_command_honor_broken_dependency` helper — the unlock/fail/return is now inline in both submit paths, which is short enough that the helper was just indirection.

lib/CL/devices/basic/basic.c
bricevideau-ai: Done, same as the pthread notify path: folded into the `finished->status < CL_COMPLETE || event->failed_dependency` clause ahead of the readiness check.

lib/CL/pocl_cl.h
bricevideau-ai: Rewritten. The comment now describes both set sites (already-failed-at-wiring in pocl_create_event_sync, and the concurrent-broadcast window in pocl_broadcast) and both check sites (device submit + notify), and explains why a flag is used rather than an inline fail (both set sites hold a lock under which pocl_update_event_failed would re-lock and deadlock).

lib/CL/devices/common.c
bricevideau-ai: Fixed — reworded the pocl_broadcast comment to be consistent with the (now corrected) pocl_cl.h comment. Both now describe the same two-set-sites / two-check-sites model rather than the earlier mismatched wording.

bricevideau-ai:
Thanks for the thorough review @jansol — all of it was on point. Pushed a revision (force-update, single commit) that addresses every comment:

- **Renamed** \`broken_dependency\` → \`failed_dependency\`.
- **Removed** the redundant forward-declaration of \`pocl_update_event_failed\` and the \`pocl_command_honor_broken_dependency\` helper — the unlock/fail/return is now inline in both submit paths.
- **Submit paths** (pthread + basic): check the flag and fail unconditionally, no readiness check first.
- **Notify paths** (pthread + basic): folded the flag check into the existing \`finished->status < CL_COMPLETE\` failure clause as \`|| event->failed_dependency\`, ahead of the readiness check, sharing its unlock/fail/relock dance.
- **Comments** in \`pocl_cl.h\` and \`common.c\` rewritten to match actual usage (two set sites, two check sites) and to stop contradicting each other.

One thing worth flagging on the notify change: the \`|| event->failed_dependency\` term is deliberately OR'd into the failure condition rather than nested under \`finished->status < CL_COMPLETE\`. \`pocl_broadcast\` can set the flag on a target whose *current* notifier finished **completed** (the window where a concurrent broadcast for a different, already-failed dependency is mid-flight), so the flag must be honored even when \`finished\` itself is non-negative.

### Validation
Re-ran the concurrent reproducers against a fresh build of this branch:
- \`start_gate\` (isolates the already-failed-at-wiring case): 5/5 clean, downstream aborts with -14.
- \`race_finish\` (failed-dependency cascade): 6/6 clean.
- pocl runtime + event test suites: no regressions.

(The bundled \`failed_dep_gate\` reproducer additionally trips a *separate* pre-existing pocl race — a double-finish of a marker event that depends on both a command event and a user event, \`pocl_util.c:1690\`. That's orthogonal to this PR and I'll open it separately with its own guard; this PR is validated against the reproducers that isolate the failed-dependency path.)

jansol (approved):
Probably fine now

franz:
@bricevideau-ai  could you also update the notify & submit callbacks in other drivers (these: TBB, Vulkan, LevelZero, CUDA, Proxy) the same way as this PR changes pthread callbacks ? The other driver callbacks have quite similar code.

Kerilk:
Claude's working on it, I'll ask if it can add a test, this should be testable.

bricevideau-ai:
I added a regression test for this in a separate commit (`tests/runtime/test_event_failed_dependency.c`): it sets a user event to a failed status and then enqueues a command waiting on it (on an out-of-order queue), asserting the dependent command is terminated with a negative status rather than completing. I verified it **passes with the fix and fails when the fix is reverted** (the already-failed-at-enqueue case completes with status 0 without the guard). It also covers the fail-later/broadcast→notify path as a regression guard. I registered it for the `cuda`/`proxy`/`vulkan`/`level0` label sets so the other drivers exercise it in CI.

@franz — re: replicating the submit/notify guard in the TBB, Vulkan, LevelZero, CUDA, and Proxy drivers: happy to do that. One practical note — my local build only has the host-CPU (pthread/basic) driver enabled, so I cannot compile-test the other five locally; I would be relying on CI to build them. Two options:

1. I push the same submit-path flag check + `|| event->failed_dependency` notify term to all five in this PR, and we lean on CI to validate the builds (and the new test above to validate behavior on whichever of them CI runs).
2. We merge this PR (host-CPU fix + test, already approved) and I do the other drivers as a focused follow-up.

I lean slightly toward (1) since the changes are mechanical and the test is in place, but I am fine either way — let me know your preference. Note that CUDA is a bit different from the others: its `notify` sets `event->status = -1` directly rather than going through the `pocl_update_event_failed` cascade, and its `submit` allocates event data before the readiness gate, so that one needs a slightly adapted version rather than a verbatim copy.

bricevideau-ai:
@franz I went with option 1 and pushed the guard for all five drivers, so it is all in this PR:

- **TBB, Vulkan, Level0, Proxy** — verbatim mirror of the pthread/basic change: `failed_dependency` check in `submit` (fail via `pocl_update_event_failed` and return) and `|| event->failed_dependency` folded into the `notify` failure clause. For Proxy the submit check sits *after* the per-event data is allocated, so the failure cascade (which broadcasts on the event cond) sees valid event data; TBB has no per-event data, and Vulkan/Level0 allocate it at queue time via `update_event`, so for those the check is at the top of `submit` as in pthread/basic.

- **CUDA** — needed a CUDA-native version rather than a copy, because its failure path is different: `notify` sets `event->status = -1` directly instead of calling `pocl_update_event_failed`, and `finalize_command` runs `pocl_update_event_running` (→ `CL_RUNNING`) *before* the `status < 0` check, so a pre-set negative status would be masked. So instead: `pocl_cuda_submit_node` skips the command-issuing `switch` when `failed_dependency` is set (no device work on aborted inputs) but still records the end event, and `finalize_command` reads `failed_dependency` before `update_event_running` and terminates the command on it. The worker always appends the node to the running queue after submit, so finalize runs regardless and is the right place to honor the flag.

I do not build CUDA / TBB / Vulkan / LevelZero / Proxy locally, so I am relying on the CI runners (incl. the CUDA one) plus the new `tests/runtime/test_event_failed_dependency` to validate these. The test is labeled for the `cuda`/`proxy`/`vulkan`/`level0` sets so it runs against each where CI builds it. Will watch the CI result and fix up anything that surfaces.

PR #2216, opened 2026-06-25
Guard pocl_update_event_finished against concurrent double-finish

bricevideau-ai:
[pocl-double-finish-repro.zip](https://github.com/user-attachments/files/29316780/pocl-double-finish-repro.zip)
## Problem

An event can be driven to a terminal state **twice, concurrently, from two `pocl_broadcast` threads**, tripping:

```
pocl_util.c:1690: pocl_update_event_finished:
  Assertion `event->status > CL_COMPLETE' failed.
```

(With the assert compiled out, the second finisher instead mutates/re-lists/releases an event the first already terminated and may have recycled — an intermittent `SIGSEGV`/`SIGABRT` in `clReleaseEvent` on a nearby thread.)

This affects **any command with more than one wait-list dependency** — it is not specific to markers/barriers. Two interleavings produce it on a failing chain:

- **Two dependencies finish in error concurrently.** Each `pocl_broadcast` calls the device `notify()`, which takes the `finished->status < CL_COMPLETE` branch and fails the waiter via `pocl_update_event_finished` — twice.
- **One dependency fails the waiter while another dependency's completion drives it terminal too.** For a marker/barrier that is completion-as-termination; for a regular command the failed-dependency path can also race the submit/execute path.

## Why the transitions race

`pocl_broadcast` tests the waiter's "still pending" gate (`status == CL_SUBMITTED || CL_QUEUED`) **before** calling `notify`, and `notify` drops the event lock (a lock-order dance with the command queue) **before** the new status is written. So a second broadcast can read the stale pre-terminal status, pass the gate, and call into `pocl_update_event_finished` on an event the first finisher is already terminating.

## Root cause

`pocl_update_event_finished` assumes it is the sole finisher of an event. That holds for a single terminating transition, but not when two dependency broadcasts converge on the same multi-dependency waiter. The `assert (event->status > CL_COMPLETE)` is asserting the very invariant the function does not enforce on entry.

## Fix

OpenCL event execution state is **monotonic**: once terminal (`CL_COMPLETE` or a negative error) it never changes again, so a second finish is a no-op. Enforce that on entry — under the `cq` + `event` locks the function already takes, return if the event is already terminal:

```c
POCL_LOCK_OBJ (cq);
POCL_LOCK_OBJ (event);
if (event->status <= CL_COMPLETE)
  {
    POCL_UNLOCK_OBJ (event);
    POCL_UNLOCK_OBJ (cq);
    return;
  }
assert (event->status > CL_COMPLETE);
```

The check is under the event lock, so it is atomic against the other finisher: exactly one caller finishes the event, the other returns cleanly.

(The guard is intentionally unconditional, not scoped to a command type. Re-enabling the assert for "unaffected" types and re-testing showed there are none — a plain `clEnqueueReadBuffer` with two concurrently-failed dependencies double-finishes just like a marker does.)

## Testing

Two reproducers on a Debug build (assert active):
- a plain `clEnqueueReadBuffer` whose wait list holds two user events both set to a negative status (the pure two-errors case), and
- a marker depending on an `Unmap` completion plus a failed host-gate user event.

Both aborted at `pocl_util.c:1690` before the guard; **0 asserts after** (40/40 and 30/30 clean). pocl runtime + event test suites pass with no regressions.

### DIFF
--- a/lib/CL/pocl_util.c
+++ b/lib/CL/pocl_util.c
@@ -1665,6 +1665,37 @@ pocl_update_event_finished (cl_int status, const char *func, unsigned line,
   cl_command_queue cq = event->queue;
   POCL_LOCK_OBJ (cq);
   POCL_LOCK_OBJ (event);
+  /* Idempotency guard: an event can be driven to a terminal state twice,
+   * concurrently, from two different pocl_broadcast threads. This affects ANY
+   * command with more than one wait-list dependency, not just markers/barriers.
+   * Two interleavings produce it on a failing chain:
+   *
+   *   - Two dependencies finish in ERROR concurrently. Each pocl_broadcast
+   *     calls notify(), which takes the finished->status < CL_COMPLETE branch
+   *     and fails the waiter via pocl_update_event_finished -- twice.
+   *   - One dependency FAILS (failing the waiter) while another dependency's
+   *     COMPLETION drives the waiter terminal too (for a marker/barrier that is
+   *     completion-as-termination; for a regular command it can additionally
+   *     race the submit/execute path).
+   *
+   * These transitions are not mutually exclusive because pocl_broadcast tests
+   * the waiter's "still pending" gate (status == CL_QUEUED/SUBMITTED) BEFORE
+   * calling notify, and notify drops the event lock (a lock-order dance with
+   * the command queue) before the new status is written -- so a second
+   * broadcast can read the stale pre-terminal status and call in here on an
+   * event the first finisher is already terminating.
+   *
+   * OpenCL event state is monotonic: once terminal (CL_COMPLETE or a negative
+   * error) it never changes again, so a second finish is a no-op. The check is
+   * taken under the event lock, so exactly one caller finishes the event and
+   * the other returns cleanly. Bail under the locks we already hold
+   * (reverse-order unlock). */
+  if (event->status <= CL_COMPLETE)
+    {
+      POCL_UNLOCK_OBJ (event);
+      POCL_UNLOCK_OBJ (cq);
+      return;
+    }
   assert (event->status > CL_COMPLETE);
   if ((cq->properties & CL_QUEUE_PROFILING_ENABLE)
       && (cq->device->has_own_timer == 0))


bricevideau-ai:
## Reproducers

Two standalone reproducers (no pocl headers needed, just an OpenCL ICD). Build against a **Debug** pocl so the assert is active, then loop ~40×; the race is timing-dependent.

```
gcc -O2 -o pocl_double_error_repro    pocl_double_error_repro.c    -lOpenCL -lpthread
gcc -O2 -o pocl_failed_dep_gate_repro pocl_failed_dep_gate_repro.c -lOpenCL -lpthread
for i in $(seq 1 40); do ./pocl_double_error_repro; done
```

**`pocl_double_error_repro.c`** — pure two-errors case: a plain `clEnqueueReadBuffer` whose wait list holds two user events, both set to a negative status concurrently. No marker, no completion. Shows the bug is not marker-specific.

<details><summary>pocl_double_error_repro.c</summary>

```c
/* Reproducer: a plain command (clEnqueueReadBuffer) whose wait list contains
 * TWO user events that are both set to an ERROR status concurrently from two
 * threads. Neither dependency completes successfully and the waiter is not a
 * marker/barrier -- yet the two failed-dependency broadcasts both call
 * pocl_update_event_finished on the same waiter, double-finishing it.
 *
 * This isolates the general case: the double-finish is NOT specific to
 * marker/barrier and NOT dependent on a completion-vs-error race. Two errors
 * on any multi-dependency command suffice.
 *
 * Build:  gcc -O2 -o pocl_double_error_repro pocl_double_error_repro.c -lOpenCL -lpthread
 * Expect (buggy pocl):  intermittent  pocl_util.c: Assertion 'event->status > CL_COMPLETE' failed
 * Expect (fixed pocl):  "PASS: read terminated, no double-finish"
 */
#define CL_TARGET_OPENCL_VERSION 120
#include <CL/cl.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

#define N 1024

static cl_event eA, eB;       /* the two user-event dependencies            */
static cl_event start;        /* gates the read enqueue (commit before fire) */

/* Two worker threads, one per user event, both firing an ERROR at the same
 * time so their broadcasts into the shared waiter race. */
static void *fire_A (void *p) { (void)p; clWaitForEvents (1, &start);
                                clSetUserEventStatus (eA, -1); return NULL; }
static void *fire_B (void *p) { (void)p; clWaitForEvents (1, &start);
                                clSetUserEventStatus (eB, -1); return NULL; }
static void *watchdog (void *p){ (void)p; sleep (10);
                                 fprintf (stderr, "HANG\n"); _exit (42); }

int main (void)
{
  cl_platform_id plat; cl_device_id dev; cl_int err;
  clGetPlatformIDs (1, &plat, NULL);
  clGetDeviceIDs (plat, CL_DEVICE_TYPE_DEFAULT, 1, &dev, NULL);
  cl_context ctx = clCreateContext (NULL, 1, &dev, NULL, NULL, &err);
  cl_command_queue Q =
    clCreateCommandQueue (ctx, dev, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
  cl_mem buf = clCreateBuffer (ctx, CL_MEM_READ_WRITE, N * 4, NULL, &err);

  start = clCreateUserEvent (ctx, &err);
  eA    = clCreateUserEvent (ctx, &err);
  eB    = clCreateUserEvent (ctx, &err);

  /* A no-op command gated on `start` so the graph is committed before we fire,
   * mirroring the start-gate timing that widens the race window. */
  cl_event gate_ev;
  clEnqueueMarkerWithWaitList (Q, 1, &start, &gate_ev);

  /* The waiter: a PLAIN read-buffer depending on BOTH user events. */
  int *host = malloc (N * 4);
  cl_event wl[2] = { eA, eB };
  cl_event read_ev;
  cl_int enq = clEnqueueReadBuffer (Q, buf, CL_FALSE, 0, N * 4, host, 2, wl, &read_ev);
  fprintf (stderr, "[main] read enqueue = %d; releasing start, both deps -> error\n", enq);

  pthread_t tA, tB, tw;
  pthread_create (&tw, NULL, watchdog, NULL);
  pthread_create (&tA, NULL, fire_A, NULL);
  pthread_create (&tB, NULL, fire_B, NULL);

  clSetUserEventStatus (start, CL_COMPLETE);

  cl_int w = clWaitForEvents (1, &read_ev);
  fprintf (stderr, "[main] clWaitForEvents(read_ev) = %d\n", w);

  pthread_join (tA, NULL);
  pthread_join (tB, NULL);

  cl_int st = 0;
  clGetEventInfo (read_ev, CL_EVENT_COMMAND_EXECUTION_STATUS, sizeof (st), &st, NULL);
  fprintf (stderr, "[main] read exec status = %d\n", st);

  if (st < 0)
    { printf ("PASS: read terminated, no double-finish\n"); return 0; }
  printf ("UNEXPECTED: read status %d\n", st);
  return 1;
}
```

</details>

**`pocl_failed_dep_gate_repro.c`** — marker/aggregator case: a marker depends on an `Unmap` completion plus a host-gate user event; on the abort path both broadcasts converge on the marker.

<details><summary>pocl_failed_dep_gate_repro.c</summary>

```c
// Reproducer: a command whose wait-list event is completed with a NEGATIVE
// (failed) status is executed anyway instead of being terminated, leading to a
// crash (SIGSEGV in a pocl worker thread: pocl_exec_command -> memcpy) on
// memory the terminated dependency already released.
//
// Spec (OpenCL 3.0 §5.10, clSetUserEventStatus / event execution status):
// completing an event with a negative value terminates it; commands in the same
// or other queues that wait on it (and the event itself) are terminated, with
// the error propagated. Such a waiting command must NOT execute.
//
// ---------------------------------------------------------------------------
// Why this exact shape (and what it deliberately avoids)
// ---------------------------------------------------------------------------
// This mirrors a real host-callback pattern: a host thread runs a callback on a
// mapped buffer and, on failure, must abort the rest of an already-enqueued
// graph. The pieces:
//
//   start   : user event gating the FIRST enqueue. Released only AFTER the whole
//             graph is enqueued, so the graph is fully committed before anything
//             runs — this makes the bug DETERMINISTIC rather than a timing race.
//   fire    : user event gating the unmap; always completed CL_COMPLETE (one
//             clean unmap — no second/defensive unmap).
//   proceed : user event gating the final read; completed NEGATIVE on the error
//             path to abort the downstream read (CL_COMPLETE on the --success
//             control path).
//
//   Graph (one out-of-order queue):
//     fill  (gated on [start])
//     map   (gated on [fill])              -> mapev (non-blocking map)
//     unmap (gated on [fire])              -> un
//     read  (gated on [un, proceed])       -> read_ev   (LAST command)
//   worker thread: wait mapev; fire=CL_COMPLETE; proceed = (-1 | CL_COMPLETE)
//   main thread:   release start=CL_COMPLETE; clWaitForEvents(read_ev)
//
// It uses clWaitForEvents on the last event (NOT clFinish) on purpose, to avoid
// conflating this bug with a separate clFinish-on-a-terminated-command issue.
// There is one CLI control only — `--success` — which completes `proceed`
// normally; that path must always work and is the baseline.
//
// ---------------------------------------------------------------------------
// Expected vs observed
// ---------------------------------------------------------------------------
//   Correct:  the read is terminated (its execution status is negative,
//             CL_EXEC_STATUS_ERROR_FOR_EVENTS_IN_WAIT_LIST); it does NOT run.
//             Program prints PASS and exits 0.
//   Bug:      the read executes on the buffer the terminated chain released ->
//             SIGSEGV in a pocl worker thread; or it "succeeds" (status
//             CL_COMPLETE) having run on terminated data -> prints FAIL.
//
// A watchdog thread aborts after a few seconds (exit 42) so the program can
// never hang the harness; this bug manifests as a crash, not a hang, on pocl.
//
// Build: cc pocl_failed_dep_gate_repro.c -lOpenCL -lpthread -o repro
// Run:   OCL_ICD_VENDORS=<pocl OpenCL/vendors dir> ./repro
//        OCL_ICD_VENDORS=<...>                      ./repro --success   (control)
// Exit:  0 = correct (read terminated / control completed); 1 = FAIL (read ran
//        on terminated data); 42 = watchdog (hang); 2 = setup error; SIGSEGV = bug.
#define CL_TARGET_OPENCL_VERSION 120
#include <CL/cl.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#define N 1024
#define WATCHDOG_SECS 8

static cl_event fire, proceed, mapev;
static cl_command_queue Q;
static int err_path = 1; /* default: abort path (proceed = -1) */

#define CHECK(expr)                                                            \
  do {                                                                         \
    cl_int _e = (expr);                                                        \
    if (_e != CL_SUCCESS) {                                                    \
      fprintf(stderr, "setup error %d at line %d (%s)\n", _e, __LINE__, #expr);\
      exit(2);                                                                 \
    }                                                                          \
  } while (0)

/* Host worker: waits for the map to complete, fires the unmap, then either
   aborts (proceed = -1) or proceeds (proceed = CL_COMPLETE). */
static void *worker(void *arg) {
  (void)arg;
  clWaitForEvents(1, &mapev); /* map completes only after `start` is released */
  clSetUserEventStatus(fire, CL_COMPLETE);
  clSetUserEventStatus(proceed, err_path ? -1 : CL_COMPLETE);
  return NULL;
}

static void *watchdog(void *arg) {
  (void)arg;
  sleep(WATCHDOG_SECS);
  fprintf(stderr, "[watchdog] %ds elapsed with no return -> HANG\n",
          WATCHDOG_SECS);
  _exit(42);
}

int main(int argc, char **argv) {
  for (int i = 1; i < argc; i++)
    if (!strcmp(argv[i], "--success"))
      err_path = 0;

  cl_platform_id plat;
  cl_device_id dev;
  CHECK(clGetPlatformIDs(1, &plat, NULL));
  CHECK(clGetDeviceIDs(plat, CL_DEVICE_TYPE_DEFAULT, 1, &dev, NULL));
  char name[256] = {0}, drv[256] = {0};
  clGetDeviceInfo(dev, CL_DEVICE_NAME, sizeof(name), name, NULL);
  clGetDeviceInfo(dev, CL_DRIVER_VERSION, sizeof(drv), drv, NULL);
  fprintf(stderr, "Device: %s | driver %s | path: %s\n", name, drv,
          err_path ? "abort (proceed=-1)" : "success (proceed=CL_COMPLETE)");

  cl_int err;
  cl_context ctx = clCreateContext(NULL, 1, &dev, NULL, NULL, &err);
  CHECK(err);
  /* Out-of-order queue: commands wait on their event dependencies rather than
     submission order, so the gating is meaningful. */
  Q = clCreateCommandQueue(ctx, dev, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
                           &err);
  CHECK(err);
  cl_mem buf = clCreateBuffer(ctx, CL_MEM_READ_WRITE, N * sizeof(int), NULL,
                              &err);
  CHECK(err);

  /* `start` gates the first enqueue so the whole graph is committed before any
     of it runs. */
  cl_event start = clCreateUserEvent(ctx, &err);
  CHECK(err);

  int seven = 7;
  cl_event fillev;
  CHECK(clEnqueueFillBuffer(Q, buf, &seven, sizeof(seven), 0, N * sizeof(int), 1,
                            &start, &fillev));
  void *ptr = clEnqueueMapBuffer(Q, buf, CL_FALSE, CL_MAP_READ | CL_MAP_WRITE, 0,
                                 N * sizeof(int), 1, &fillev, &mapev, &err);
  CHECK(err);

  fire = clCreateUserEvent(ctx, &err);
  CHECK(err);
  proceed = clCreateUserEvent(ctx, &err);
  CHECK(err);
  cl_event un;
  CHECK(clEnqueueUnmapMemObject(Q, buf, ptr, 1, &fire, &un));

  pthread_t wt, wd;
  pthread_create(&wt, NULL, worker, NULL);
  pthread_create(&wd, NULL, watchdog, NULL);

  /* LAST command: a read gated on [unmap, proceed]. On the abort path `proceed`
     becomes negative, so this read must be terminated, not executed. */
  cl_event wait_list[2] = {un, proceed};
  int *host = malloc(N * sizeof(int));
  cl_event read_ev;
  cl_int enq = clEnqueueReadBuffer(Q, buf, CL_FALSE, 0, N * sizeof(int), host, 2,
                                   wait_list, &read_ev);
  fprintf(stderr, "[main] whole graph enqueued (read enqueue = %d). "
                  "Releasing start.\n",
          enq);

  /* Whole graph is enqueued -> release it. */
  clSetUserEventStatus(start, CL_COMPLETE);

  /* Wait on the LAST event only (deliberately not clFinish). */
  fprintf(stderr, "[main] clWaitForEvents(read_ev)...\n");
  cl_int we = clWaitForEvents(1, &read_ev);
  fprintf(stderr, "[main] clWaitForEvents returned %d\n", we);

  /* Join the worker before teardown so its CL calls finish before the context
     is destroyed. */
  pthread_join(wt, NULL);

  cl_int read_status = 0;
  clGetEventInfo(read_ev, CL_EVENT_COMMAND_EXECUTION_STATUS,
                 sizeof(read_status), &read_status, NULL);
  fprintf(stderr, "[main] read command execution status = %d\n", read_status);

  int rc;
  if (!err_path) {
    /* control: success path must complete normally */
    rc = (read_status == CL_COMPLETE) ? 0 : 1;
    fprintf(stderr, rc == 0 ? "PASS (control): read completed.\n"
                            : "FAIL (control): read did not complete.\n");
  } else if (read_status < 0) {
    fprintf(stderr, "PASS: read terminated (not executed) — correct.\n");
    rc = 0;
  } else {
    fprintf(stderr, "FAIL: read ran despite a failed dependency "
                    "(status=%d) — executed on terminated data.\n",
            read_status);
    rc = 1;
  }
  printf("%s\n", rc == 0 ? "PASS" : "FAIL");
  return rc;
}
```

</details>

franz (approved):

PR #2218, opened 2026-06-26
Add fp16 support for ldexp and add kernel tests

woodychow:
This fixes a failure in the OpenCL Conformance Test Suite (CTS) 'bruteforce/ldexp' test. When compiling for fp16-enabled devices, the test would fail to load due to missing 'half' overloads of '_cl_ldexp' (e.g., unresolved symbol '_Z9_cl_ldexpDv4_DhDv4_i').

This patch:
1. Adds a fallback macro for '__builtin_ldexpf16' in 'lib/kernel/ldexp.cl' that casts inputs to 'float', calls '__builtin_ldexpf', and casts the result back to 'half'.
2. Defines fp16 test harness helper macros ('DEFINE_BODY_V_FP16' and 'CALL_FUNC_V_FP16') in 'tests/kernel/common.cl'.
3. Adds test cases for 'half' precision scalar and vector variants of 'ldexp' to 'tests/kernel/test_ldexp.cl'.

### DIFF
--- a/lib/kernel/ldexp.cl
+++ b/lib/kernel/ldexp.cl
@@ -24,9 +24,14 @@
 
 #include "templates.h"
 
-// there is no Clang builtin ldexp for fp16 type
+#ifdef cl_khr_fp16
+#ifndef __builtin_ldexpf16
+#define __builtin_ldexpf16(a, b) ((half)__builtin_ldexpf((float)(a), (b)))
+#endif
+#else
 #undef __IF_FP16
 #define __IF_FP16(X)
+#endif
 
 DEFINE_BUILTIN_V_VJ(ldexp)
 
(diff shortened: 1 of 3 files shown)


woodychow:
I am not sure about the test changes (especially in tests/kernel/common.cl). Let me know what you think.

woodychow:
Looking at the issue carefully, most complex math functions (trigo, log, exp) are disabled with

```
#undef __IF_FP16
#define __IF_FP16(X)
```

Would it be a good idea to apply the same pattern to all these functions?

franz:
Hi, thanks for the effort! The fallback to float `ldexp` itself is OK, but the problem is the placement of the code. `lib/kernel/ldexp.cl` is one of the two possible options for ldexp, the other is `lib/kernel/sleef-pocl/ldexp.cl`.

Figuring out where to place FP16 code is a bit complicated, because PoCL's math builtins can be compiled with LLVM vectorized builtins or without (cmake option ENABLE_HOST_CPU_VECTORIZE_BUILTINS). If that option is enabled, some of the vectorized builtins will provide their variants of FP16 functions.

Anyway, @maleadt recently provided a PR #2224 with almost all of the missing builtins. With your PR on half-conversions and a few fixes of my own, I think PoCL could pass full CTS with FP16 enabled.

woodychow:
All good. Thanks for letting me know!

Yeah, I needed full fp16 support and #2224 looks good.

PR #2220, opened 2026-06-28
cpu: disable AArch64 outline-atomics for in-process-linked kernels

maleadt:
Outline-atomics lowers integer atomicrmw to static-only libgcc __aarch64_* helpers; in-process-linked kernels leave them unresolved at dlopen time and the atomics silently no-op. Disable it in codegen (so both the lld and ORC backends benefit) to emit atomics inline.

Fixes a regression introduced by #2194


franz (approved):

maleadt:
Probably needs a backport to https://github.com/pocl/pocl/tree/release_7_2

franz:
yep backported

PR #2224, opened 2026-06-30
core-math: provide missing FP16 host math overloads

maleadt:
When `cl_khr_fp16` is enabled, the host CPU kernel library advertises half support but several `_cl_*` half math entry points are absent in vectorized builtin configurations, because the SLEEF/libmvec wrappers only cover float and double. SPIR wrappers and OpenCL C kernels can then fail at program creation or link time with unresolved half math symbols.

Add the missing FP16 overloads to the host core-math library. Operations that have native LLVM half support stay in `additionalf16.cl`; the remaining builtins live in `promotedf16.cl` and are implemented by promoting through float, handling vector forms and pointer address spaces explicitly. Include the new source in both vectorized and non-vectorized host builds, and add the needed FP16 builtin declaration helpers.

Without this, the added tests fails on a build like:

```
CMAKE_GENERATOR=Ninja
CMAKE_BUILD_TYPE=Debug
ENABLE_TESTS=ON
ENABLE_SLEEF=ON
ENABLE_HOST_CPU_VECTORIZE_SVML=ON
ENABLE_HOST_CPU_VECTORIZE_LIBMVEC=OFF
ENABLE_HOST_CPU_VECTORIZE_SLEEF=OFF
ENABLE_HOST_CPU_VECTORIZE_BUILTINS=ON
KERNELLIB_HOST_CPU_VARIANTS=distro
LLVM_VERSION_FULL=20.1.8jl
LLVM_LIBS=-lLLVM-20jl
CPU_USE_LLD_LINK=defined
HOST_CPU_ENABLE_SPIRV=defined
OCL_KERNEL_TARGET_CPU="znver5"
HOST_LD_FLAGS="-shared -nostartfiles /opt/intel/oneapi/compiler/latest/lib/libsvml.a /opt/intel/oneapi/compiler/latest/lib/libirc.a"
```

... with:

```
Running test test_fp16_math_builtins...

Unknown OpenCL error 1 in clCreateProgram call failed
 on line 42
FAIL

Error(s) while linking:
Cannot find symbol _Z9_cl_rootnDv8_DhDv8_i in kernel library
Cannot find symbol _Z9_cl_rootnDhi in kernel library
Cannot find symbol _Z9_cl_ldexpDhi in kernel library
Cannot find symbol _Z10_cl_remquoDhDhPU9CLgenerici in kernel library
Cannot find symbol _Z8_cl_powrDv8_DhS_ in kernel library
[...]

PR #2226, opened 2026-07-02
cpu: enable cl_khr_fp16 on MinGW hosts

maleadt:
FP16 host math support was gated to CMAKE_HOST_SYSTEM_NAME == "Linux" (commit 9de7551) to dodge a _Float16 build failure on FreeBSD.

Native MinGW satisfies every real requirement: clang codegens _Float16 for the host CPU target, libgcc provides the __extendhfsf2/__truncsfhf2 conversion helpers (so the FreeBSD failure mode does not occur), and the host compiler supports _Float16. Add MINGW to the host allowlist, keeping FreeBSD and other untested hosts gated off. Validated on native MinGW/MSYS2 (LLVM 22): full build plus the tests/kernel/test_fp16_* / test_halfs / test_shuffle_half / test_printf_vectors_halfn suite pass.

Also fix a latent crash this exposes: __pocl_print_floats_<T> classified values with isnan/isinf/signbit on the raw element type, and the _Float16 path traps at runtime under MinGW libgcc while printing half inf/nan. Classify through double instead (half->double and float->double are exact, double->double is a no-op), preserving nan/inf/sign for every type. Fixes test_printf_vectors_halfn.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1645,10 +1645,14 @@ endif()
 # LLVM <15 doesn't support _Float16;
 # LLVM 15 crashes on some code; LLVM 16 seems to work but fails with ldexp; LLVM 18 fails with isinf
 # CONFORMANCE=ON disables FP16 b/c PoCL support is incomplete
+# The host check is on the *build* machine, not the target -- so Yggdrasil's
+# Linux cross-builds already ship working FP16 in the macOS/Windows binaries.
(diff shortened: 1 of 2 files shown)


maleadt:
Passes now on the Julia side after a rebase included https://github.com/pocl/pocl/pull/2224

franz (approved):
LGTM

PR #2228, opened 2026-07-03
Don't crash on an unparseable program.bc; report the error

maleadt:
parseModuleIR/parseModuleIRMem can return null for a corrupt or version-incompatible program.bc (e.g. a stale kernel-cache entry left by an older driver build). The callers only had asserts, compiled out in a release driver, so this dereferenced null in parseModuleGVarSize and crashed the whole process.


franz (approved):
LGTM

PR #2229, opened 2026-07-03
Windows: close the fd, not the native handle, in the LLVM file utils

maleadt:
pocl_read_file2/pocl_write_file2 obtained a descriptor from openFileForRead / openFileForWrite / createUniqueFile but released it via closeFile(convertFDToNativeFile(fd)). On Windows that fd wraps a HANDLE created through _open_osfhandle(): closing the HANDLE leaves the CRT fd-table slot allocated. pocl.dll shares that table with the host process, so every kernel compilation leaked a couple of slots and the host eventually hit EMFILE. Close the fd directly with Process::SafelyCloseFileDescriptor() instead.

Because of this, we had to periodically recycle test workers in OpenCL.jl.

### DIFF
--- a/lib/CL/pocl_llvm_file_utils.cc
+++ b/lib/CL/pocl_llvm_file_utils.cc
@@ -68,6 +68,7 @@ IGNORE_COMPILER_WARNING("-Wunused-parameter")
 #include <llvm/Bitcode/BitcodeWriter.h>
 
 #include <llvm/Support/Errc.h>
+#include <llvm/Support/Process.h>
 
 POP_COMPILER_DIAGS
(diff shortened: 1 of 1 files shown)


franz (approved):
LGTM

pjaaskel:
Perhaps worth cherry picking to the 7.2 branch?

maleadt:
I see it on `release_next` already, so I presume it'll end up in 7.2?

franz:
Yes, I use `release_next` to trigger the CIs on commits I intend to push into the release branch. It's now pushed into 7.2

Issue #2113, opened 2026-02-02
get_local_id(0) % N returns 0 for all work-items when kernel contains abort/unreachable code path


Noerr:
## Summary

When a kernel contains an abort-like code path (`while(1) {}` or similar unreachable code) inside a conditional `if ((get_local_id(0) % N) == 0)` block, **all other** `get_local_id(0) % N` operations in the kernel incorrectly return 0 for every work-item.

This causes incorrect results: all work-items compute the same value instead of distributing work across the workgroup.

## Environment

- **POCL version**: 7.1 (LLVM 19) and 7.2-pre/main @ 2b01e4ec1 (LLVM 21) - both affected
- **Platform**: macOS 15.x (Sequoia), Apple Silicon (arm64)
- **Device**: cpu (ARM)
- **Observed via**: ChipStar (HIP → SPIR-V → POCL), but SPIR-V analysis confirms correct input

## Minimal Pattern

```c
void helper_with_abort(__local int* data) {
    // ... computation ...
    if (error_condition) {
        while(1) {}  // or any unreachable/abort pattern
    }
}

__kernel void buggy_kernel(__global float* output) {
    __local int shared_data[2];

    // First modulo - used as condition for calling abort-containing function
    if ((get_local_id(0) % 36) == 0) {
        helper_with_abort(shared_data);
    }
    barrier(CLK_LOCAL_MEM_FENCE);

    // Second modulo - BUG: returns 0 for ALL work-items
    int offset = get_local_id(0) % 36;  // Should be 0,1,2,...,35
                                         // Actually returns 0 for all

    output[get_group_id(0) * 36 + offset] = compute(offset);
    // Result: all 36 work-items write to the same location (offset=0)
}
```

## Observed Behavior

With workgroup size = 36:

| Work-item | Expected `get_local_id(0) % 36` | Actual |
|-----------|--------------------------------|--------|
| 0 | 0 | 0 |
| 1 | 1 | 0 |
| 2 | 2 | 0 |
| ... | ... | 0 |
| 35 | 35 | 0 |

**Result**: Only node 0 gets computed; nodes 1-35 are never written (remain NaN).

## SPIR-V Analysis

The SPIR-V input to POCL was examined and **both modulo operations are correctly generated**:

```spirv
; First modulo (in conditional block)
%rem_i = OpUMod %uint %conv_i_i %uint_36
%cmp_i = OpIEqual %bool %rem_i %uint_0
[...]

PR #2235, opened 2026-07-14
llvmopencl: guard implicit barriers by source dominance

maleadt:
ImplicitConditionalBarriers can choose an insertion block nested below a divergent guard. If another control-flow path bypasses that block and still reaches the source barrier, work-item-loop lowering can execute the guarded path for non-participating work-items.

Require the insertion block to dominate the source barrier. This lets the source barrier's all-or-none execution guarantee apply to the implicit barrier, while retaining valid dynamically uniform cases such as lane-indexed bounds checks before flattened subgroup shuffles.

Add two OpenCL C regressions: one where a bypass makes insertion unsafe, and one where every work-item dynamically passes a lane-dependent bounds check before `sub_group_shuffle`.

Related: #1747, #1794, #1971, #2113.

Implemented with the help of GPT 5.6

### DIFF
--- a/tests/regression/test_divergent_implicit_barrier_entry.c
+++ b/tests/regression/test_divergent_implicit_barrier_entry.c
@@ -1,0 +1,148 @@
+// Copyright (c) 2026 PoCL developers
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


pjaaskel (changes requested):
Thanks for tackling these annoying corner cases. Anyhow, what happens here exactly is not entirely clear to me. Can you add more comments about where it added the implicit barrier before and where it does not (or does) after your fix? Perhaps add it inside your test code as a commented out /* implicit_barrier() */ or such. Or in the commit log.

tests/regression/test_divergent_implicit_barrier_entry.c
```diff
@@ -0,0 +1,148 @@
...
+  err = clBuildProgram(program, 1, &device, NULL, NULL, NULL);
+  if (err != CL_SUCCESS) {
+    size_t log_size = 0;
+    clGetProgramBuildInfo(program, device, CL_PROGRAM_BUILD_LOG, 0, NULL,
+                          &log_size);
+    char *log = malloc(log_size);
+    if (log != NULL) {
+      clGetProgramBuildInfo(program, device, CL_PROGRAM_BUILD_LOG, log_size,
+                            log, NULL);
+      fprintf(stderr, "Build error:\\n%s\\n", log);
```
pjaaskel:
```suggestion
      fprintf(stderr, "Build error:\n%s\n", log);
```

maleadt:
> Perhaps add it inside your test code as a commented out /* implicit_barrier() */ or such.

Added something like it.

pjaaskel:
Got it. You use the latter barrier to propagate the all-or-none semantics to the split point. The barrier implicitly communicates the fact that although there's an id-dependent exit before it, it won't be taken dynamically in a OpenCL-legal kernel launch.

maleadt:
> You use the latter barrier to propagate the all-or-none semantics to the split point.

Yes. The source barrier’s all-or-none semantics can be propagated to the continuing path after the split only when the insertion block dominates the source barrier. Otherwise another path may bypass the insertion point and still reach the source barrier, i.e., making it not all-or-none.

maleadt:
My original check simply did `VUA.isUniform(&F, Pos)`, but that caused miscompilations in `WorkitemLoops`. I guess it relies on the canonical IR, with certain implicit barriers inserted? In any case, the dominator-based approach is better, since it is less conservative.

pjaaskel:
I cannot recall exactly, but VUA probably just thinks the BB is divergent due to an id-derived branch out as it doesn't utilize the same rationale of using barriers after the block to be able to assume uniformity. Maybe a better fix indeed would be to make VUA understand this case. But I can merge this anyhow after CIs psas.

maleadt:
Chipstar CI failure looks unrelated?

pjaaskel (approved):

franz:
@maleadt  yeah unrelated.

PR #2239, opened 2026-07-17
WorkitemLoops: share local memory across region replicas

maleadt:
AI-generated fix for what I observed in https://github.com/pocl/pocl/pull/2235#issuecomment-5006799743: simply avoiding certain implicit barriers (by checking `isUniform`) resulted in a miscompilation in `WorkitemLoops`.

---

Work-item loop unrolling and conditional barrier handling replicate parallel regions. If a region contains a compiler-expanded local-memory allocation, lowering the original and cloned calls independently gives each replica different storage. Subgroup operations then read incomplete data after their barrier.

Make peeled and unrolled replicas reuse the allocation call from the original region before local-memory calls are lowered. Add a regression that forces region unrolling around a subgroup shuffle and fails with both loops handlers before the fix.

### DIFF
--- a/lib/llvmopencl/WorkitemLoops.cc
+++ b/lib/llvmopencl/WorkitemLoops.cc
@@ -133,6 +133,8 @@ private:
   void fixMultiRegionVariables(ParallelRegion *Region);
   void addContextSaveRestore(llvm::Instruction *instruction);
   void releaseParallelRegions();
+  void shareReplicatedLocalMemAllocas(
+      ParallelRegion &Region, llvm::ValueToValueMapTy &ReferenceMap);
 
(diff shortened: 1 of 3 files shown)


pjaaskel:
Yet another bug in the peeling code. Peeling was removed in the next compiler version.  Hopefully we'll get it to a publishable state soon.

pjaaskel (approved):

maleadt:
OpenCL.jl CI failure is because we already have a test for the back-ported version of https://github.com/pocl/pocl/pull/2235, so merging that should resolve it here (and on `main`) too.

Chipstar CI failure unrelated.

PR #2245, opened 2026-07-22
Avoid reloading the executable in PoCL JIT when statically linked.

woodychow:
When PoCL is statically linked into a PIE (Position Independent Executable) binary on Linux, `pocl_dynlib_pathname` resolves to the path of the main executable.

During JIT initialization, PoCL attempts to `dlopen()` this path to resolve host symbols.

- On older glibc versions (e.g. < 2.29), glibc allows `dlopen()` on the running PIE executable, which causes the dynamic linker to reload it and rerun all global constructors. In complex C++ applications, this duplicate initialization of global state can lead to crashes (e.g. in libraries with non-idempotent initializers like protobuf or custom thread pools).
- On newer glibc versions (>= 2.29), `dlopen()` on a PIE executable is explicitly blocked by the dynamic linker (see glibc BZ #24323 [1] and commit 6f90352528 [2]), which avoids the crash but causes `addLibrarySearchGenerator` to fail and print warnings.

This patch checks if the path to be loaded matches the current executable (via `/proc/self/exe` on Linux) and skips calling `addLibrarySearchGenerator` if it does. The symbols are already available in the process-global scope due to static linking, so skipping the `dlopen` is safe and avoids both the crash on older glibc and the loading failure/warnings on newer glibc.

[1] https://sourceware.org/bugzilla/show_bug.cgi?id=24323
[2] https://sourceware.org/git/?p=glibc.git;a=commit;h=6f903525283995874c728362624564c78275688a

### DIFF
--- a/lib/CL/pocl_llvm_orc.cc
+++ b/lib/CL/pocl_llvm_orc.cc
@@ -556,7 +556,7 @@ int pocl_jit_initialize(const char *TripleStr, const char *CPU) {
         HOST_CPU_COMPILER_RT_LIBRARY);
 #endif
 
-#ifdef HAVE_DLFCN_H
+#if defined(HAVE_DLFCN_H) && defined(POCL_BUILD_SHARED_LIB)
   /* The JIT's default process-symbol search uses the global dlsym scope, which
      misses libpocl's private dependencies (libgcc_s for builtins like
      __aeabi_uldivmod or __extendhfsf2, libm before glibc 2.34) when an ICD
--- a/lib/CL/CMakeLists.txt
+++ b/lib/CL/CMakeLists.txt
@@ -25,6 +25,10 @@
 
 include_directories(AFTER "devices" ".")
 
+if(BUILD_SHARED_LIBS)
+  add_compile_definitions(POCL_BUILD_SHARED_LIB)
+endif()
+
 add_subdirectory("devices")
 
 ###############################################################################


lib/CL/pocl_llvm_orc.cc
```diff
@@ -568,12 +587,16 @@ int pocl_jit_initialize(const char *TripleStr, const char *CPU) {
...
+  const bool IsStatic = SelfPath && isCurrentExecutable(SelfPath);
+  // Skip silently if PoCL is statically linked into the executable.
+  if (!IsStatic) {
+    if (!addLibrarySearchGenerator(SelfPath))
+      POCL_MSG_PRINT_LLVM(
+          "pocl_jit: could not add libpocl's own libraries ('%s') to the symbol "
+          "search; kernels needing libpocl's private dependencies (libgcc_s, "
+          "libm) may fail to resolve symbols if those are not process-global\n",
+          SelfPath ? SelfPath : "<unknown>");
+  }
```
linehill: We do know when PoCL is going to be a static library so would it be better if:

```suggestion
#ifdef POCL_BUILD_SHARED_LIB
  if (!addLibrarySearchGenerator(SelfPath))
    POCL_MSG_PRINT_LLVM(
        "pocl_jit: could not add libpocl's own libraries ('%s') to the symbol "
        "search; kernels needing libpocl's private dependencies (libgcc_s, "
        "libm) may fail to resolve symbols if those are not process-global\n",
        SelfPath ? SelfPath : "<unknown>");
}
#endif
```
?

POCL_BUILD_SHARED_LIB would reflect CMake's [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) configuration.

woodychow: Indeed. This is much cleaner. Thanks!

linehill (approved):
LGTM, thanks.

PR #2248, opened 2026-07-25
Prefer RequirePassInfoMixin

boomanaiden154:
LLVM 23 made this preferred over normal PassInfoMixin and setting isRequired to return true. LLVM 24 will move PassInfoMixin to a detail namespace, causing a compilation error without this patch.

### DIFF
--- a/lib/llvmopencl/AllocasToEntry.h
+++ b/lib/llvmopencl/AllocasToEntry.h
@@ -33,11 +33,19 @@
 
 namespace pocl {
 
-class AllocasToEntry : public llvm::RequiredPassInfoMixin<AllocasToEntry> {
+class AllocasToEntry
+#if LLVM_MAJOR >= 23
+    : public llvm::RequiredPassInfoMixin<AllocasToEntry> {
+#else
+    : public llvm::PassInfoMixin<AllocasToEntry> {
+#endif
 public:
   static void registerWithPB(llvm::PassBuilder &B);
   llvm::PreservedAnalyses run(llvm::Function &F,
                               llvm::FunctionAnalysisManager &AM);
+#if LLVM_MAJOR < 23
+  static bool isRequired() { return true; }
+#endif
 };
 
   } // namespace pocl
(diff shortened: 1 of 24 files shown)


lib/llvmopencl/AllocasToEntry.h
```diff
@@ -33,15 +33,13 @@
 
 namespace pocl {
 
-class AllocasToEntry : public llvm::PassInfoMixin<AllocasToEntry> {
+class AllocasToEntry : public llvm::RequiredPassInfoMixin<AllocasToEntry> {
```
linehill: Should:
```suggestion
#if LLVM_MAJOR < 23
class AllocasToEntry : public llvm::PassInfoMixin<AllocasToEntry> {
#else
class AllocasToEntry : public llvm::RequiredPassInfoMixin<AllocasToEntry> {
#endif
```
Because PoCL has to support at least two most recent LLVM major versions ([link](https://portablecl.org/docs/html/install.html#clang-llvm-target-compiler-supported-versions)).

lib/llvmopencl/AllocasToEntry.h
```diff
@@ -33,15 +33,13 @@
...
 public:
   static void registerWithPB(llvm::PassBuilder &B);
   llvm::PreservedAnalyses run(llvm::Function &F,
                               llvm::FunctionAnalysisManager &AM);
-  static bool isRequired() { return true; }
 };
```
linehill, line 49:
```suggestion
#if LLVM_MAJOR < 23
  static bool isRequired() { return true; }
#endif
};
```

linehill:
I was thinking that it might be cleaner to define this in some common header:
```c++
namespace pocl {
template<typename T>
class RequiredPassInfoMixin : public
#if LLVM_MAJOR >= 23
    : public llvm::RequiredPassInfoMixin<T> {
#else
    : public llvm::PassInfoMixin<T> {

    static bool isRequired() { return true; }
#endif
};
```

And change PoCL passes to derive from it:
```diff
-class FlattenGlobals : public llvm::PassInfoMixin<FlattenGlobals> {
+class FlattenGlobals : public pocl::RequiredPassInfoMixin<FlattenGlobals> {
```
But I'm fine moving forward with current state.

linehill (approved):
LGTM.

The Julia CI failure is seen in other PRs so it may be ignored.

PR #2259, opened 2026-07-29
Do not let the SPIR-V reader terminate the application (#2258)

### LINKED ISSUE #2258: clBuildProgram terminates the host process instead of returning CL_BUILD_PROGRAM_FAILURE for a SPIR-V module it cannot accept
Handed a SPIR-V module that the SPIR-V reader rejects, `clBuildProgram` kills the calling process. No error is returned, no build log is produced, and the application cannot recover. This is an OpenCL specification violation: `clBuildProgram` should return `CL_BUILD_PROGRAM_FAILURE` and leave the reason in the build log.

Reproducer: about 25 lines of `clCreateProgramWithIL` plus `clBuildProgram`, given a module that uses an extension the PoCL build does not accept (`SPV_INTEL_function_pointers` in our case). The process exits mid call. Before the fix the test below produces no stdout at all, because the process is gone before `clBuildProgram` returns.

Root cause is upstream in SPIRV-LLVM-Translator rather than in PoCL: `SPIRV::SPIRVDbgError` defaults to `SPIRVDbgErrorHandlingKinds::Exit` (`lib/SPIRV/libSPIRV/SPIRVDebug.cpp:52`), so `SPIRVErrorLog::checkError()` calls `std::exit()` inside `readSpirv()` (`lib/SPIRV/libSPIRV/SPIRVError.h:163`). PoCL calls `readSpirv()` without selecting the non fatal behaviour, so the error handling already present around both call sites in `lib/CL/pocl_llvm_spirv.cc` never runs. Filed upstream as KhronosGroup/SPIRV-LLVM-Translator#3939.

Not confined to one release:

| PoCL | behaviour |
|---|---|
| 2.16.0 (latest-llvm-21.0) | `exit(10)` inside `clBuildProgram` |
| 2.15.0 (macOS) | `exit(11)`, same path |
| 7.1, 2025.01.14 | module accepted, then LLVM assert `CallGraph::getOrInsertFunction ... Function not in current module!`, abort, no hang |
| 6.0 | `*** stack smashing detected ***` |

A fix and a regression test follow in a PR. One caveat worth stating up front: `SPIRVDebug.h`, which declares `SPIRVDbgErrorHandlingKinds` and `SPIRVDbgError`, is not among the installed `LLVMSPIRVLib` headers, so PoCL has to redeclare them. That works but is fragile, since a reordering of the enum upstream would silently select the wrong value. The durable fix is the upstream issue above; this is the local mitigation until then.

Downstream context: this surfaced through chipStar, where the process death happened while chipStar held a mutex across the backend compile, so its atexit handler then deadlocked against itself and the whole run hung until the CI timeout rather than failing.

pvelesko:
Fixes #2258. `clBuildProgram` killed the calling process for any SPIR-V module the reader rejected, because `SPIRV::SPIRVDbgError` defaults to `Exit` and `SPIRVErrorLog::checkError()` calls `std::exit()` inside `readSpirv()`, so the error handling already present around both call sites never ran. Selecting the non fatal behaviour before reading makes `clBuildProgram` return `CL_BUILD_PROGRAM_FAILURE` with the reason in the build log.

Caveat reviewers should see up front: `SPIRVDebug.h` is not among the installed `LLVMSPIRVLib` headers, so the two declarations are repeated locally. That is fragile, since reordering the enum upstream would silently select the wrong value. Filed upstream as KhronosGroup/SPIRV-LLVM-Translator#3939 suggesting the knob be exposed through the public `SPIRV::TranslatorOpts`; this is the local mitigation until then.

Regression test added, gated on `HAVE_LLVM_SPIRV_LIB`, with the SPIR-V module embedded as a `cl_uint` array so it stays reviewable and needs no data file. Before the fix it produces no stdout at all, because the process is gone before `clBuildProgram` returns.

Testing: 322 of 326 tests pass. The 4 failures (`test_as_type` and `test_bitselect`, loopvec and cbs) are pre-existing, verified by reverting only the fix file and re-running, which times out identically at 700 s. PoCL runtime suite 43/43.

Not exercised: validated on one CPU device with LLVM 21. The `convertBCorSPV` call site is reasoned-correct but has not been run on Level Zero or CUDA devices.

### DIFF
--- a/lib/CL/pocl_llvm_spirv.cc
+++ b/lib/CL/pocl_llvm_spirv.cc
@@ -80,6 +80,24 @@ static const pocl_version_t MaxSPIRVLibSupportedVersion{1, 1};
 static const pocl_version_t MaxSPIRVLibSupportedVersion{1, 0};
 #endif
 
+/* SPIRV::SPIRVDbgError selects what the SPIR-V reader does with a module it
+ * cannot accept. It defaults to SPIRVDbgErrorHandlingKinds::Exit, so
+ * SPIRVErrorLog::checkError() calls std::exit() and takes the whole
+ * application down. readSpirv() reports failures through its return value and
+ * its error string, and every caller below acts on those, so select the
+ * non fatal behaviour instead.
+ *
+ * The declarations are repeated here because SPIRVDebug.h, which declares
+ * them, is not among the installed LLVMSPIRVLib headers. */
+namespace SPIRV {
+enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+extern SPIRVDbgErrorHandlingKinds SPIRVDbgError;
+} // namespace SPIRV
+
+static void poclMakeSPIRVReaderErrorsNonFatal() {
+  SPIRV::SPIRVDbgError = SPIRV::SPIRVDbgErrorHandlingKinds::Ignore;
+}
+
 #else // HAVE_LLVM_SPIRV_LIB
 static const pocl_version_t MaxSPIRVLibSupportedVersion{1, 5};
 #endif
@@ -266,6 +284,7 @@ int pocl_regen_spirv_binary(cl_program Program, cl_uint DeviceI) {
   char *Content = nullptr;
   uint64_t ContentSize = 0;
 
+  poclMakeSPIRVReaderErrorsNonFatal();
   if (!readSpirv(LLVMCtx, Opts, InputSS, Mod, Errors)) {
     POCL_MSG_ERR("LLVMSPIRVLib failed to read SPIR-V with errors:\n%s\n",
                  Errors.c_str());
@@ -576,6 +595,7 @@ static int convertBCorSPV(char *InputPath,
     Mod = nullptr;
 
     // TODO maybe use context from program ?
+    poclMakeSPIRVReaderErrorsNonFatal();
     if (!readSpirv(LLVMCtx, Opts, InputSS, Mod, Errors)) {
       BuildLog->append("LLVMSPIRVLib: Write failed with errors:\n");
       BuildLog->append(Errors.c_str());
--- a/tests/runtime/test_clBuildProgram_unsupported_spirv.c
+++ b/tests/runtime/test_clBuildProgram_unsupported_spirv.c
@@ -1,0 +1,144 @@
+/* Tests that clBuildProgram reports an error for a SPIR-V module it cannot
+   accept, instead of terminating the host process.
+
+   clBuildProgram() reaches llvm::readSpirv(), whose documented contract is to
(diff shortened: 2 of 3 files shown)


pjaaskel:
Thanks. Did you use AI on these commits? If, please add Co-authored-by: line to the commits which mentions the tool/model.

franz:
Strange, the CodeQL CI is failing the build with 
```
  [2026-07-29 17:47:38] [build-stdout] /usr/bin/ld: ../../lib/CL/libpocl.so.2.17.0: undefined reference to `SPIRV::SPIRVDbgError'
```
but none of the other CI builds are failing.

tests/runtime/test_clBuildProgram_unsupported_spirv.c
```diff
@@ -0,0 +1,144 @@
...
+#include <stdlib.h>
+
+/* A SPIR-V 1.4 kernel that calls through a function pointer, so it declares
+   OpCapability FunctionPointersINTEL and OpExtension
+   "SPV_INTEL_function_pointers". That extension is not in the set PoCL enables
+   for the CPU devices, so the SPIR-V reader rejects the module with
+   "input SPIR-V module uses extension 'SPV_INTEL_function_pointers' which were
+   disabled by --spirv-ext option". Any module the reader rejects reproduces
+   this equally well. */
+static const cl_uint UnsupportedExtensionSpirv[] = {
```
pjaaskel, line 45: Please add a .spv instead, there are other tests that do this. Easier to inspect/modify.

lib/CL/pocl_llvm_spirv.cc
```diff
@@ -80,6 +80,24 @@ static const pocl_version_t MaxSPIRVLibSupportedVersion{1, 1};
...
+ * SPIRVErrorLog::checkError() calls std::exit() and takes the whole
+ * application down. readSpirv() reports failures through its return value and
+ * its error string, and every caller below acts on those, so select the
+ * non fatal behaviour instead.
+ *
+ * The declarations are repeated here because SPIRVDebug.h, which declares
+ * them, is not among the installed LLVMSPIRVLib headers. */
+namespace SPIRV {
+enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+extern SPIRVDbgErrorHandlingKinds SPIRVDbgError;
```
pjaaskel, line 94: Has this global been in the SPIRVDebug.h always? Wondering if it has been renamed at some point, and the CI fails as it's not linked in with that name.

pjaaskel (changes requested):

pvelesko:
main fix in upstream https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3941 
will open a new PR once that one is merged

PR #2261, opened 2026-07-30
update CIs to use LLVM 23

franz:


pjaaskel:
Can we do something to fix the Julia CI? @maleadt

linehill:
> Can we do something to fix the Julia CI? @maleadt

It seems, the #2264 fixes that - the failures seems to be half type related and the PR has fixes to it and the Julia CI is passing there.

maleadt:
Hmm, I wasn't aware of additional issues (at least not past https://github.com/pocl/pocl/pull/2235). Good to know that https://github.com/pocl/pocl/pull/2264 fixes those, at least?

franz:
ok, rebased onto main after merging #2264 seems to have resolved the failing Julia CIs

linehill (approved):
LGTM.

PR #2262, opened 2026-08-03
[llvmopencl] Fix LoopInfo::verify() call for LLVM 24

woodychow:
In LLVM 24, LoopInfo::verify(const DominatorTreeBase<BlockT, false> &DomTree) const was changed to LoopInfo::verify() const, removing the DominatorTree argument: https://github.com/llvm/llvm-project/commit/022bf7ba97fa1897bf52cf3db2e3fa42cd19b29f (PR https://github.com/llvm/llvm-project/pull/212414)

Add an #if LLVM_MAJOR >= 24 guard around LI.verify() in BarrierTailReplication.cc to support LLVM 24 while preserving backward compatibility with older LLVM versions.


franz (approved):
LGTM

PR #2263, opened 2026-08-03
Validate that memory objects passed to image kernel arguments are valid images

woodychow:
When a memory object is passed to a kernel argument of type POCL_ARG_TYPE_IMAGE, PoCL checked IS_CL_OBJECT_VALID, but did not verify whether the memory object was actually an image (is_image == CL_TRUE). Consequently, passing a buffer object to an image argument returned CL_SUCCESS instead of CL_INVALID_MEM_OBJECT as required by the OpenCL specification:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetKernelArg

This fixes the negative_set_kernel_arg_invalid_image test in CTS.


franz (approved):
LGTM

PR #2264, opened 2026-08-04
Fix half CTS conversion tests

woodychow:
Update lib/kernel/convert_type.py (and regenerate convert_type.cl) to:
- Generate scalar half-to-int saturated conversions and properly clamp overflowed values in directed rounding modes (_rtn, _rtp, _rtz).
- Add half-to-integer conversion pairs to dfli so directed rounding conversions use saturated conversions (convert_{SRC}_sat) when comparing rounded float values back against integer inputs, preventing short/ushort overflow on half_rtp_short, half_rtn_short, and half_rtz_short.
- Index rtn_thresholds by (dst, src) so half_rtn_short and half_rtz_short return 0x1.ffcp+14 (32752.0) for short inputs >= 32753.
- Cast vector comparison operand literals to the element type of the vector (e.g. ({SRC}){THRESH}) to avoid Clang rank mismatch build errors.

Specifically, this fixes 45 OpenCL CTS conversions tests on PoCL, including:
- conversions/half_rt[n|p|z]_[int|long|short|uint|ulong|ushort]
- conversions/[char|int|short|uchar|uint|ushort]_sat_[rte|rtn|rtp|rtz]_half
- conversions/[char|int|short|uchar|uint|ushort]_sat_half

### DIFF
--- a/lib/kernel/convert_type.cl
+++ b/lib/kernel/convert_type.cl
@@ -29656,10 +29656,25 @@ ulong16 convert_ulong16_sat(ulong16 x)
 
 #if defined(cl_khr_fp16)
 _CL_ALWAYSINLINE _CL_OVERLOADABLE _CL_READNONE
+char convert_char_sat(half x)
+{
+  x = select(x, (half)(0), (short)isnan(x));
+  char y = convert_char(x);
+  y = select(y, (char)CHAR_MIN, convert_char(x < (half)CHAR_MIN));
+  y = select(y, (char)CHAR_MAX, convert_char(x >= (half)CHAR_MAX));
+  return y;
+}
+#endif
+
+#if defined(cl_khr_fp16)
+_CL_ALWAYSINLINE _CL_OVERLOADABLE _CL_READNONE
 char2 convert_char2_sat(half2 x)
 {
-  x = clamp(x, (half)CHAR_MIN, (half)CHAR_MAX);
-  return convert_char2(x);
+  x = select(x, (half2)(0), (short2)isnan(x));
+  char2 y = convert_char2(x);
+  y = select(y, (char2)CHAR_MIN, convert_char2(x < (half2)CHAR_MIN));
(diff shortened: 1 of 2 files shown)


lib/kernel/convert_type.cl
```diff
@@ -29654,353 +29654,579 @@ ulong16 convert_ulong16_sat(ulong16 x)
...
 #endif
 
+#if defined(cl_khr_fp16)
+_CL_ALWAYSINLINE _CL_OVERLOADABLE _CL_READNONE
+char convert_char_sat(half x)
+{
+  x = select(x, (half)(0), (short)isnan(x));
+  char y = convert_char(x);
+  y = select(y, (char)CHAR_MIN, convert_char(x < (half)(-0x1p+7f)));
+  y = select(y, (char)CHAR_MAX, convert_char(x >= (half)(0x1p+7f)));
```
linehill: Can't we use `<X>_{MIN,MAX}` instead of floating point literals for better readability? 
```suggestion
  y = select(y, (char)CHAR_MIN, convert_char(x < (half)(CHAR_MIN)));
  y = select(y, (char)CHAR_MAX, convert_char(x >= (half)(CHAR_MAX)));
```

woodychow: I did not introduce limit_max_float and limit_min_float in convert_type.py. It might be better to remove them in another PR.

woodychow: No CTS regression on my side with `limit_max_float` and `limit_min_float` removed.

woodychow: Added another commit. Please take another look.

linehill (approved):
I don't fully grasp what the changes in the `convert_*_{rtz,rtp,rte}` are doing but if they are fixing CTS issues, I'm giving this a pass.

franz:
Tested the half-type conversions with CTS test, they're now passing.

PR #2269, opened 2026-08-05
Fix tail replication and load rematerialization across barriers

### LINKED ISSUE #388: AcceleratedKernels.cumsum runs endlessly when using pocl's cpu backend on Intel CPU
I've been able to reproduce it on two Intel devices. Would be nice if someone could test on AMD.
Rusticl works just fine here.

Julia: 1.11.3
OpenCL: 0.10.5
pocl: pocl_jll.jl (v7.0.0+15)

maleadt:
Fixes two work-group lowering bugs encountered in OpenCL.jl (@christiangnrd the AK radix sort and `accumulate!` issues):

- Share replicated join tails among edges leaving the same parallel region.
- Do not rematerialize loads whose values may change across parallel regions.

`BarrierTailReplication` previously created a separate tail for every incoming edge to a join block. When multiple edges came from one parallel region, this gave the region multiple exits. `WorkitemLoops` could then discard exits not taken by the first work-item, which is invalid for divergent branches.
Replicas are now keyed by the region-entry barrier and join block. Edges from the same region share one tail, while different regions retain separate copies.

`WorkitemLoops` also rematerialized loads in later regions. Such a load can observe writes made after its original definition, breaking patterns such as:

```c
value = scratch[neighbor];
barrier(...);
scratch[me] = value;
```

Loads are now rematerialized only from:
- readonly, non-aliasing arguments;
- constant globals; or
- PoCL work-item geometry variables.

Other loaded values are preserved in per-work-item context storage.

Added a regression covering local-memory loads and aliased global arguments across a barrier. The tail-replication failure does not currently have a direct regression. Its required multi-edge region shape arises after LLVM control-flow transformations and could not be reproduced reliably from OpenCL C.

Since I'm not particularly familiar with either of these passes, I'm relying on LLMs for my understanding, so I'd appreciate thorough review.

Also fixes https://github.com/JuliaGPU/OpenCL.jl/issues/388


pjaaskel:
We have also pruned BTR out in the new kernel compiler branch so this issue is likely not there.

maleadt:
Oh great. I also had Claude `spirv-reduce` the original MWE, so I committed that as a test for the BTR issue.

pjaaskel:
@linehill I vaguely remember we have fixed something like that in the remat in the priv branch. Please check.

pjaaskel (approved):

franz:
do we want this cherry-picked into 7.2 branch ?

linehill:
> @linehill I vaguely remember we have fixed something like that in the remat in the priv branch. Please check.

We do bu t there are differences. In the private branch we allow rematerializing context saves when they don't cross stores to the same context array. What we don't have in the private branch is rematerialization of loads from constant memory.

PR #2273, opened 2026-08-10
few backports from 7.2

franz:


.github/workflows/build_msvc.yml
```diff
@@ -87,6 +87,9 @@ jobs:
       - name: Load Env vars
         id: load-env
         run: |
+          # this DLL exists also in C:/Windows/system32 ; this causes a problem
+          # for CMake when resolving runtime DLL dependencies, hence remove it
```
linehill: Can there be a bit more detail about the problem here?

linehill (approved):
LGTM.

PR #2279, opened 2026-08-13
Minor (mostly AlmaIF driver related) fixes

soronen:
Hi, 

I was working on updated device driver for ttasim, and although there is little reason to merge it since the driver uses a private branch of OpenASIP, along the way I made some smaller changes that could be useful to other users.

Most of these are minor fixes in the AlmaIF driver (giving a struct a function ptr to wrong function, fixing up the version ID to match what OpenCL conformance test suite expects, changing debug prints to use base 16 for addresses, increasing polling interval so it isn't as spammy, etc.)

The broadest change touches `lib/llvmopencl/LLVMUtils.cc`, since `regenerate_kernel_metadata()` gets called by all SPMD devices now.  It fixes two crashes I had with kernel compilation running GROMACS.

### DIFF
--- a/lib/CL/devices/almaif/AlmaifShared.hh
+++ b/lib/CL/devices/almaif/AlmaifShared.hh
@@ -128,7 +128,7 @@ struct AQLQueueInfo {
 #define ALMAIF_CQ_WRITE (offsetof(AQLQueueInfo, write_index))
 #define ALMAIF_CQ_LENGTH (offsetof(AQLQueueInfo, size))
 
-#define ALMAIF_DRIVER_SLEEP 200
+#define ALMAIF_DRIVER_SLEEP 10000
 
 enum ALMAIF_DEVICE_TYPE : size_t {
   POCL_ALMAIFDEVICE_XRT = 0xA,
--- a/lib/llvmopencl/LLVMUtils.cc
+++ b/lib/llvmopencl/LLVMUtils.cc
@@ -157,31 +157,39 @@ regenerate_kernel_metadata(llvm::Module &M, FunctionMapping &kernels)
   NamedMDNode *WGSizes = M.getNamedMetadata("opencl.kernel_wg_size_info");
   if (WGSizes != NULL && WGSizes->getNumOperands() > 0)
     {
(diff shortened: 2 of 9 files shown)


lib/llvmopencl/LLVMUtils.cc
```diff
@@ -157,31 +157,39 @@ regenerate_kernel_metadata(llvm::Module &M, FunctionMapping &kernels)
...
+        SmallVector<Metadata *, 8> operands;
+        operands.push_back(llvm::ValueAsMetadata::get(NewKernel));
+        for (unsigned opr = 1; opr < wgsizeMD->getNumOperands(); ++opr) {
+          operands.push_back(wgsizeMD->getOperand(opr));
+        }
+        MDNode *new_wg_md = MDNode::get(M.getContext(), operands);
+        ValidNodes.push_back(new_wg_md);
+      }
         }
+        WGSizes->clearOperands();
```
pjaaskel: This segment looks broken. Carved out change?

pjaaskel (approved):
After CI passes LGTM.

lib/CL/devices/almaif/AlmaifShared.hh
```diff
@@ -128,7 +128,7 @@ struct AQLQueueInfo {
 #define ALMAIF_CQ_WRITE (offsetof(AQLQueueInfo, write_index))
 #define ALMAIF_CQ_LENGTH (offsetof(AQLQueueInfo, size))
 
-#define ALMAIF_DRIVER_SLEEP 200
+#define ALMAIF_DRIVER_SLEEP 10000
```
gtkiku, line 131: Pedantry: Commit message says *reduce* polling interval, but this *increases* it.

soronen: Thanks, renamed the commit

lib/llvmopencl/LLVMUtils.cc
soronen: Missed that part with the formatter, fixed.

PR #2282, opened 2026-08-18
Do not let the SPIR-V reader terminate the application (#2258)

### LINKED ISSUE #2258: clBuildProgram terminates the host process instead of returning CL_BUILD_PROGRAM_FAILURE for a SPIR-V module it cannot accept

pvelesko:
Replaces #2259, which redeclared the internal `SPIRV::SPIRVDbgError` locally and failed the CodeQL link with `undefined reference to SPIRV::SPIRVDbgError`. The knob is now public API via KhronosGroup/SPIRV-LLVM-Translator#3941 (merged, from issue #3939), so this selects it through `TranslatorOpts` instead.

`clBuildProgram` killed the calling process for any SPIR-V module the reader rejected, because `SPIRVErrorLog::checkError()` calls `std::exit()` and the error handling already present around both `readSpirv()` call sites never ran. Setting the kind in `setupTranslOpts()` covers both readers plus `writeSpirv()`, since the option travels with the options object.

Only translator versions carrying #3941 have the setter, so `HAVE_LLVM_SPIRV_LIB_ERROR_HANDLING` gates its use, detected by looking for `setErrorHandlingKind` in the installed `LLVMSPIRVOpts.h`. Older translators build unchanged and keep the old behaviour, and the regression test is gated on the same flag.

Third commit is separate: `pocl_regen_spirv_binary()` reported the reason through `POCL_MSG_ERR`, which is compiled behind `pocl_debug_messages_filter`, so `CL_PROGRAM_BUILD_LOG` carried only the generic device line. It now goes through `appendToProgramBuildLog()`, matching what `convertBCorSPV()` already did.

Testing: LLVM 22.1.5 with a translator carrying #3941, 332/332 ctest pass. Reproducer verified to fail without the fix (exit 10, no stdout) and without the build log plumbing (assertion at line 137). LLVM 21 with an in tree translator builds clean, flag 0, test not registered.

Not exercised: one CPU device only. The `convertBCorSPV()` and `getSpecConstInfo()` call sites are correct by inspection but not run; the test module has a valid magic number so it does not reach the spec constant path.

Used Claude Opus 5, noted in the commit trailers per the request on #2259.

Fixes #2258


franz (approved):
LGTM

PR #2286, opened 2026-08-20
cuda: Add support for IMAGE1D_BUFFER

jansol:
For some reason llama.cpp has a hard requirement on images, but only the IMAGE1D_BUFFER type. This implements just enough image support to cover that need.

The kernel functions are taken as-is from the CPU kernels.


pjaaskel (approved):
After the CIs pass and changelog.

PR #2289, opened 2026-08-20
remote: implement server-side barrier and marker commands

jansol:
Shorting these on the client was a handy shortcut earlier, but it's causing problems if servers receive commands that depend on the marker/barrier..


pjaaskel (approved):
LGTM, but please remember to update the change log.

PR #2296, opened 2026-08-20
remote: Kernel build fixes

jansol:
Mostly trial and error workarounds for various driver behaviours that I'm not convinced are entirely up to spec...

Fetching kernel binaries and trying to recreate the same programs from them later was also horribly broken. This PR makes that kind of application-driven caching work, but pocl's automatic client-side caching when building the same kernels again from source does not seem to kick in.


pjaaskel (approved):

pjaaskel:
I quickly glimpsed over these. Good useful fixes. Please check the CIs and document the new features and fixes to the change log. The PR stacking feature seems nice.

Issue #1750, opened 2024-12-31
Issue with Segmentation Fault Using Proxy Driver in PoCL with clblast/clblas Functions  Question


He32100:
I am using proxy driver to move the Mali GPU to the PoCL platform. The cmake command I used is:
`sudo cmake -DENABLE_ICD=0 -DENABLE_LOADABLE_DRIVERS=0 -DENABLE_PROXY_DEVICE=1 -DWITH_LLVM_CONFIG=/usr/bin/llvm-config-17 ..`
And I have set the environment variable:`export POCL_DEVICES="proxy pthread"`
Now, I can find both the CPU and GPU devices in PoCL platform. The output is as follows:
```
rock@rockpi-4b:~/pocl-55b7a4e6c51b87f29fd48d701670885556f8cb7f/build_proxy$ POCL_BUILDING=1 POCL_DEVICES="proxy cpu" ./bin/poclcc -l
arm_release_ver of this libmali is 'r18p0-01rel0', rk_so_ver is '5'.
LIST OF DEVICES:
0:
  Vendor:   ARM
  Name:     cpu-cortex-a72-cortex-a53
  Version:  OpenCL 3.0 PoCL HSTR: cpu-aarch64-unknown-linux-gnu-cortex-a72
1:
  Vendor:   ARM
  Name:     Mali-T860
  Version:  OpenCL 1.2 v1.r18p0-01rel0.db8fd8841edf74cef96ef24ce665edac HSTR: pocl-proxy: Mali-T860
```
Before including any `CL/*.h` header files, I added `#include "rename_opencl.h".` I want to run an OpenCL program that calls the CLBlast library function `CLBlastSgemm` or the clBLAS library function` clblasSgemm `to perform computations and test if the devices on the PoCL platform are usable. However, regardless of whether I use the CPU or GPU, the program crashes with a segmentation fault.

On the other hand, if I don’t use PoCL and instead use OpenCL, the program runs normally. Additionally, I recompiled PoCL using the cmake command:`cmake -DWITH_LLVM_CONFIG=/usr/bin/llvm-config-17 ..`
And when I use the CPU device from PoCL, the program works fine.
The issue occurs when I build PoCL with the proxy driver and use the clblast or clblas library functions in the program, causing a segmentation fault. Does this problem have anything to do with using the proxy driver to build PoCL?

Here is my OpenCL program:
```
[...]

PR #2297, opened 2026-08-23
OpenVX_CTS: bump to openvx_1.3.2 and narrow test filter

kiritigowda:
## Summary

- Bump the OpenVX Conformance Test Suite checkout from `openvx_1.3` to `openvx_1.3.2`.
- Narrow the `vx_test_conformance` exclusion filter down to only the tests that are genuinely broken upstream, so the suite now exercises the full vision conformance surface instead of over-excluding.

## Filter change

The previous filter excluded 8 broad patterns. After validating against MIVisionX built with the OpenCL backend on PoCL's CPU device, most of those exclusions were stale — those tests pass now. The filter is reduced to just the ORB-scale Gaussian pyramid cases with `VX_BORDER_UNDEFINED`:

```
--filter=*:-GaussianPyramid.GraphProcessing/*VX_BORDER_UNDEFINED*VX_SCALE_PYRAMID_ORB
         :-GaussianPyramid.ImmediateProcessing/*VX_BORDER_UNDEFINED*VX_SCALE_PYRAMID_ORB
```

Result on PoCL (CPU device, MIVisionX OpenCL backend): **5816 / 5816 pass, 0 failures** with these 8 cases excluded.

## Why these 8 remain excluded

The excluded cases are a **known upstream MIVisionX bug in the non-AMD OpenCL path**, not a PoCL issue:

- MIVisionX's own CPU Reference model passes these cases.
- They pass on native AMD hardware (verified on gfx1151: 12/12 ORB tests, full vision suite 5816/5816 identical stock vs patched).
- They fail only on non-AMD OpenCL devices (e.g. PoCL), where MIVisionX substitutes software emulation for `cl_amd_media_ops`.

[ROCm/MIVisionX#1750](https://github.com/ROCm/MIVisionX/pull/1750) (merged into `develop`) fixed one contributing defect in this path — the `amd_bytealign` emulation masked the byte offset with `& 31` instead of `& 3`. That fix is necessary and correct, but a **second, independent defect** in the ORB `GaussianPyramid` OpenCL kernel remains: with #1750 applied, full unfiltered conformance on PoCL still fails these same 8 cases. The exclusion is therefore retained for now and documented in-code with a pointer to the upstream work. It can be dropped once the remaining upstream fix lands and is verified here.

## Test plan

- [x] Build PoCL (ICD) and MIVisionX (`-DBACKEND=OPENCL`) against PoCL's CPU device
- [x] `openvx_1.3.2` CTS builds and runs against the OpenCL-backed MIVisionX
- [x] Full vision conformance passes with the narrowed filter (5816/5816)
- [x] Confirmed the 8 excluded cases are the only unfiltered failures, all ORB + `VX_BORDER_UNDEFINED`
- [x] Confirmed no PoCL-side regressions from the previously-excluded (now-passing) tests

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -80,27 +80,29 @@ jobs:
         run: |
           cat ${{ github.workspace }}/.github/variables.txt >> $GITHUB_ENV
 
-      # Build MIVisionX (the OpenVX implementation the CTS links against) from
-      # the latest develop with the OpenCL backend, instead of relying on the
-      # MIVisionX version preinstalled in /opt/rocm on the runner. This keeps the
-      # tested implementation current, so fixes merged upstream are exercised and
-      # the conformance filter reflects only genuinely-open MIVisionX defects.
-      - name: Build MIVisionX (develop, OpenCL backend)
+      # Build MIVisionX (the OpenVX implementation the CTS links against) with the
+      # OpenCL backend, instead of relying on the MIVisionX version preinstalled
+      # in /opt/rocm on the runner (which can lag upstream). A dedicated, tested
+      # branch (pocl-VX) is used rather than develop so the pinned implementation
+      # is stable and the conformance filter reflects only genuinely-open
+      # MIVisionX defects rather than transient develop breakage.
+      - name: Build MIVisionX (pocl-VX branch, OpenCL backend)
         id: build_mivisionx
         if: ${{ matrix.config == 'openvx' }}
         timeout-minutes: 90
         run: |
+          MIVISIONX_BRANCH=pocl-VX
           MIVISIONX_SRC=${EXAMPLES_DIR}/MIVisionX
           MIVISIONX_ROOT=${EXAMPLES_DIR}/mivisionx-install
           rm -rf "${MIVISIONX_ROOT}"
           if [ ! -d "${MIVISIONX_SRC}/.git" ]; then
             rm -rf "${MIVISIONX_SRC}"
-            git clone --depth 1 -b develop https://github.com/ROCm/MIVisionX.git "${MIVISIONX_SRC}"
+            git clone --depth 1 -b "${MIVISIONX_BRANCH}" https://github.com/ROCm/MIVisionX.git "${MIVISIONX_SRC}"
           else
-            git -C "${MIVISIONX_SRC}" fetch --depth 1 origin develop
+            git -C "${MIVISIONX_SRC}" fetch --depth 1 origin "${MIVISIONX_BRANCH}"
             git -C "${MIVISIONX_SRC}" reset --hard FETCH_HEAD
           fi
-          echo "MIVisionX develop HEAD: $(git -C "${MIVISIONX_SRC}" rev-parse --short HEAD)"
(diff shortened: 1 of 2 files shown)


franz:
Thanks!

LGTM overall, however I remember disabling a few (1-2 ?)  testcases (related to Graphs) that were ocassionally failing on the CI (possibly a race condition somewhere). Actually it's in commit b8f5c9cd64b0d, testcase `Graph.GraphState`. We can re-run the CI a few times to see how reliable is it with the new filter.

kiritigowda:
### CI fix: build MIVisionX from develop instead of the runner's preinstalled copy

The `openvx` job was failing on ~27 tests (`SmokeTestBase.vxRegisterKernelLibrary`, `Array.vxCreateArray`, `Image.UniformImage`, `MinMaxLoc.OnRandom`, ...) — **none of them the ORB cases this PR filters**. Root cause: the CTS was linking against the MIVisionX preinstalled in `/opt/rocm` on the runner, which lags MIVisionX `develop`. The newer `openvx_1.3.2` CTS exposed failures already fixed upstream. All 27 pass against MIVisionX develop.

Latest commit makes the `openvx` CI job build MIVisionX `develop` (OpenCL backend) and point the CTS at it via a new `CUSTOM_MIVISIONX_ROOT` testsuite option, so the tested implementation stays current and isn't tied to runner provisioning. When `CUSTOM_MIVISIONX_ROOT` is unset the testsuite still defaults to `/opt/rocm`, so local/dev behavior is unchanged.

Verified locally end-to-end (same build commands as the workflow) against MIVisionX develop on a PoCL CPU device:
```
#REPORT: FILTERED 15286 8195 5816 5816 5816 0 (version 1.3.2)   -> 5816/5816 pass, 0 fail
```

.github/workflows/build_linux.yml
```diff
@@ -80,6 +80,39 @@ jobs:
         run: |
           cat ${{ github.workspace }}/.github/variables.txt >> $GITHUB_ENV
 
+      # Build MIVisionX (the OpenVX implementation the CTS links against) from
+      # the latest develop with the OpenCL backend, instead of relying on the
```
pjaaskel: I'd rather see a known PoCL-good MIVisionX commit here. Otherwise PoCL test suite can start failing because of MIVisionX regressions.

pjaaskel (changes requested):

kiritigowda:
Addressed the review feedback about tracking `develop`: the CI now builds MIVisionX from the dedicated, tested [`pocl-VX`](https://github.com/ROCm/MIVisionX/tree/pocl-VX) branch instead of `develop`, so the pinned OpenVX implementation is stable and not exposed to transient `develop` breakage.

`pocl-VX` is currently at `9aad3ec` (includes the merged `amd_bytealign` emulation fix, ROCm/MIVisionX#1750). Verified locally end-to-end with the exact workflow commands (clone `pocl-VX` → build `openvx`/`vxu` with `-DBACKEND=OPENCL` → install the `amd_openvx/openvx` subproject → build + run the CTS on a PoCL CPU device): `5816/5816` pass, `0` fail with the filter.

pjaaskel (approved):
LGTM, thanks.

PR #2305, opened 2026-09-04
clSetKernelExecInfo: keep SVM and USM pointer lists independent

maleadt:
CL_KERNEL_EXEC_INFO_SVM_PTRS and CL_KERNEL_EXEC_INFO_USM_PTRS_INTEL both populate the kernel indirect pointer list. Clearing the whole list on every call made setting either parameter discard the pointers registered by the other.

Record the parameter that registered each pointer and only replace matching entries. Add a regression test that sets and clears the SVM and USM lists in both orders.


franz (approved):
LGTM

PR #2306, opened 2026-09-04
Command-buffers: retire the submission before publishing CL_COMPLETE

bricevideau-ai:
## Problem

Replaying a single finalized command-buffer in a loop intermittently fails with `CL_INVALID_OPERATION` from `clEnqueueCommandBufferKHR`, at a **random** iteration (observed 386 … 6594), even though the application waits on each submission's completion event before re-enqueueing.

That application is conformant. pocl advertises `cl_khr_command_buffer` 0.9.6, whose state machine says:

> **Pending**: Once a command-buffer has been enqueued to a command-queue it enters the Pending state **until completion**, at which point it moves back to the Executable state.

Waiting on the completion event means completion has happened, so the buffer must already be back in Executable and the next enqueue must succeed. (The argument also holds under 0.9.8, which removed the Pending state and instead defines *simultaneous use* against previous submissions "not in the `CL_COMPLETE` state", explicitly naming events as a way to express the prerequisite — but 0.9.6 is the revision this patch is written against.)

## Cause

`pocl_update_event_finished` writes `event->status = CL_COMPLETE` and drops the event lock ~30 lines **before** retiring the submission (decrementing `command_buffer->pending` and restoring the Executable state).

`pocl_pthread_wait_event` tests `event->status > CL_COMPLETE` **before** waiting on the condition variable:

```c
POCL_LOCK_OBJ (event);
while (event->status > CL_COMPLETE)
    POCL_WAIT_COND (e_d->event_cond, event->pocl_lock);
POCL_UNLOCK_OBJ (event);
```

So an application thread entering `clWaitForEvents` inside that window takes the (free) event lock, sees `CL_COMPLETE`, **never waits at all**, and its next `clEnqueueCommandBufferKHR` is rejected by `pocl_cmdbuf_is_ready`.

An application that blocks in `clWaitForEvents` *before* the buffer completes never hits this — it is woken from `notify_event_finished`, which runs after the retire. That explains every observed property: it needs host work between enqueue and wait, the failing iteration is random, and it disappears under an API tracer.

## Evidence

Standalone reproducer, varying only the host work between enqueue and wait:

| spin iterations | first failure |
|---|---|
| 0 | never (20000 replays × 4 trials) |
| 2 000 | iteration 6594 |
| 20 000 | iteration 897 |
| 200 000 | iteration 386 |

The failure rate scales monotonically with the size of the window, which is the window this patch closes.

## Fix

Retire the submission before publishing the terminal status.

Only the bookkeeping moves earlier. It is ordered against the status write by the event lock already held, and `command_buffer->mutex` is a leaf — no event or queue lock is taken under it anywhere (cf. `pocl_cmdbuf_record_command`, `clEnqueueCommandBufferKHR`) — so nesting it there adds no lock-order edge. `clReleaseCommandBufferKHR` stays outside the locks: it can free the object.

## Testing

- `tests/runtime/test_command_buffer` gains 2000 replays with host work between enqueue and wait. **Fails without this change, passes with it.**
- Reproducer: 5 × 20000 replays clean at the worst spin, where unpatched failed at 386.

Tested on the CPU (pthread) driver.

### DIFF
--- a/tests/runtime/command_buffer_common.h
+++ b/tests/runtime/command_buffer_common.h
@@ -1,0 +1,103 @@
+/* Shared setup for the cl_khr_command_buffer runtime tests.
+
+   Copyright (c) 2026 pocl developers
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to
+   deal in the Software without restriction, including without limitation the
+   rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+   sell copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+   IN THE SOFTWARE.
+*/
+
+#ifndef POCL_TESTS_COMMAND_BUFFER_COMMON_H
+#define POCL_TESTS_COMMAND_BUFFER_COMMON_H
+
+#include <string.h>
+
+#include "poclu.h"
+
+#define CMDBUF_STR(x) #x
+
+/* The extension entry points the command-buffer tests use. */
+struct cmdbuf_ext
+{
+  clCreateCommandBufferKHR_fn clCreateCommandBufferKHR;
+  clCommandCopyBufferKHR_fn clCommandCopyBufferKHR;
+  clCommandCopyBufferRectKHR_fn clCommandCopyBufferRectKHR;
+  clCommandFillBufferKHR_fn clCommandFillBufferKHR;
+  clCommandNDRangeKernelKHR_fn clCommandNDRangeKernelKHR;
+  clCommandBarrierWithWaitListKHR_fn clCommandBarrierWithWaitListKHR;
+  clFinalizeCommandBufferKHR_fn clFinalizeCommandBufferKHR;
+  clEnqueueCommandBufferKHR_fn clEnqueueCommandBufferKHR;
+  clReleaseCommandBufferKHR_fn clReleaseCommandBufferKHR;
+  clGetCommandBufferInfoKHR_fn clGetCommandBufferInfoKHR;
+};
+
+/* Resolve the entry points. Returns 0 on success, or 77 (the ctest skip code)
+   if the platform does not implement command buffers. */
+static inline int
+cmdbuf_load_ext (cl_platform_id platform, struct cmdbuf_ext *ext)
+{
+#define CMDBUF_GET(name)                                                      \
+  ext->name = clGetExtensionFunctionAddressForPlatform (platform, #name)
+
+  CMDBUF_GET (clCreateCommandBufferKHR);
+  if (ext->clCreateCommandBufferKHR == NULL)
+    {
+      printf ("Command buffers are not supported, skipping test\n");
+      return 77;
+    }
+  CMDBUF_GET (clCommandCopyBufferKHR);
+  CMDBUF_GET (clCommandCopyBufferRectKHR);
+  CMDBUF_GET (clCommandFillBufferKHR);
+  CMDBUF_GET (clCommandNDRangeKernelKHR);
+  CMDBUF_GET (clCommandBarrierWithWaitListKHR);
+  CMDBUF_GET (clFinalizeCommandBufferKHR);
+  CMDBUF_GET (clEnqueueCommandBufferKHR);
+  CMDBUF_GET (clReleaseCommandBufferKHR);
+  CMDBUF_GET (clGetCommandBufferInfoKHR);
+  return 0;
+
+#undef CMDBUF_GET
+}
+
+/* The kernel the command-buffer tests record. */
+static const char *cmdbuf_vector_addition_src = CMDBUF_STR (
+    kernel void vector_addition (global const int *tile1,
+                                 global const int *tile2, global int *res) {
+      size_t index = get_global_id (0);
+      res[index] = tile1[index] + tile2[index];
+    });
+
+/* Build `vector_addition` for `device`. */
+static inline cl_int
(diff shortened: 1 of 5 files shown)


jansol (changes requested):
The code changes are fine but the comments are way too wordy and not helpful to someone reading the code by itself.

The test running a ton of parallel enqueues feels like it should be a separate test case from test_command_buffer (although the setup and command buffer recording code makes sense to share between them). The comment in the test should also be split up and the parts should be placed in the relevant points of the code.

bricevideau-ai:
Thanks — all three addressed in the force-pushed update.

**Comments.** The in-function comment is down from 45 lines to 10: the ordering invariant, and why `clReleaseCommandBufferKHR()` can't be hoisted along with the bookkeeping (on the last reference it calls `PoCLReleaseCommandQueue()`, which locks the queue this function is still holding). Everything about how the bug was found now lives only in the commit message.

**Test split.** The replay loop is its own case, `tests/runtime/test_command_buffer_replay.c`. `test_command_buffer.c` is back to byte-identical with main. The shared pieces — extension-pointer loading, kernel source, program build — moved to a new `tests/runtime/command_buffer_common.h`, following the `image_test_utils.h` convention; the tiled buffer setup stays private to `test_command_buffer`, since the replay test only needs one small set.

I also dropped the `CL_COMMAND_BUFFER_STATE_KHR` query I had added to `test_command_buffer`. It goes with a separate question — pocl marks an in-flight buffer `FINALIZED`, which reads oddly against 0.9.8 where that state means "an argument is unset" — and it had no business in this PR.

**Test comment.** Split into four, placed where each part applies: why no `SIMULTANEOUS_USE` flag, why more than one command, why the host-work spin exists, and what the loop asserts.

One clarification on "a ton of parallel enqueues" — they are strictly serialised. Each iteration blocks on the completion event before the next enqueue, so the previous submission is in `CL_COMPLETE` and this is not simultaneous use. The host work between enqueue and wait is the load-bearing part: without it the application is always *already* blocked in `clWaitForEvents()` when the buffer completes, and that interleaving is the one that never hit the bug. As written the test fails at around replay 1676 without the fix and passes with it.

Full `ctest` is 342/342 here (pocl CPU device, LLVM 20, aarch64).

bricevideau-ai:
Correction to my previous comment, and a second fix.

**The header was not actually shared.** I said the extension-pointer loading, kernel source and program build were "shared with `test_command_buffer` through the new `command_buffer_common.h`". They were not — the header had exactly one consumer, and `test_command_buffer.c` was byte-identical to main with its own inline copy of all of it. I had reverted that file wholesale while splitting the test out and then described the intent rather than the diff. Sorry.

Now done properly: `test_command_buffer.c` uses the header (−55 lines), so it has two consumers.

I did *not* share the command recording, despite you suggesting it, and want to flag that rather than leave it looking like an oversight. The two tests record materially different buffers — a tiled copy/fill/NDRange/barrier sequence over 16 tiles versus three plain NDRanges — and a recorder parameterised over both would suit neither. Happy to be overruled if you'd rather have it.

**The regression test was also too weak**, which I only found because re-checking the above made me re-run the verification. Detection is probabilistic: the application has to arrive at `clWaitForEvents()` inside the window, and the rate rises steeply with the host work between enqueue and wait. First failure lands around replay 6600 at a spin of 2000, ~900 at 20000, ~390 at 200000. At the 20000 I had originally committed it caught the regression **3 runs out of 6** — it would have passed CI on a buggy build about half the time. Raised to 200000: now 8/8 without the fix, 8/8 passing with it, and the test still runs in 0.36 s.

Full `ctest` 342/342.

bricevideau-ai:
You're right that a delay-based trigger is architecture-dependent, and tuning it here would only have meant it stopped detecting anything on the next machine. Replaced it with something derived from the API instead of from wall-clock.

The loop now polls `clGetEventInfo()` and re-enqueues as soon as the previous submission reads `CL_COMPLETE`. That is also a cleaner statement of the rule under test: simultaneous use is defined against previous submissions "not in the `CL_COMPLETE` state", and this one demonstrably is not. `clGetEventInfo()` takes the event lock and returns `event->status`, so a `CL_COMPLETE` reading proves the runtime has published the terminal status and dropped that lock — on any CPU, with no constant to tune.

It is still probabilistic: what remains is a race against the retire. But the window is now a fixed span of runtime code (between the event unlock and the retire) rather than a wall-clock interval, which is the part that would have varied per machine.

Measured here — 200 replays, first failure between replay 3 and 30:

|  | detection |
|---|---|
| poll, 200 replays | **12/12** without the fix, 8/8 pass with it, 0.06 s |
| tuned spin, 2000 replays | 3/6 without the fix |

So it is both far more reliable and ~30x cheaper. I would still not claim it is guaranteed on every architecture — if you know of a pocl-side hook that would make a race like this deterministic in a test I would rather use that, but I did not want to add a test-only hook to production code uninvited.

Full `ctest` 342/342.

tests/runtime/command_buffer_common.h
```diff
@@ -0,0 +1,103 @@
...
+  *program = clCreateProgramWithSource (context, 1,
+                                        &cmdbuf_vector_addition_src, &length,
+                                        &error);
+  if (error != CL_SUCCESS)
+    return error;
+  error = clBuildProgram (*program, 1, &device, NULL, NULL, NULL);
+  if (error != CL_SUCCESS)
+    return error;
+  *kernel = clCreateKernel (*program, "vector_addition", &error);
+  return error;
```
jansol: These can and should use the same CHECK_CL_ERROR as the test body.

tests/runtime/test_command_buffer.c
```diff
@@ -25,79 +25,32 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "command_buffer_common.h"
 #include "poclu.h"
```
jansol: Redundant since the common header includes this.

tests/runtime/test_command_buffer_replay.c
```diff
@@ -0,0 +1,139 @@
...
+        NULL));
+  CHECK_CL_ERROR (ext.clFinalizeCommandBufferKHR (cmdbuf));
+
+  /* Re-enqueue as soon as the previous submission is observably complete.
+     cl_khr_command_buffer defines simultaneous use against previous
+     submissions "not in the CL_COMPLETE state", so once clGetEventInfo has
+     reported CL_COMPLETE this is not simultaneous use and the next enqueue
+     must succeed. It did not: the buffer used to be retired only AFTER its
+     completion event was published, so an enqueue issued in between was
+     rejected CL_INVALID_OPERATION.
```
jansol: Specifically mention that this is a regression test for a race condition.

jansol:
Actually, reading up on the command buffer spec updates up to version 0.9.8 (which is getting bumped to non-experimental version 1.0 in KhronosGroup/OpenCL-Docs#1614 with no further spec changes), I don't think there is much point in this PR.

More specifically, version spec version 0.9.8 does away with the entire concept of simultaneous use as it was specified previously, and the pending count bookkeeping is no longer necessary (see KhronosGroup/OpenCL-Docs#1411). So instead of this race condition fix, I'd rather have a new PR that updates the code to match the 0.9.8/1.0 specification and removes the pending count tracking. That also eliminates the need for a new test.

Consolidating the command buffer extension setup into a header shared between the tests for command buffers and command buffers with images would still be useful. That change should go into a dedicated separate PR.

jansol:
Opened #2314 as a note about the v1.0 implementation.

PR #2315, opened 2026-09-09
tests: share the command-buffer extension loading between the CB tests

bricevideau-ai:
Split out of #2306 as requested there.

`test_command_buffer` and `test_command_buffer_images` each carried their own copy of the entry-point struct and the ~20 lines of `clGetExtensionFunctionAddressForPlatform` boilerplate, differing only in which subset they declared.

`tests/runtime/command_buffer_common.h` now holds one struct with the union of what the two use, plus a loader that also owns the "not supported, skipping test" 77 return. Each test reads the subset it needs, so neither changes beyond the two lines replacing the block.

No behaviour change. `runtime/test_command_buffer` and `runtime/test_command_buffer_images` pass unchanged; full `ctest` 341/341 here (CPU device, LLVM 20, aarch64).

`test_command_buffer_multi_device` has the same struct and loader, but establishes command-buffer support through a different path before loading, so converting it is a slightly different change — happy to fold it in if you'd prefer it in one go.

### DIFF
--- a/tests/runtime/command_buffer_common.h
+++ b/tests/runtime/command_buffer_common.h
@@ -1,0 +1,85 @@
+/* Shared cl_khr_command_buffer entry point loading for the runtime tests.
+
+   Copyright (c) 2026 PoCL developers
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to
+   deal in the Software without restriction, including without limitation the
+   rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+   sell copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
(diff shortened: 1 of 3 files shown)


jansol:
The multi device test currently skips the check for the base command buffer extension and instead checks support for extensions that build on top of it. If the layered extensions are supported, the base line command buffers are implicitly assumed to be supported as well. But there is no harm in also checking for the base extension like in the other tests, so let's use the shared code there too. The additional checks in the multi-device test can stay where they are, so the change should not be any larger than in the other two tests.

bricevideau-ai:
Done — `test_command_buffer_multi_device` now uses the shared loader too. Same shape as the other two: the struct and the loader block become `struct cmdbuf_ext` plus one `cmdbuf_load_ext` call, its layered-extension checks untouched in place, and the base-extension check simply lands in front of them. `clRemapCommandBufferKHR` joined the shared union since only that test uses it.

All three command-buffer tests pass; full `ctest` 341/341.

jansol (approved):

PR #2319, opened 2026-09-15
CPU kernel library: keep libclc pown when vectorizing builtins

anun333:
## Summary

`llvm.powi` is not accurate enough for `pown`'s 16 ULP bound: CTS `pown` fails at 72-210 ULP typically, up to ~2e6 ULP on subnormal results. No vector math library (glibc libmvec or SLEEF's gnuabi build) provides an accelerated `pown` at all, so swapping it for `llvm.powi` has zero performance upside and a severe correctness downside. This just stops the swap and keeps the libclc scalar implementation, which is correct.

Independently reproduced against current upstream `main` before filing: unpatched, `pown` fails with errors ranging from ~1,700 ULP up to ~6.6e10 ULP across widths and precisions (worse than originally measured, likely a different subnormal case). Confirmed via `nm -D` that neither `libmvec.so` nor SLEEF's `libsleefgnuabi.so` export a `pown` symbol.

## Note on provenance

This is a re-file. It was originally opened as #2307 and closed automatically when the filing GitHub account was deleted (shows as "closed by ghost" in the timeline) — not a maintainer decision on the content. Rebased cleanly onto current `main` with no conflicts before re-filing.


anun333:
Additional evidence found while testing this against narrower kernel-library variants, which I think strengthens the case here beyond what the description currently claims.

Building with `-DKERNELLIB_HOST_CPU_VARIANTS=distro` and selecting the SSE2 variant at runtime (`POCL_KERNELLIB_NAME=sse2`), **unpatched `main` fails `pown` badly**:

| function | width | max ULP vs bound 16 |
|---|---|---|
| `pown` float | 1 | 1,490,595 |
| `pown` float | 3 | 836,118 |
| `pown` float | 4 | 762,628 |
| `pown` float | 8 | 544,782 |
| `pown` double | 4 | 2,205,099,412 |

With this patch applied, all of those pass. (`rootn` double at width 1 returns `inf` both before and after — that one is separate and not addressed here.)

So this isn't only a CTS-bound question on the default build: on a distro-style multi-variant build the current `llvm.powi` swap produces results wrong by six to nine orders of magnitude, and keeping libclc `pown` fixes it.

Measured with an independent ULP harness (PyOpenCL driving real kernels, mpmath references), glibc 2.39 / LLVM 22 / znver3 host.

anun333:
Adding the configuration boundary, which I left out of the description and should not have — it changes how this should be prioritised.

The swap only happens under `ENABLE_HOST_CPU_VECTORIZE_BUILTINS`, which removes `libclc-pocl/pown.cl` from `SOURCES_WITH_SLEEF` (`lib/kernel/host/CMakeLists.txt`, the `list(REMOVE_ITEM ...)` block). `pown` then resolves to a Clang builtin and expands to `llvm.powi`. That option carries its own `# implies CONFORMANCE=OFF` comment, and it is off by default.

So, stated plainly:

* a **default** build is unaffected — `pown` keeps the libclc implementation;
* a **conformance** build is unaffected, and cannot enable the option;
* the failing configuration is `ENABLE_HOST_CPU_VECTORIZE_BUILTINS=ON`, which is already non-conformant for other reasons.

The ULP figures in the description are real and were measured, but quoting "up to ~6.6e10 ULP" without that boundary reads as though shipped builds are affected, and they are not. My apologies for the ambiguity.

What that boundary does *not* change is the argument for the patch: no vector math library provides an accelerated `pown` at all — no row in LLVM's `VecFuncs.def` for glibc libmvec or SLEEF-gnuabi — so removing `pown.cl` buys no vectorization and costs correctness even in a configuration that has already opted out of conformance. The patch is two lines and touches nothing a default build compiles.

*AI tool usage: this analysis was prepared with AI assistance; all results above were run and checked.*

franz:
> no vector math library provides an accelerated pown at all — no row in LLVM's VecFuncs.def for glibc libmvec or SLEEF-gnuabi

Actually, I think Intel SVML provides accelerated `pown` variants. At least I can see it in the list of symbols (`nm libsvml.a | grep svml_pown`). But you're right that glibc's libmvec does not provide it.

anun333:
Thanks for both points; the branch has changed since, so taking them against the current version.

**SVML.** You're right that `libsvml` exports `svml_pown`, and "no vector math library provides it" was wrong as written. But I don't think an `if(NOT ENABLE_HOST_CPU_VECTORIZE_SVML)` would change behaviour, because LLVM has no route to it: `VecFuncs.def` has no `powi`/`pown` row for any library (SVML's pow rows are all the two-float `pow`), and a per-lane `pown(x, n)` needs a `(<N x float>, <N x i32>)` mapping that none of the VFABI parameter kinds in the table express. `llvm.powi`'s own vector form keeps a scalar exponent. So an SVML build would still emit `llvm.powi` and never call `svml_pown`. If there's a mechanism I've missed I'm happy to add the condition.

**CI.** The branch now keeps `llvm.powi` for literal exponents (the `__builtin_constant_p` fast path above), and `vecbuiltin.cl` calls `pown(a[gid], 4)`, so `CHECK: llvm.powi.v` holds as is and I've left the test alone. Happy to drop the line anyway if you'd rather not have CI depend on that fold.

**Scope change.** Double now stays on the builtin, because routing it to libclc hits llvm/llvm-project#224127 (details and CTS numbers in the updated description).

PR #2311, opened 2026-09-06
CPU: per-function deny list for vectorized math, library-driven builtin swap, extended x86 libmvec table

ghost:
Stacked on #2309 (the trig lane fix; the regression test added here needs it on conformant builds), so the first commit of this PR is that fix and the 19 commits after it are this series. Design discussion: #2310.

Summary: PoCL's CPU device hands LLVM a per-library vector-function table filtered by a measured deny list, derives the Clang-builtin swap from what that library can actually vectorize, allows libmvec and SLEEF vector math under `ENABLE_CONFORMANCE` (opt-in), and extends the x86 libmvec table to the 26 functions glibc 2.35 and later export. On this host the default non-conformance build goes from 101 to 103 of 103 on the CTS `math_brute_force` sweep, and the conformant build with libmvec on passes the full CTS main suite (161 tests, non-wimpy math).

## What it does

1. **Filtered vector library table.** PoCL hands LLVM a
   `TargetLibraryInfoImpl` built from LLVM's own `VecFuncs.def` rows for
   the selected library, minus functions measured to exceed the OpenCL C
   ULP bounds (`lib/CL/pocl_vecmath_deny.h`, generated by CMake from one
   list), minus `POCL_VECMATH_DENY`, plus `POCL_VECMATH_ALLOW`. The table
   is registered both at codegen (`initPassManagerForCodeGen`) and in the
   pipeline that runs the loop vectorizer (`populateModulePM`), because
   `InjectTLIMappings` and `LoopVectorize` only see it there. Denied
   functions keep their scalar implementation; nothing else changes. The
   deny table follows the library that actually answers (on x86, SLEEF's
   gnuabi build is reached through the libmvec table, so
   `HOST_CPU_LIBMVEC_LIBRARY` decides).
2. **Library-driven builtin swap.** The Clang-builtin swap in
   `lib/kernel/host/CMakeLists.txt` is computed: exact operations always
   (they become native intrinsics), transcendentals only when the selected
   library has a row for them and the function is not denied. A swapped
   builtin's scalar libm fallback has to be within bounds too, because
   kernels on explicit vector types do not loop-vectorize and call it. A
   denied function must leave the swap: a scalar libm call inside the work-item
   loop makes `LoopVectorize` reject the whole loop ("call instruction
   cannot be vectorized"), so every kernel calling it would run scalar,
   which is worse than the inlined SLEEF source. The CORE-Math FP16
   selection is derived from the same list (Clang has native half builtins
[...]

Issue #600, opened 2018-01-05
pocl 1.0 fails to build on armv7 when CUDA is in use


flaub:
I'm attempting to build against CUDA 6.5 at the moment, since that was the easiest (only?) version of CUDA I can find for ARMv7. I've looked at the compiler error and it appears that one issue might be related to CUDA 6.5, but the other looks like a real build issue for ARM+CUDA. Here's the error:

```
[ 80%] Building C object lib/CL/devices/cuda/CMakeFiles/pocl-devices-cuda.dir/pocl-cuda.c.o
In file included from /home/flaub/src/pocl/pocl-1.0/lib/CL/./pocl_cl.h:47:0,
                 from /home/flaub/src/pocl/pocl-1.0/lib/CL/devices/common.h:28,
                 from /home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c:28:
/home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c: In function ‘pocl_cuda_submit_kernel’:
/home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c:1002:17: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘size_t {aka unsigned int}’ [-Wformat=]
     POCL_ABORT ("[CUDA] Total constant buffer size %u exceeds %lu allocated\n",
                 ^
/home/flaub/src/pocl/pocl-1.0/lib/CL/./pocl_debug.h:97:25: note: in definition of macro ‘POCL_ABORT’
         fprintf(stderr, __VA_ARGS__);                                   \
                         ^
/home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c: In function ‘pocl_cuda_submit_node’:
/home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c:1104:16: warning: implicit declaration of function ‘cuStreamWaitValue32’ [-Wimplicit-function-declaration]
       result = cuStreamWaitValue32 (stream, dev_ext_event_flag, 1,
                ^
/home/flaub/src/pocl/pocl-1.0/lib/CL/devices/cuda/pocl-cuda.c:1105:37: error: ‘CU_STREAM_WAIT_VALUE_GEQ’ undeclared (first use in this function)
                                     CU_STREAM_WAIT_VALUE_GEQ);
[...]

PR #2320, opened 2026-09-15
CPU: per-function deny list for vectorized math, library-driven builtin swap, extended x86 libmvec table

anun333:
## Summary

Upstream's vectorized-math builtin selection is all-or-nothing: `ENABLE_HOST_CPU_VECTORIZE_{LIBMVEC,SLEEF,SVML}` hands LLVM a whole library, and `ENABLE_CONFORMANCE` forbids all of them because per-function ULP was never verified. This series measures every candidate function against the OpenCL CTS and denies only the ones that actually fail, instead of banning the whole library:

**Four** libmvec functions are denied — `log`, `cbrt`, `exp`, `log1p`. SLEEF has a separate list: `asinh` and `cbrt` unconditionally, `log1p` below 3.9.0, and `pow` below 3.8. The measured reason for each, with the OpenCL bound it exceeds:

- **Float `log`**: 3.009 ULP vs. bound 3, confirmed at CTS width 1.
- **Double `exp`**: denies libmvec (glibc 2.39) at one specific bit pattern, 3.0077 ULP vs. bound 3 at input `-0x1.31000000000d0p-6` (observed `0x1.f68e9226f16cep-1` against a correct `0x1.f68e9226f16d1p-1`; scalar glibc `exp` is 0.0077 ULP there) — a rare tail case the full CTS density catches but sampled sweeps miss (20,000-sample direct check found nothing else over 3, max 2.63). glibc documents up to 4 ULP for libmvec; OpenCL requires 3.
- **`log1p`** (both precisions): same shape of bug, same root cause (glibc's own tolerance exceeds OpenCL's). Its bound is 2 ULP, not 3: float 2.004 ULP at `0x1.55a146p-2` (width 4), double 2.524 ULP at `0x1.0000000000201p-11`. Denied, falls back to SLEEF.
- **`cbrt`** (both precisions): a different mechanism — the *vectorized* cbrt is fine; it's the **scalar fallback** (used whenever a kernel has explicit vector types, which don't auto-vectorize) that's out of bound. Established the rule that a swapped builtin's scalar fallback needs to be in-bound too, not just its vector path.
- **SLEEF's own denials**: `asinh` returns `inf` for `|x| > sqrt(MAX)` in both precisions and in every version measured; `cbrt` is the scalar-fallback case above; `log1p` overflows near MAX below 3.9.0; and `pow(-DBL_MAX, 1)` returns `-inf` below 3.8 (SLEEF issue #600). The last two are gated on the detected SLEEF version. The first two are not, so a 3.9 build still denies `asinh` and `cbrt`.
- Extends `VecFuncs.def`'s x86 libmvec table for glibc >= 2.35 (adds rows LLVM's own table lacks), gated to systems new enough to have them.
- Denying a function in any precision now pulls it out of the swap entirely in both, because a scalar call with no vector-library mapping otherwise makes the loop vectorizer reject the whole work-item loop, silently de-optimizing every kernel that calls it.
- Includes the fix now filed as **#2327** (float sin/cos/tan lane cross-talk in the vector argument reduction) as a prerequisite — this branch is stacked on it. It was originally #2309, which is closed and shows a deleted author; #2327 is the live filing and the one to review.

## Independent re-verification before filing

- Rebased cleanly onto current `main`, zero conflicts.
- PoCL's own regression suite: 146/146 pass on the rebased tree, including `test_vecmath_ulp_{cbs,loopvec}` and `test_trig_lane_mix_{cbs,loopvec}`.
- Independently rebuilt ULP-sweep harness (PyOpenCL + mpmath references, not reusing any original test code): 250/250 rows clean across 25 functions x widths 1/2/3/4/8 x float/double. **This is a sample, and it has since been shown to be one**: a full `math_brute_force` run on a SLEEF-backed build found three further functions over bound — `acosh`, `cosh`, `sinh` — that the sweep called clean. See the comments below.
- Reproduced the exact cited exp bug bit-for-bit against unpatched `main` (`0x1.f68e9226f16cep-1`, 3.0077 ULP) once built with the same work-group configuration as the original discovery.
- Reproduced the exact cited trig lane cross-talk bug digit-for-digit against unpatched `main` (`sin=0.0078124204`) once built with `-DENABLE_CONFORMANCE=ON`, matching the original discovery's build variant.

## Note on provenance

This is a re-file. It was originally opened as #2311 (21 commits; this branch has since grown to 28) and closed automatically when the filing GitHub account was deleted (shows as "closed by ghost" in the timeline) — not a maintainer decision on the content. The original full CTS run on this series passed 159/161 (the 2 failures are what commits in this series deny).


anun333:
Cross-reference, in case it's useful for reviewing the extended-table commit in this series:

The commit here that maintains a local extended x86 libmvec table exists because `VecFuncs.def` upstream doesn't list the functions glibc 2.35+ actually exports. I've now filed llvm/llvm-project#223817 to close part of that gap upstream (12 functions × both precisions × the `b`/`d` ABI classes, referencing llvm/llvm-project#204678).

If that lands, the local table here shrinks on its own — the `add only the libmvec rows LLVM's own table lacks` commit already tracks which `(name, VF)` pairs LLVM covers and skips them, so no change would be needed on this side.

Not a dependency in either direction; just so the duplication is visible and intentional rather than looking like drift.

anun333:
Self-reporting a regression I found in this series while testing it against
narrower kernel-library variants. A default `native` build is unaffected.
`-DKERNELLIB_HOST_CPU_VARIANTS=distro` builds are affected, which is what
distributions ship.

## Symptom

Host: Zen 3 (AVX2, no AVX-512). `distro` build, SSE-class variant selected
via `POCL_KERNELLIB_NAME=sse2`. `sin`/`cos`/`tan` return wrong results at
vector widths above 128 bits (`float8`, `double4`, `double8`); widths that
fit an SSE register are correct. Unpatched `main` is correct at all widths.

## Root cause

**The vector-library call that gets emitted does not depend on the
kernel-library variant.** For this kernel:

```c
__kernel void k(__global const double* i, __global double* o) {
  int x = get_global_id(0);
  double8 v = vload8(x, i);
  vstore8(sin(v), x, o);
}
```

the undefined symbols in the generated `k.so.o` are:

| build | variant | emitted call | correct? |
|---|---|---|---|
| this series | `sse2` | `_ZGVdN4v_sin` | **no** |
| this series | `avx2` | `_ZGVdN4v_sin` | yes |
| unpatched `main` | `sse2` | *(none)* | yes |

Identical symbol for both variants. `_ZGVdN4v_sin` is the **AVX2** 4-wide
double entry point, and per the x86-64 vector ABI it takes its argument in
a 256-bit YMM register.

But the object compiled for the `sse2` variant has no YMM registers at all:

```
$ objdump -d k.so.o | grep -c ymm
0
$ objdump -d k.so.o | grep -c xmm
16
```

So the argument is passed in XMM registers while the callee reads YMM0.
The call does not fault — the host has AVX2, so libmvec's instruction
stream is legal — it simply reads the wrong register. Hence wrong results
rather than a crash.

Unpatched `main` emits no vector-library call for this kernel, which is
why it is correct: it uses the variant's own kernel-library implementation.

Denying the functions at runtime on the same build restores correct results,
confirming the swap is the path:

```
POCL_VECMATH_DENY=sin,sinf,cos,cosf,tan,tanf,llvm.sin.f32,llvm.sin.f64,\
llvm.cos.f32,llvm.cos.f64,llvm.tan.f32,llvm.tan.f64
```

## Where it enters

Bisect at the `sse2` variant, sin/cos/tan, all widths, float and double:

| commit | result |
|---|---|
| unpatched `main` | correct |
| trig fix alone | correct |
| `bd2b5cf23` vector library table in the IR pipeline | correct |
| `c580a5dbb` register the table in `populateModulePM` too | **SIGILL** |
| `e3df253b0` | **SIGILL** |
| `9340e4f9f` AVX-512 rows only on AVX-512 hosts | wrong results |
| series HEAD | wrong results |

The SIGILL at `c580a5dbb` is the same class of bug one ISA level up: that
commit makes the `_ZGVe` (AVX-512) rows reachable, and on a host without
AVX-512 the emitted call really is illegal — exactly as `9340e4f9f`'s
message anticipates. `9340e4f9f` fixes the crash by dropping those rows:

```c
__builtin_cpu_init();
HaveAVX512 = __builtin_cpu_supports("avx512f");
```

> The CPU device compiles for the host, so ask the host.

That assumption holds for a `native` build. It does not hold for a
`distro` build, where the selected variant can be narrower than the host.
With the AVX-512 rows gone, the vectorizer falls to the AVX2 rows — which
the host supports and the SSE2 variant cannot satisfy.

**The submitted state does not crash.** I verified HEAD and unpatched
`main` both run cleanly with an AVX-512 variant forced on this host. The
SIGILL is confined to intermediate commits, so the series is not
bisect-clean, but its final state is crash-free.

## Suggested direction

`9340e4f9f` already has the right mechanism and asks the wrong question.
Filtering rows against the **target kernel-library variant's** feature
level rather than the host's — and extending the same ladder one rung down,
so `_ZGVd` is dropped for SSE-class variants exactly as `_ZGVe` is dropped
today — would cover both cases with one rule. That is contained to the same
function, so it belongs in this PR rather than a follow-up.

Marked draft so it is not sitting in the review queue while this is open.

Measured with an independent ULP harness (PyOpenCL driving real kernels,
mpmath references), glibc 2.39 / LLVM 22 / znver3. Happy to share it or cut
a smaller standalone reproducer.

anun333:
Fixed in 72d213658, and taking this out of draft.

## The fix

Filter the vector-library table on the **compilation target** rather than the host, one rule per ISA class — which subsumes the AVX-512 case that `9340e4f9f` was handling rather than replacing it:

| class | ISA | width | dropped unless target has |
|---|---|---|---|
| `_ZGVb` | SSE2 | 128-bit | *(baseline on x86-64)* |
| `_ZGVc` | AVX | 256-bit | `avx` |
| `_ZGVd` | AVX2 | 256-bit | `avx2` |
| `_ZGVe` | AVX-512 | 512-bit | `avx512f` |

The target CPU is `device->llvm_cpu`, which `pocl_get_distro_cpu_name()` already derives from the selected kernel-library variant — so the value was available, it just never reached the filter. Threaded to all three `createFilteredTLII` call sites; `initPassManagerForCodeGen` gained an `MCPU` parameter that both its callers already had in scope. Feature queries use `llvm::X86::getFeaturesForCPU` rather than `__builtin_cpu_supports`.

## A second host-dependent check, worth flagging separately

While fixing the above I found the row *builder* also called `hostHasAVX512()`. That function is memoised once per process and its table is shared across devices, so it could bake one device's target capabilities into a table another device then uses — a latent problem independent of the variant issue, and one that would affect any process serving two CPU devices with different targets.

It now emits every ISA variant glibc exports unconditionally and leaves all per-target filtering to `createFilteredTLII`, which is the only place that knows the target. No host-based capability checks remain in this path.

## Verification

| check | before | after |
|---|---|---|
| `sse2` / `ssse3` / `sse41` variants, sin/cos/tan all widths | 21/30 | **30/30** |
| `avx` / `avx2` variants | 30/30 | 30/30 |
| emitted symbol, `sse2` variant, `double8 sin` | `_ZGVdN4v_sin` | **`_ZGVbN2v_sin`** |
| auto-selected variant, 250-row ULP sweep | 250/250 | 250/250 |
| `ctest -R 'vecmath\|trig\|regression'` | 146/146 | 146/146 |

The last two lines are the ones I cared most about — the fix should not change behaviour on a `native` build, which is the common case, and the vecmath FileCheck tests assert on emitted call names so they would catch an over-broad filter.

anun333:
The full CTS run this PR was waiting on is done. `a2b39a2fc` asked for exactly this — *"Keep them default-off under ENABLE_CONFORMANCE until a full CTS run with them is on record"* — so here it is, including the part that argues against flipping anything yet.

### Result

**150 of 153 units exit 0.** 105/105 `math_brute_force` sub-tests, 48/48 serial suites. 12h15m wall.

Build: `ENABLE_CONFORMANCE=ON`, `ENABLE_HOST_CPU_VECTORIZE_LIBMVEC=ON`, SLEEF off, native target, glibc 2.39, LLVM 22, OpenCL-CTS at `1a6f465`. `math_brute_force` run one sub-test per process with `--num-worker-threads 1`; the serial suites got the whole device.

Three failures, and **none of them is caused by this branch**. Each was re-run against a build of the merge-base `aaaa2d67b` with no patches, and each reproduces there identically:

| unit | failure | filed |
|---|---|---|
| `math_edge_cases` | 19 signed-zero cases in `cospi`/`sinpi`/`tanpi`/`atanpi` | #2322 (fix, 19 → 0) |
| `math_tgamma` | 16.081993 / −16.744629 ULP vs a 16 ULP bound | #2325 |
| `suite_basic` | `bufferreadwriterect`, 1 of 106 | #2326 |

So on this target the deny list does what it was built to do: every function that reaches libmvec is inside its OpenCL bound, and the four denied functions stay on their scalar paths.

### The part that matters more

**That run was on a native target, and the result does not carry over to baseline targets.** Re-measuring at `-DLLC_HOST_CPU=x86-64` turned up #2324: double `powr` returns `inf` for `2^3` once the work-item loop vectorizes, with libmvec on. It reproduces on unpatched `aaaa2d67b` too, so it is not this branch's doing — but this branch is what makes that configuration reachable under `ENABLE_CONFORMANCE`, so it is this branch's problem to disclose.

Concretely, I would not flip any default until #2324 is understood, and probably not for distro/baseline builds even then without a CTS run at that target.

### One thing I would change about my own reasoning here

Every ULP figure in the deny list, including the provenance note in `CMakeLists.txt`, was measured on one machine (znver3, with FMA). 66 files under `lib/kernel/` branch on `HAVE_FMA32`/`HAVE_FMA64`, so a no-FMA baseline target compiles different code with different rounding, and libmvec's `b`-class implementations are not the `d`-class ones that were measured. The numbers are evidence about one microarchitecture presented as though general.

The re-measurement at baseline came out clean for the deny list itself — 247 of 250 harness cells inside bound, and the 3 outliers were `powr`/`pown`/`rootn` tripping over #2324 rather than an accuracy problem. But the deny list ought to say which ISA level it was measured at, and ideally be re-derivable rather than transcribed. Happy to add that if the approach here is otherwise wanted.

anun333:
Pushed three commits (6b5106205, 3e13df239, 987fd98a7) that correct the SLEEF-as-libmvec path in this series. Summary, so the review has the context:

**What was wrong.** For a SLEEF-named `LIBMVEC`, the series only used LLVM's own libmvec rows and skipped PoCL's extended x86 rows, with the comment "the extended rows only exist in glibc". That was mine and it was false: SLEEF's GNU-ABI build exports the whole set (`_ZGVdN4v_acos` … `_ZGVdN4vv_hypot`, `cbrt`, `erf`, `log1p`, `tgamma`, …) in all four ISA classes, in 3.5.1 and 3.9.0 alike — 725 entry points each. The same gate was in the kernel library's Clang-builtin swap (keyed on `HOST_CPU_GLIBC_VERSION_NUM`, which is only computed for glibc). Net effect: with SLEEF, only `sin cos tan exp log pow` were vectorized; the other 19 stayed on the inlined kernel library while looking like a SLEEF measurement.

**The fix.** Extended rows and the swap now apply when `LIBMVEC_IS_SLEEF` too (SLEEF has no version gate to honour). A probe kernel calling the twelve extended functions now emits all 24 `_ZGV*` calls on both SLEEF versions; before, zero.

**The deny list, now measured.** Full ULP sweep (25 functions × 2 precisions × widths 1,2,3,4,8) with SLEEF actually answering, plus direct calls into the vector entry points:

- `asinh`: returns inf for |x| > √MAX (1.34e154 double, 1.84e19 float), both versions; true value ≈ 355 / 44 there.
- `log1p`: returns inf for x > ~1e307 (double) / ~3e38 (float) in 3.5.1; correct in 3.9.0. Gated below 3.9.0 (no version in between measured).
- `cbrt`: double 2.336 ulp vs 2 on explicit 3-wide types, the scalar-glibc fallback, same mechanism as the libmvec entry.
- `pow`: the existing <3.8 entry (issue #600), reproduced on 3.5.1 (`_ZGVdN4vv_pow(-DBL_MAX, 1) = -inf`), correct in 3.9.0.

With the list applied both cells sweep 250/250. Where both libraries pass, SLEEF is tighter than glibc at double (`sin` 0.62 vs 2.15 ulp, `cos` 0.67 vs 2.41, `tan` 0.59 vs 1.77).

A conformance CTS run on the 3.5.1 cell (Ubuntu noble's `libsleefgnuabi.so.3.5`) is queued behind the current baseline run; I'll report it here.

anun333:
Splitting this series so it can be reviewed in pieces. First piece out: **#2327**, the libclc per-lane trig reduction fix and its regression test (3 files). It is the oldest commit here and is unrelated to the vector-math work, so it is now on its own.

For anyone reviewing the rest, the dependency structure, checked by cherry-picking each candidate onto current main:

- **standalone, applies and stands on its own:** the libclc trig fix (now #2327).
- **standalone patch, but inert on its own:** `linker: move the no-builtins restriction from functions to call sites` is guarded by `ENABLE_HOST_CPU_VECTORIZE_BUILTINS` and references the swap list this series introduces, so it would be dead code without the rest. It stays in the stack.
- **ordered, cannot be reordered:** the deny-list entries conflict if taken before the commit that moves the lists into CMake, and the conformance opt-in conflicts if taken before the CMake detection it builds on.

So the remainder is a stack rather than independent parts, and I will re-file it in dependency order rather than as one 20-file PR. I will post the order here before pushing anything, so review effort is not spent twice.

anun333:
I have corrected the description. It had drifted from the branch in four ways, and one of them understated the scope of the change — flagging it rather than editing quietly.

* **The deny list is four libmvec functions, not three.** `log` is denied (float 3.009 ULP vs. bound 3, confirmed at CTS width 1) and the summary did not mention it. The later CTS comment on this PR says "the four denied functions", so the code and the evidence were consistent; only the description was behind. `log` is common enough that anyone weighing the risk of this series should have seen it named.
* **`pow` on SLEEF older than 3.8 is also denied** and went unmentioned — a different mechanism, `pow(-DBL_MAX, 1)` returning `-inf` (SLEEF issue #600). It is gated on the detected SLEEF version, so a 3.8-or-newer build denies nothing.
* **The prerequisite is #2327, not #2309.** #2309 is closed and shows a deleted author; #2327 is the live re-file of the same fix. The description pointed at the dead one.
* **`log1p`'s bound is 2 ULP, not 3**, which the description implied by grouping it with `exp`. The measured figures are float 2.004 ULP at `0x1.55a146p-2` (width 4) and double 2.524 ULP at `0x1.0000000000201p-11`.

One clarification rather than a correction: the `exp` figure was given as `0x1.f68e9226f16cep-1`, which is the **observed result**, not the input. The input is `-0x1.31000000000d0p-6`; the correct result there is `0x1.f68e9226f16d1p-1`, so the observed value is three units of last place low, which is the 3.0077 ULP. Scalar glibc `exp` is 0.0077 ULP at the same point — the error is specific to the vector variants.

None of the measurements changed. The commit count in the provenance note is also updated: #2311 had 21 commits and this branch has since grown to 28.

*AI tool usage: this analysis was prepared with AI assistance; all results above were run and checked.*

anun333:
The SLEEF deny list in this series is short by three functions. I found it two ways on the same day, and neither result depends on the other.

**From the conformance suite.** A full `math_brute_force` run on a SLEEF-backed build at this branch's head fails `math_acosh`, `math_cosh` and `math_sinh`. None of the three is on the deny list.

**From a re-measurement.** A script that re-derives the list on the machine it runs on — against the suite's own bounds from `function_list.cpp`, and using `Ulp_Error`'s definition including the tighter unit at a binade boundary — finds exactly the same three, and locates the inputs:

| function | precision | ULP | bound | worst input | 3.5.1 | 3.9.0 |
|---|---|---:|---:|---|---|---|
| `acosh` | double | **inf** | 4 | `0x1.181fabafd6156p+512` | fails | fails |
| `acosh` | float | **8.92e43** | 4 | `0x1.0bc008p+64` | fails | fails |
| `cosh` | float | **6657617.18** | 4 | `0x1.63a44ap+6` | fails | fails |
| `sinh` | float | **6657617.18** | 4 | `0x1.63a44ap+6` | fails | fails |
| `tanh`, `atanh` | both | ≤ 1.0005 | 5 | — | pass | pass |

The failing figures are identical on 3.5.1 and 3.9.0 to every digit, so these are not version regressions, and the entries should be unconditional as `asinh` and `cbrt` already are, rather than version-gated like `log1p` and `pow`. All three are confirmed routed to the vector library on both builds, so this measures SLEEF and not the inlined kernel-library fallback.

### Two mechanisms, and one of them is already in the list

**`acosh` fails exactly where `asinh` does.** `acosh(x)` is `log(x + sqrt(x*x - 1))`, and `x*x` overflows past `sqrt(MAX)`. That is the same sentence this series already writes about `asinh`, whose deny note reads "returns inf for |x| > sqrt(MAX) in both precisions". The list caught one sibling and missed the other.

**`cosh` and `sinh` give up just inside the float32 range.** At `0x1.63a44ap+6` ≈ 88.82 the true value is ≈ 1.9e38, comfortably below `FLT_MAX` ≈ 3.4e38, and both return infinity. The suite forgives premature overflow for `float` *within the allowed error*; this is six million ULP outside it.

### Suggested change

```cmake
-set(POCL_VECMATH_DENY_SLEEF_ENTRIES "asinhf;asinh;cbrtf;cbrt")
-set(POCL_VECMATH_DENY_SLEEF_FUNCS "asinh;cbrt")
+set(POCL_VECMATH_DENY_SLEEF_ENTRIES
+    "asinhf;asinh;acoshf;acosh;coshf;cosh;sinhf;sinh;cbrtf;cbrt")
+set(POCL_VECMATH_DENY_SLEEF_FUNCS "asinh;acosh;cosh;sinh;cbrt")
```

with the note extended to record the mechanism and the inputs.

This is the same treatment PoCL already applies to an inaccurate vector function: 401c005d6 worked around an fp32 SLEEF `ldexp` that was over half a ULP by computing that one function differently, with 0d3965332 as the follow-up. A deny entry is that decision expressed per function rather than per file, and it leaves the accurate inlined implementation in place.

### How it was measured

One sub-test per process, `ENABLE_CONFORMANCE=ON`, vectorized math on through SLEEF, glibc 2.39, LLVM 22.1.8, x86-64. The library that actually mapped is read from `/proc/self/maps` rather than from the configure-time value, because 3.5 and 3.9 share the `libsleefgnuabi.so.3` soname: `libsleefgnuabi.so.3.5` and `libsleefgnuabi.so.3.9` respectively.

The wider point, and the reason I went looking: the previous SLEEF list came from a 250-point sweep that reported the 3.9.0 cell clean. It was a sample, and the list built from it was a sample too. `acosh` was missed even though its mechanism was already written down one line above, for `asinh`.

anun333:
The earlier run I posted here was `math_brute_force` only, 105 of 105. This
is the full suite, 153 of 153, on the SLEEF configuration at the current
head — the first complete run this branch has on record.

**153 of 153 units. Six failures, and they are four different things:**

| unit | whose | note |
|---|---|---|
| `math_acosh`, `math_cosh`, `math_sinh` | **this branch's deny list** | The list is short by the hyperbolic family. Measured, with mechanisms and inputs, in the comment above |
| `math_tgamma` | PoCL's, pre-existing | #2325, and known here for a while — the test post-dates the conformance version PoCL reports |
| `math_math_edge_cases` | PoCL's, pre-existing | fails on libmvec cells too, so unrelated to SLEEF and to this branch |
| `suite_basic` | **not PoCL's** | `test_bufferreadwriterect.cpp:220`, 1 of 106 — KhronosGroup/OpenCL-CTS#2765 |

Build: `ENABLE_CONFORMANCE=ON`, vectorized math on through SLEEF
(`libsleefgnuabi.so.3.5`, read from `/proc/self/maps` rather than from the
configure value, since 3.5 and 3.9 share the soname), glibc 2.39, LLVM
22.1.8, x86-64.

### What I think this does and does not establish

It is the run the opt-in was gated on: vectorized math under
`ENABLE_CONFORMANCE`, full suite, on record. The record is honest rather
than clean — one family of failures is this branch's to fix and the other
three items are pre-existing or upstream's.

It does **not** cover libmvec. The 105-unit run above was that
configuration for the math units only, so a full-suite libmvec run at this
head still does not exist, and I would not argue for any default change
without one.

anun333:
I have corrected the description again, rather than editing it quietly. Three changes, and two of them were plainly wrong:

- **"a 3.8 or newer build denies nothing"** — false. `asinh` and `cbrt` are denied for SLEEF *unconditionally*; only `log1p` (below 3.9.0) and `pow` (below 3.8) are version-gated. A 3.9 build still denies two functions.
- **The deny list summary named only the libmvec four plus `pow`**, which reads as though SLEEF denies just `pow`. SLEEF has its own list, and it always did.
- **The "250/250 rows clean" line now says it is a sample.** It is, and the comment above this one is the proof: a full `math_brute_force` run on a SLEEF-backed build fails `acosh`, `cosh` and `sinh`, none of which the sweep flagged. Offering a clean sample as evidence of completeness was the weakest thing in the description, and it happens to be the thing this series is about.

Nothing in the branch changed; only the prose describing it.

anun333:
@franz, would this be easier to review as a stack? It's 28 commits over 20 files, and the smaller PRs have been going through much faster. I can re-send it as five PRs, opening each one only after the previous one merges:

1. **The mechanism:** the vector-library table filtered per function, the deny-list header, and the ULP regression test.
2. **The kernel-library builtin swap,** with the linker change that depends on it, the `frexp` exclusion (llvm/llvm-project#224127), and the core-math half-precision follow-ups.
3. **The extended x86 libmvec rows,** filtered by compilation target, with the glibc and SLEEF detection.
4. **The measured deny-list entries,** each with its measurement in the commit message. This includes `acosh`, `cosh` and `sinh` for SLEEF from my comment above, which this branch doesn't have yet, so as it stands it fails those three math tests on a SLEEF build.
5. **The conformance opt-in, docs and FileCheck tests.**

#2327 is already split out this way. If you'd rather keep it as one PR, I'll push the three missing deny entries here instead. Either way I'll leave this open until whatever replaces it covers it, and won't add more here in the meantime.

anun333:
@franz, a gentle ping on the question above. The first piece is ready and stands on its own: the per-function deny mechanism and its ULP regression test, 4 files. With it, all 139 regression tests pass, and float `log` with libmvec goes from 3.8 ULP (bound 3) to 0.76. If splitting works for you, I'll close this PR and open that piece in its place, so the number of open PRs doesn't grow.

PR #2322, opened 2026-09-16
libclc: fix the sign of exact zeros in sinpi, cospi, tanpi and atanpi

anun333:
`math_edge_cases` (added to the CTS in May, KhronosGroup/OpenCL-CTS#2665) reports 19 signed-zero failures on the CPU device. This fixes all of them. Not a regression — a new test finding old behaviour.

Four separate causes:

**`sinpi`, `cospi`, `tanpi`** derive the result's sign from the parity of the integral part of `x`. That is right for every non-zero result and wrong when the result is an exact zero, because the parity term then lands on the zero itself:

```
sinpi(1.0)  gave -0.0, must be +0.0   (zero takes the sign of x)
cospi(0.5)  gave -0.0, must be +0.0   (+0.0 at every half-integer)
tanpi(1.0)  gave +0.0, must be -0.0   (sign of x flipped by parity)
```

`tanpi` is the clearest of the three: the `>= 2^23` and `>= 2^52` paths in the same function already return `xodd` for integral arguments, so this only makes the small-argument path agree with them.

**`atanpi`** (doubles only) took its sign with `return x == v ? z : -z;` where `v` is `fabs(x)`. For `x = -0.0` that test is true, since `-0.0 == +0.0`, so the positive branch was taken. Now takes the sign bitwise, which is what `atanpi_fp32.cl` already does. The float version already passed.

### Verification

x86-64 host CPU device, `ENABLE_CONFORMANCE=ON`:

- `math_edge_cases`: **19 failures → 0, PASSED**
- full brute-force `sinpi`, `cospi`, `tanpi`, `atanpi`: all still pass (4/4, 4/4, 3/3, 3/3), so no non-zero result moved
- checked `atan`, `asin`, `asinpi`, `tan`, `sinh`, `atanh`, `erf` at `-0.0`: all already correct

`libclc/atan_fp64.cl` carries the same `x == v` line, but the host CPU kernel library builds `atan` from `sleef-pocl/atan.cl`, so that copy is unreachable there — left alone rather than changed blind.

*AI tool usage: noted in the commit messages (`Assisted-by:`), per the AI Tool Use Policy.*


anun333:
Confirmed on the configuration distributions actually ship, which the description did not cover — it was verified on a `-DLLC_HOST_CPU=x86-64` build.

Unpatched `aaaa2d67b`, `-DENABLE_CONFORMANCE=ON -DKERNELLIB_HOST_CPU_VARIANTS=distro`, with `POCL_KERNELLIB_NAME=sse2` so a pre-Haswell variant is selected the way a user's CPU would select it. No vector math library, no patches:

```
math_edge_cases: 19 FAIL lines
  tanpi   8
  cospi   6
  sinpi   4
  atanpi  1
```

**Nineteen, in exactly the four functions this PR touches**, matching the count in the description. Sample:

```
FAIL: tanpi(0xbff0000000000000) - expected 0x0000000000000000, got 0x8000000000000000
FAIL: tanpi(0x4008000000000000) - expected 0x8000000000000000, got 0x0000000000000000
```

For contrast, the same test on a `-DLLC_HOST_CPU=x86-64` build of a different branch also gives **19**, with the same breakdown. So the count is stable across the distro-variant path and the forced-CPU path, and it is signed-zero handling in all 19 cases rather than anything target-dependent — which is what the fix assumes.

That makes this a conformance failure in packaged PoCL on any CPU, not only on baseline-targeted builds: `math_edge_cases` is a 2026 addition to the CTS, so it is newly-visible old behaviour rather than a regression, exactly as the description says.

*AI tool usage: noted in the commit messages (`Assisted-by:`), per the AI Tool Use Policy.*

franz (approved):
Thanks for the patch, LGTM.

Yep, CTS occasionally catches new bugs when it introduces new tests.

PR #2323, opened 2026-09-16
cpu: don't strdup an uninitialised fallback kernel-library name

anun333:
`pocl_cpu_init_common` declares `kernellib_fallback[]` and fills it in the `KERNELLIB_HOST_DISTRO_VARIANTS` branch and in the default branch, but not in the `HOST_CPU_FORCED` branch — which then `strdup()`s it anyway. On a build configured with `-DLLC_HOST_CPU=...` that duplicates whatever happened to be on the stack, and `pocl_llvm_build.cc` builds a library load path out of the result.

Valgrind on a `-DLLC_HOST_CPU=x86-64` build, doing nothing but `clGetDeviceIDs`:

```
Conditional jump or move depends on uninitialised value(s)
   at strlen
   by strdup (strdup.c:41)
   by pocl_cpu_init_common
   by pocl_pthread_init
   by pocl_init_devices
   by POclGetDeviceIDs
 Uninitialised value was created by a stack allocation
   at pocl_cpu_init_common
```

A forced-CPU build has exactly one kernel library, so the right answer is that it has no fallback: initialise the buffer empty and pass `NULL` when it is still empty. Every consumer already checks for `NULL` (`pocl_llvm_build.cc:1102`, `pocl_mlir_build.cc:70,81`), and the other two branches are unaffected.

Found while testing baseline-targeted builds (`-DLLC_HOST_CPU=x86-64`), which is how distributions build for a wide CPU spread.

*AI tool usage: noted in the commit message (`Assisted-by:`), per the AI Tool Use Policy.*


anun333:
One correction to my own description, and it narrows the claim.

I wrote that `-DLLC_HOST_CPU=x86-64` is "how distributions build for a wide CPU spread". That is wrong. Debian's `debian/rules` and Fedora's `pocl.spec` both build x86 with **`KERNELLIB_HOST_CPU_VARIANTS=distro`** and select the variant at run time; neither forces `LLC_HOST_CPU` on x86 at all. I had assumed it rather than reading the packaging, and I have corrected the same claim elsewhere.

The bug is unaffected — it is in the `HOST_CPU_FORCED` branch, which is reached exactly when `LLC_HOST_CPU` is set, and the Valgrind trace above is from such a build. But the population is not "distributions on x86". It is:

- **Fedora on aarch64 and riscv64**, which do force it: `pocl.spec` passes `-DLLC_HOST_CPU=cortex-a53` for aarch64 and `-DLLC_HOST_CPU=generic-rv64` for riscv64. Fedora ships PoCL for riscv64 today, so these are shipped builds hitting the uninitialised `strdup`.
- anyone building for a fixed target, which includes embedded and SDK builds, and is the documented reason the option exists.

So: fewer affected users than I implied, but still shipped packages rather than only local experiments, and on the two architectures where a forced CPU is the packager's only sensible choice.

Nothing in the patch changes. I would just not want the description's last line standing as a claim about Debian.

*AI tool usage: noted in the commit message (`Assisted-by:`), per the AI Tool Use Policy.*

franz (approved):
Thanks, this is indeed a bug. We do have an address-sanitizer enabled CI, but it only runs one build configuration (not using HOST_CPU_FORCED).

PR #2327, opened 2026-09-16
libclc: select the large-argument trig reduction per lane

anun333:
## Summary

`__pocl_argReductionS` computes the small-argument trig reduction for all lanes, then, if *any* lane is >= 2^23, overwrites *every* lane with the Payne-Hanek large-argument result. That path is invalid for tiny inputs, so a vector whose lanes mix large and small values returns wrong results in the small lanes.

`sin`, `cos`, `tan` and `sincos` on `float` are affected whenever one lane of an explicit vector type is >= 2^23 and another is below roughly 1e-5. Scalar calls and `double` are unaffected.

## Reproducer

`sin((float4)(FLT_MIN, 1e8f, 1, 1))`: lane 0 returns `0x1.fffeaap-8` (0.0078) instead of `FLT_MIN`; `cos` returns `0x1.fffcp-1` instead of `1.0`.

The regression test in this PR, run against current main:

```
vec4_k lane 0 (x=0x1p-126):      sin 0x1.fffeaap-8  vs scalar 0x1p-126
                                 cos 0x1.fffcp-1    vs 0x1p+0
                                 tan 0x1.000156p-7  vs 0x1p-126
vec4_k lane 2 (x=0x1.b7cdfep-34): sin 0x1.b7ca9cp-7 vs scalar 0x1.b7cdfep-34
vec4_k lane 4 (x=0x1.4f8b58p-17): sin 0x1.4f89d8p-7 vs scalar 0x1.4f8b58p-17
FAIL: 6 lane mismatches
```

Same binary against a build with the fix: `OK`. The test covers `float4` and `float8` and compares each lane against the scalar result for the same input.

## Fix

Compute both reductions when any lane needs the large one, and select per lane, instead of replacing the whole vector. `SV_ANY(cond)` still guards the expensive path, so vectors with no large lane are unchanged.

## Notes

This was previously filed as #2309 from a GitHub account that has since been deleted, which orphaned the PR; it is re-filed here unchanged apart from being rebased onto current main. It is also the first commit of #2320, carved out so it can be reviewed on its own: the bug is in libclc's argument reduction and has nothing to do with that series' vector-math work.


anun333:
The two `test_trig_lane_mix` failures on the LLVM 20 and 21 jobs were the test, not the fix. It compared each vector lane bit for bit with the scalar kernel, and on those runners the two paths differ by one ulp in places: `cos(1e-5f)` is `1.0` from the vector path and `0x1.fffffep-1` from the scalar one (the true value rounds to `1.0`). Both are well inside the 4-ulp bound.

I've pushed a change to the test: it now checks every lane, and the scalar kernel, against a double-precision reference within the OpenCL bounds (sin and cos 4 ulp, tan 5 ulp). The values from the failing run are all under 1 ulp that way. Against a conformance build without the fix, where float trig comes from libclc, it still fails on the same six lanes as before, by up to 5.6e42 ulp; with the fix, it passes, along with the rest of the regression tests.

anun333:
The previous version passed on Linux x86-64 (LLVM 20 and 21), ARM64 and RISC-V, and failed only in the MinGW jobs, on lane 5 (`1e20f`): the test took its references from the host's `sin`, `cos` and `tan`, and MinGW's are off by millions of ulp at that argument, while PoCL's results there are within 0.3 ulp of the true values. This push replaces them with a fixed table computed in 300-bit arithmetic, so the test no longer depends on the host C runtime. Happy to squash the test commits into one if you'd prefer a cleaner history.

PR #2328, opened 2026-09-16
CPU: disable reciprocal estimates on targets without FMA

anun333:
## Summary

Under `-cl-fast-relaxed-math`, single-precision division on x86 targets without FMA is 3.0 ulp off; the OpenCL relaxed bound is 2.5 ulp. This is the configuration distributions build (`LLC_HOST_CPU=x86-64`, and every `KERNELLIB_HOST_CPU_VARIANTS=distro` variant below `avx2`). The fix tells LLVM not to use its reciprocal estimate on those targets.

## How it was found

OpenCL CTS `math_brute_force`, sub-test `divide`, on a conformance build of current main at `LLC_HOST_CPU=x86-64`:

```
  1:          divide fp32  rlx.
ERROR: divide: -3.000000 ulp error at {-0x1.09ce42p-38, -0x1.0b2ffap-44}: *0x1.fd5a3p+5 vs. 0x1.fd5a2ap+5 (0x427ead15) at index: 101410
  2:          divide fp32     passed
  3:          divide fp64     passed
  4:          divide fp16     passed
```

Only the relaxed sub-case fails. The same test on a native build (`cpu-znver3`, FMA) passes all four.

## Attribution

The same CTS sub-test was run on three further baseline cells, not just replayed:

| cell | source | `ENABLE_CONFORMANCE` | vector library | result |
|---|---|---|---|---|
| upstream `aaaa2d67b`, `LLC_HOST_CPU=x86-64` | main | ON | none | `rc=1`, 3.0 ulp, same inputs and bits |
| upstream `aaaa2d67b`, `LLC_HOST_CPU=x86-64` | main | OFF | libmvec | `rc=1`, identical |
| #2320's branch | | ON | none | `rc=1`, identical |
| #2320's branch, native (`znver3`) | | ON | libmvec | passed |

Two source trees, three vector-library settings, one failure. It is independent of any vector-math work and reproduces on plain main.

Replaying the CTS inputs through a one-line kernel `o[i] = a[i] / b[i]` built with `-cl-fast-relaxed-math`, then disassembling the cached object:

| target | result | vector loop |
|---|---|---|
| `x86-64` | `0x1.fd5a2ap+5`, 3.0 ulp | `rcpps` + 3 × `mulps` |
| `znver3` | 1.0 ulp max | `vrcpps` + `vfmsub213ps` + `vfnmadd213ps` |
| either, without relaxed math | exact | `divps` |

## Mechanism

With the `arcp` fast-math flag, LLVM's x86 backend replaces a vector `fdiv` by a 12-bit reciprocal estimate (`rcpps`) refined with one Newton-Raphson step. With FMA the step is a single rounding and the result stays within 1 ulp. Without FMA it is two multiplies and a subtract, rounding three times, and the CTS measures 3 ulp. C fast-math makes no accuracy promise there, so LLVM is within its rights; OpenCL relaxed math promises 2.5 ulp, so the implementation has to hold the line.

The knob is the `reciprocal-estimates` function attribute. Confirmed with clang directly: `-cl-fast-relaxed-math -march=x86-64` gives `rcpps`, adding `-mrecip=none` (which sets `"reciprocal-estimates"="none"`) gives `divps`.

## Fix

`unifyLLVMFunctionAttributes` already sets `target-cpu` / `target-features` on every function. When `target-features` lacks `+fma`, also set `reciprocal-estimates` to `none`. FMA targets are unchanged (their estimate passes the CTS). `sqrt` and `rsqrt` were checked at `x86-64` too: they lower to `sqrtps` and `sqrtps` + `divps`, no estimate involved, so division is the only affected operation today.

## Verification

Same CTS sub-test, same settings, on this branch at `LLC_HOST_CPU=x86-64`:

```
  1:          divide fp32  rlx................passed    0.00 @ {0x0p+0, 0x0p+0}
  2:          divide fp32     ................passed    0.00 @ {0x0p+0, 0x0p+0}
  3:          divide fp64     ................passed    0.00 @ {0x0p+0, 0x0p+0}
  4:          divide fp16     ................passed    0.00 @ {0x0p+0, 0x0p+0}
PASSED 4 of 4 sub-tests.
```

The failing cells above and this one differ only by this patch: same source
revision otherwise, same CMake flags, same CTS binary, same
`--num-worker-threads 1 -t 1` invocation.

The investigation and the patch were done with Claude Code; the measurements above were run and read by me.


anun333:
This also fixes `math_brute_force reciprocal`, which I hit later in the same CTS run. Same root cause, same shape:

```
  1: reciprocal fp32  rlx.
ERROR: reciprocal: -3.000000 ulp error at 0x1.04aefap-126 (0x0082577d):
       *0x1.f6cd22p+125 vs. 0x1.f6cd1cp+125
```

Only the relaxed sub-case, `-3.000000` ulp again, and it passes on a native (FMA) build. `1.0f / x` lowers through the same `rcpps` plus one Newton-Raphson step as a general division, so it has the same two-extra-roundings problem without FMA.

With this patch, at `ENABLE_CONFORMANCE=ON` and `LLC_HOST_CPU=x86-64`:

```
### done math_reciprocal rc=0 secs=453
PASSED 4 of 4 sub-tests.
```

So the one-hunk change fixes `divide` and `reciprocal` at the baseline target. Both were run with the same CTS binary and the same `--num-worker-threads 1 -t 1` invocation as the failing cells.

anun333:
One correction to my own description, after checking the actual packaging rather than assuming it.

I wrote that `LLC_HOST_CPU=x86-64` is "the configuration distributions build". It is not. Debian's `debian/rules` (salsa, `main`) and Fedora's `pocl.spec` (rawhide) both build x86 with **`KERNELLIB_HOST_CPU_VARIANTS=distro`** and `ENABLE_CONFORMANCE=ON`, so the kernel-library variant is selected at run time from the host's CPU features.

What that changes about who is affected: a user on any AVX2-or-newer CPU gets the `avx2` (haswell) variant, which has FMA, and is **not** affected by this bug. The affected population is users whose CPU selects one of the variants below `avx2` — `sse2`, `ssse3`, `sse41`, `avx`, `avx_f16c` — i.e. hardware without FMA. That is still a real and shipped population (Sandy/Ivy Bridge and older, and VMware guests pinned to a pre-Haswell EVC baseline), and the fix applies identically to them, because the patch keys on `+fma` in the target features rather than on any particular `LLC_HOST_CPU`. But "every distribution user" was an overstatement and I would rather say so than leave it.

The verification I reported was at `LLC_HOST_CPU=x86-64`; I am rerunning the same sub-test on an upstream `distro` build with `POCL_KERNELLIB_NAME=sse2` forced, which is the exact code path a pre-FMA user hits, and will post that result here.

anun333:
Following up on my own correction above, with a measurement instead of an inference.

I said the affected population is "users whose CPU selects a variant below `avx2`", but everything I had measured was on a synthetic `-DLLC_HOST_CPU=x86-64` build, which is not a configuration anyone ships. So I ran the full CTS on the configuration Debian and Fedora actually build, with a pre-`avx2` variant selected the way a user's CPU would select it.

### Configuration

Unpatched `aaaa2d67b` (current `main` at the time of writing), no patches of any kind:

```
cmake -DENABLE_CONFORMANCE=ON -DKERNELLIB_HOST_CPU_VARIANTS=distro
POCL_KERNELLIB_NAME=sse2        # the SSE-class variant, as a pre-Haswell CPU would select
```

x86-64 Zen 3 host, glibc 2.39, LLVM 22.1.8, OpenCL-CTS at `1a6f465`. `LLC_HOST_CPU` is **not** set — this is the distro-variant path, and `ENABLE_HOST_CPU_VECTORIZE_LIBMVEC` is off, i.e. exactly what a package build produces. `math_brute_force` sub-tests run one per process with `--num-worker-threads 1 -t 1`.

I verified the variable takes effect rather than assuming it: the emitted object has **0 VEX-encoded instructions** at `sse2` and **125** at `avx2`, and measured accuracy moves with it (`exp` float max 0.7858 vs 0.8508 ULP — the no-FMA / FMA split).

### Result

The run is still going — **98 of 153 units done, 3 failures** — and two of the three are this PR:

```
ERROR: divide: -3.000000 ulp error at {-0x1.09ce42p-38, -0x1.0b2ffap-44}:
       *0x1.fd5a3p+5 vs. 0x1.fd5a2ap+5 (0x427ead15) at index: 101410
divide FAILED — FAILED 1 of 4 sub-tests

ERROR: reciprocal: -3.000000 ulp error at 0x1.04aefap-126 (0x0082577d):
       *0x1.f6cd22p+125 vs. 0x1.f6cd1cp+125
reciprocal FAILED — FAILED 1 of 4 sub-tests
```

Both fail **only** the relaxed fp32 sub-test; the non-relaxed fp32, fp64 and fp16 sub-tests pass. The inputs are bit-identical to the ones I reported from the `x86-64` build, which is the expected signature of one root cause rather than two.

The third failure is `math_math_edge_cases`, which is pre-existing and also fails on a native build, so it is unrelated to this PR.

### What this changes about the claim

This is a conformance failure in PoCL **as distributions build it**, on any CPU that selects a variant without FMA — pre-Haswell hardware, and EVC-pinned hypervisor guests. It does not require `LLC_HOST_CPU`, it does not require a vector math library, and it does not require any patch of mine. That is a stronger and narrower statement than the one I originally put in the description, and it is the one I should have made.

### What I have not yet verified

The fix in this PR was confirmed on a `-DLLC_HOST_CPU=x86-64` build of the merge-base plus this patch: `math_divide` and `math_reciprocal` both pass there, where they fail without it. **I have not yet run that same check on the distro/`sse2` cell above.** I will post it when I have, since the configuration where the bug was just demonstrated is the configuration the fix ought to be demonstrated in.

Happy to post the full per-sub-test log or the build manifest for either cell if that would be useful.

anun333:
Following up on the check I said I would run.

The bug was demonstrated on the distro-variant configuration with a
pre-`avx2` variant forced — the one a package build produces and a
pre-Haswell CPU selects. The fix had only been demonstrated at
`-DLLC_HOST_CPU=x86-64`, which is not a configuration anyone ships. Same
cell, same selection, now with this patch applied:

| | `math_divide` | `math_reciprocal` |
|---|---|---|
| unpatched `aaaa2d67b`, distro variants, `POCL_KERNELLIB_NAME=sse2` | **FAILED** | **FAILED** |
| the same, plus this patch | **rc=0** | **rc=0** |

```
### done math_divide      rc=0  2434s
### done math_reciprocal  rc=0   471s
```

Against the unpatched run on the identical configuration:

```
### done math_divide      rc=1  2354s
### done math_reciprocal  rc=1   261s
```

Build: `-DENABLE_CONFORMANCE=ON -DKERNELLIB_HOST_CPU_VARIANTS=distro`, nine
variants built, `POCL_KERNELLIB_NAME=sse2` at run time, glibc 2.39, LLVM
22.1.8, x86-64 Zen 3 host, OpenCL-CTS at the same checkout as the earlier
run. `math_brute_force` one sub-test per process with
`--num-worker-threads 1 -t 1`. `ENABLE_HOST_CPU_VECTORIZE_LIBMVEC` is off,
so no vector math library is involved on either side.

So the failure and its absence are now both shown in the configuration
distributions actually build, rather than one in a synthetic target and the
other in a real one.

franz (changes requested):
Thanks for the patch. However, I'm not sure this works on architectures other than X86.

E.g. RISC-V includes FMA instructions in the "f" extension, and AFAICT it doesn't have "+fma" in the `TargetFeatures` string, even if the CPU does in fact support FMA. Similarly for Aarch64.

So AFAICT the condition as it is now, would likely disable reciprocal estimates not just on pre-FMA x86 CPUs, but also a large part of non-x86 CPUs.

anun333:
You're right, and it was wider than pre-FMA x86. `llc -mattr=help` on LLVM 22: riscv64 has no `fma` feature at all (FMA is part of `f`/`d`), and aarch64 has only `aggressive-fma`, `fp8fma` and `ssve-fp8fma`. So `find("+fma")` never matched there and the patch disabled reciprocal estimates on every RISC-V and AArch64 target.

The same line had a second flaw: it was a substring test on a comma-separated list, and `"+fma4"` contains `"+fma"`. `clang -march=bdver1` emits `+fma4` with no `+fma`, so an FMA4-only Bulldozer was read as having FMA3. Of the CPUs I checked, `bdver2`/`bdver4` carry both and `znver3` has `+fma`, so it's narrow, but it's a real miss. (It is also the exact mistake #2330 fixes for build options, which I should have caught here.)

Pushed 6c9e81aed:

- the override applies only when the module triple is `x86`/`x86_64`;
- a small `hasTargetFeature()` splits `target-features` on `,` and compares whole tokens.

Checked three ways, since the targets that were wrong can't be run here:

1. Builds at `-DLLC_HOST_CPU=x86-64` and `znver3`.
2. x86 behaviour unchanged, read back from `parallel.bc`: `x86-64` gets `reciprocal-estimates` set; `znver3` (`+fma`) does not.
3. The condition tested against real LLVM feature strings: `x86-64` disables, `znver3` keeps, `bdver1` (`+fma4` only) now disables, `bdver2` keeps, `riscv64` (`+m,+a,+f,+d,+c`) now keeps, `aarch64` (`+fp-armv8,+neon,+aggressive-fma`) now keeps. The last three are the ones that were wrong.

The `LLVM_MAJOR < 21` branch follows the `getTargetTriple()` guard already used in `pocl_llvm_build.cc`; I've only built it against LLVM 22 locally, so CI is the check for the older ones.

anun333:
@franz, ready for another look. `6c9e81aed` scopes the override to x86 and matches whole tokens, and `51ce4bf91` corrects a code comment that repeated my earlier "distributions build x86-64" mistake. I've updated the description to match. The CI hasn't run on this PR yet; it's waiting for approval, and the RISC-V job would exercise the case you raised.

PR #2331, opened 2026-09-20
Minor compatibility fixes for broader MSYS2 platform support

andrjohns:
This PR patches a few issues I ran into while putting together an MSYS2 package recipe for pocl:

- Windows ARM64 detected as x86_64 due to missing `CMAKE_SYSTEM_PROCESSOR` match
- MinGW LLD doesn't support the `--exclude-libs` flag
- Hard error for unknown CPU even with the `distro` kernel lib arg set (under parallels VM)
- `long` and `unsigned long` is 32-bit on windows

I've also expanded the MinGW CI job to cover the new msys platforms (including windows arm64): `UCRT64, CLANG64, CLANGARM64`

If these patches all look reasonable I'll open a pr to package pocl in MSYS (i.e., `mingw-w64-*-pocl`), backporting them to the 7.2 release

Let me know if I've missed anything, thanks!


.github/workflows/build_mingw.yml
```diff
@@ -43,68 +44,54 @@ jobs:
...
-          cmake -S ${{ env.WORKSPACE }} -B ${{ env.WORKSPACE }}/build-pocl \
+          BUILD_FLAGS="-O2 -Wall -Wextra -Wno-unused-parameter -Wno-unused-variable"
+          cmake -S . -B build-pocl \
             -DCMAKE_BUILD_TYPE=Release \
             "-DCMAKE_C_FLAGS_RELEASE=${BUILD_FLAGS}"  "-DCMAKE_CXX_FLAGS_RELEASE=${BUILD_FLAGS}" \
-            -DCMAKE_INSTALL_PREFIX=${{ env.WORKSPACE }}/install-pocl \
+            -DCMAKE_INSTALL_PREFIX="$PWD/install-pocl" \
             -DBUILD_SHARED_LIBS=${{ matrix.shared_libs }} \
             -DENABLE_ICD=OFF \
-            -DLLC_HOST_CPU=haswell \
```
franz, line 81: Hmm, wild. This was added as a workaround for a bug in Github environment, but your PR removes it, and I can see it still passes on GH runners with AVX512. Perhaps GH fixed the runner environment, or MinGW for some reason doesn't trigger this bug.

lib/CL/pocl_run_command.c
```diff
@@ -67,7 +67,7 @@ static char *
 build_cmd_from_arglist (const char **args)
 {
   /* Calculate required buffer size for all arguments. */
-  size_t total_len = 0;
```
franz, line 70: Is this to avoid a warning ?

franz (approved):
Thanks. Overall, LGTM.

